PR #50, opened 2020-09-15
Fix: Docker image names in readme files

akshay-ap:


trentmc:
Good catch @akshay-ap !

Sorry for the delay in getting to this.

Some of the READMEs were fixed a while back.

There was one remaining README needing fixing. It was simpler just to change that one README via a new PR rather than use the PR here. That was done, see #98.

Thanks again for your contribution @akshay-ap :+1:

PR #78, opened 2020-12-06
Adds function for check URL from provider.

### LINKED ISSUE #86: Add a function to use the checkURL endpoint from provider
The provider has a new endpoint to check the URL contentType and contentLength before proceeding. If the URL is missing, wrong, non-confoormant, does not support GET, or does not return anything, the checkURL endpoint returns a 400 response. This issue adds the possibility to call the checkURL endpoint from ocean.py.

calina-c:
Depends on oceanprotocol/provider#16
Closes #86.


ssallam:
It seems we have two PRs doing similar things: https://github.com/oceanprotocol/ocean.py/pull/80

PR #80, opened 2020-12-09
CheckURL created.

mariacarmina:
Establish the connection with the checkURL endpoint from the provider repo. I have modified in DataServiceProvider class from data_provider package.

### DIFF
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -10,6 +10,8 @@ import re
 from collections import namedtuple
 from json import JSONDecodeError
 
+from ocean_lib.data_provider import data_service_provider
+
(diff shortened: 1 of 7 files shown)


ssallam (changes requested):
please remove all the `.idea/` files from the repo

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -10,6 +10,8 @@
 from collections import namedtuple
 from json import JSONDecodeError
 
+from ocean_lib.data_provider import data_service_provider
```
ssallam, line 13: what is this doing? we are in the data_provider already!

ssallam:
It seems we have two PRs doing similar things: #78

calina-c:
Duplicated by #78.

PR #87, opened 2021-01-06
address transaction not found issue when running on poa

ssallam:
wait for transaction with a timeout to avoid tx not found error on poa network


alexcos20 (approved):
lgtm

PR #90, opened 2021-01-06
fix starting compute using algorithm did.

ssallam:
The provider requires the algorithm transferTxId and the algorithm datatoken address when using a published algorithm. Those attributes were missing from the call to start compute in ocean.compute.


alexcos20 (approved):
lgtm

PR #96, opened 2021-01-08
remove references to brizo, replace print with logger.debug

ssallam:


mariacarmina (approved):
Nice work! Self-descriptive code.

ssallam:
Thanks Maria

Issue #106, opened 2021-01-09
ocean.pool.create() does not work within a .py file but works within the Python shell. (+ a weird thing that makes it work within .py files)


surajsjain:
I tried `pool = ocean.pool.create(token_address, data_token_amount=100.0, OCEAN_amount=10.0, from_wallet=alice_wallet)` from the marketplace flow documentation within the python shell and it worked just fine. But when I put it in a **.py** file along with other code, it threw the following error:
```
Traceback (most recent call last):
  File "weird.py", line 76, in <module>
    pool = ocean.pool.create(
  File "path_to_virtual_env/lib/python3.8/site-packages/ocean_lib/ocean/ocean_pool.py", line 69, in create
    pool_address = bfactory.newBPool(from_wallet)
  File "path_to_virtual_env/lib/python3.8/site-packages/ocean_lib/models/bfactory.py", line 16, in newBPool
    tx_id = self.send_transaction(
  File "path_to_virtual_env/lib/python3.8/site-packages/ocean_lib/web3_internal/contract_base.py", line 196, in send_transaction
    return contract_function.transact(_transact).hex()
  File "path_to_virtual_env/lib/python3.8/site-packages/ocean_lib/web3_internal/web3_overrides/contract.py", line 61, in transact
    return transact_with_contract_function(
  File "path_to_virtual_env/lib/python3.8/site-packages/ocean_lib/web3_internal/web3_overrides/contract.py", line 110, in transact_with_contract_function
    txn_hash = web3.eth.sendRawTransaction(raw_tx)
  File "path_to_virtual_env/lib/python3.8/site-packages/web3/eth.py", line 272, in sendRawTransaction
    return self.web3.manager.request_blocking(
  File "path_to_virtual_env/lib/python3.8/site-packages/web3/manager.py", line 112, in request_blocking
    raise ValueError(response["error"])
ValueError: {'code': -32000, 'message': 'replacement transaction underpriced'}
```

So, just out of curiosity, I tried the following within the **.py** file:
```
while(True):
        try:
            pool = ocean.pool.create(token_address, data_token_amount=100.0, OCEAN_amount=10.0, from_wallet=alice_wallet)
            break
        except:
            continue
```
I just told it to keep on trying until it gets it and not throw an error. Surprisingly, **this worked!!**

[...]

PR #107, opened 2021-01-12
Added sleep before creating a pool

surajsjain:
Etherium contracts take time to be created. So, if the program is run by copying the code to a .py file, it will throw an error. Therefore, the program must be paused for 30 seconds before creating the pool if it is executed through a .py file. This is linked to the issue #106


trentmc (approved):
Approving. I will make a follow-up change to tweak this.

trentmc:
Follow-up change: https://github.com/oceanprotocol/ocean.py/commit/7577639b4875fa249c6479f03c838a5dd9848652

PR #109, opened 2021-01-12
Adds function for check URL from provider

### LINKED ISSUE #86: Add a function to use the checkURL endpoint from provider

calina-c:
Supersedes #78.
Closes #86.


tests/ocean/test_compute_flow.py
```diff
@@ -17,6 +17,16 @@
     mint_tokens_and_wait, get_resource_path, wait_for_ddo)
 
 
+def test_check_url():
+    check_url_endpoint = DataServiceProvider.build_check_url_endpoint()
+    result = DataServiceProvider.check_url(
+        'http://xkcd.com/349/info.0.json', check_url_endpoint
```
alexcos20, line 23: I have uploaded that to our S3 buckets, so we can control it
```suggestion
        'https://s3.amazonaws.com/testfiles.oceanprotocol.com/info.0.json', check_url_endpoint
```

tests/ocean/test_compute_flow.py
```diff
@@ -17,6 +17,16 @@
...
+    )
+
+    assert result['contentLength'] == '629'
```
alexcos20, line 26:
```suggestion
    assert result['contentLength'] == '1161'
```

alexcos20: not sure why the same file is 1161 bytes in s3, I'm assuming that when I saved your response somehow it was converted to UTF8, doubling the size.. :)

alexcos20 (changes requested):
let's use our hosted fine and then this PR is ready to go

alexcos20 (approved):

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -68,6 +68,34 @@ def encrypt_files_dict(files_dict, encrypt_endpoint, asset_id, publisher_address
 
             return response.json()['encryptedDocument']
 
+    @staticmethod
+    def check_url(url, check_url_endpoint):
```
ssallam, line 72: * can we make `check_url_endpoint` optional, if not set, build it internally.
* doc string missing, what does this function achieve
* what is the return value? bool, string ...

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -68,6 +68,34 @@ def encrypt_files_dict(files_dict, encrypt_endpoint, asset_id, publisher_address
...
+            raise OceanEncryptAssetUrlsError(msg)
+
+        result = response.json()['result']
+
+        logger.info(
+            f"Check URL was successful, content type: {result['contentType']},"
+            f" content length {result['contentLength']}')"
+        )
+
+        return result
```
ssallam, line 97: what is the type of `result`, is it a text message or simply a boolean

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -68,6 +68,34 @@ def encrypt_files_dict(files_dict, encrypt_endpoint, asset_id, publisher_address
...
+            not response or
+            not hasattr(response, 'status_code') or
+            response.status_code != 200
+        ):
+            msg = (f'Could not determine Content-Length and Content-Type '
+                   f'{check_url_endpoint}, status {response.status_code}')
+            logger.error(msg)
+            raise OceanEncryptAssetUrlsError(msg)
+
+        result = response.json()['result']
```
ssallam, line 90: why is it {'result': value}, can it just be `value` which is a dict?

ssallam (changes requested):
see inline comments

ocean_lib/data_provider/data_service_provider.py
calina-c: that's how the endpoints seemed to be structured on provider: either a "result" with the actual result inside, or an "error" with the error description. I inspired myself from the previous endpoints and it seemed to make sense. Do you want me to change the provider endpoint to return the result directly?

ocean_lib/data_provider/data_service_provider.py
calina-c: it is currently a json, but we can return a boolean instead, for simplicity. Let me know how I should proceed.

calina-c: CC @ssallam: @alexcos20 said it must be a dictionary, not a boolean

ocean_lib/data_provider/data_service_provider.py
ssallam: if it's consistent with the response from other provider endpoints then  let's keep it like this.

ssallam (dismissed):

ocean_lib/data_provider/data_service_provider.py
alexcos20: the response should be :
{  valid = false,  result={"contentLength":"", "contentType":""} }

ssallam (approved):

PR #111, opened 2021-01-15
Dynamic provider endpoints

### LINKED ISSUE #110: Ocean.py:  Dynamic provider endpoints
Blocked by https://github.com/oceanprotocol/provider/issues/36

Right now all endpoints are hardcoded (https://github.com/oceanprotocol/ocean.py/blob/master/ocean_lib/data_provider/data_service_provider.py#L372)

Ocean.py should read the provider root endpoint (/ ) and create the required URLs based on that information.
Also errors are going to be thrown if a service that is not exposed is being accessed.

mariacarmina:
Fixes #110 .

Changes proposed in this PR:

- created a hardcoded list with the endpoints that exist in provider;
- modified the build_endpoint functions;
- ocean.py reads from the provider serviceEndpoints.

### DIFF
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -22,7 +22,6 @@ from ocean_lib.web3_internal.web3helper import Web3Helper
 
 logger = logging.getLogger(__name__)
 
-
 OrderRequirements = namedtuple('OrderRequirements', ('amount', 'data_token_address', 'receiver_address', 'nonce'))
 
 
@@ -353,21 +352,17 @@ class DataServiceProvider:
     @staticmethod
     def get_valid_endpoints():
         service_endpoints = {
-            "download": "/api/v1/services/download",
-            "computeStart": "/api/v1/services/compute",
-            "computeDelete": "/api/v1/services/compute",
-            "computeStop": "/api/v1/services/compute",
-            "computeStatus": "/api/v1/services/compute",
-            "nonce": "/api/v1/services/nonce",
-            "encrypt": "/api/v1/services/encrypt",
-            "initialize": "/api/v1/services/initialize",
+            "download": ["GET", "/api/v1/services/download"],
+            "computeStart": ["POST", "/api/v1/services/compute"],
+            "computeDelete": ["DELETE", "/api/v1/services/compute"],
+            "computeStop": ["PUT", "/api/v1/services/compute"],
+            "computeStatus": ["GET", "/api/v1/services/compute"],
+            "nonce": ["GET", "/api/v1/services/nonce"],
+            "encrypt": ["POST", "/api/v1/services/encrypt"],
+            "initialize": ["GET", "/api/v1/services/initialize"],
+            "checkURL": ["POST", "/api/v1/services/checkURL"],
         }
-        valid_names = list()
-        for (key, value) in service_endpoints.items():
-            valid_names.append(tuple([key, value]))
-        return valid_names
-
-    valid_services = get_valid_endpoints()
+        return service_endpoints
 
     @staticmethod
     def build_endpoint(service_name, provider_uri=None, config=None):
@@ -384,39 +379,25 @@ class DataServiceProvider:
         api_version = DataServiceProvider.get_api_version()
         if api_version not in provider_uri:
             provider_uri = f'{provider_uri}/{api_version}'
-        return f'{provider_uri}/services/{service_name}'
+        valid_services = DataServiceProvider.get_valid_endpoints()
+        method, url = valid_services[service_name]
+        return method, f'{provider_uri}/services/{url}'
 
     @staticmethod
     def build_encrypt_endpoint(provider_uri=None):
-        for i in range(0, len(DataServiceProvider.valid_services)):
-            if DataServiceProvider.valid_services[i][0] == 'encrypt':
-                return DataServiceProvider.build_endpoint(DataServiceProvider.valid_services[i][0],
-                                                          DataServiceProvider.valid_services[i][1])
-        # return DataServiceProvider.build_endpoint('encrypt', provider_uri)
+        return DataServiceProvider.build_endpoint('encrypt', provider_uri)
 
     @staticmethod
     def build_initialize_endpoint(provider_uri=None):
-        for i in range(0, len(DataServiceProvider.valid_services)):
-            if DataServiceProvider.valid_services[i][0] == 'initialize':
-                return DataServiceProvider.build_endpoint(DataServiceProvider.valid_services[i][0],
-                                                          DataServiceProvider.valid_services[i][1])
-        # return DataServiceProvider.build_endpoint('initialize', provider_uri)
+        return DataServiceProvider.build_endpoint('initialize', provider_uri)
 
     @staticmethod
     def build_download_endpoint(provider_uri=None):
-        for i in range(0, len(DataServiceProvider.valid_services)):
-            if DataServiceProvider.valid_services[i][0] == 'download':
-                return DataServiceProvider.build_endpoint(DataServiceProvider.valid_services[i][0],
-                                                          DataServiceProvider.valid_services[i][1])
-        # return DataServiceProvider.build_endpoint('download', provider_uri)
+        return DataServiceProvider.build_endpoint('download', provider_uri)
 
     @staticmethod
     def build_compute_endpoint(provider_uri=None):
-        for i in range(0, len(DataServiceProvider.valid_services)):
-            if DataServiceProvider.valid_services[i][0] == 'compute':
-                return DataServiceProvider.build_endpoint(DataServiceProvider.valid_services[i][0],
-                                                          DataServiceProvider.valid_services[i][1])
-        # return DataServiceProvider.build_endpoint('compute', provider_uri)
+        return DataServiceProvider.build_endpoint('compute', provider_uri)
 
     @staticmethod
     def get_initialize_endpoint(service_endpoint):
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -81,7 +81,7 @@ class DataServiceProvider:
 
     @staticmethod
     def get_nonce(user_address, config):
-        method, url = DataServiceProvider.build_endpoint('nonce')
+        _, url = DataServiceProvider.build_endpoint('nonce')
         response = DataServiceProvider._http_client.get(
             f'{url}?userAddress={user_address}'
         )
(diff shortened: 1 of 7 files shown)


mariacarmina:
Need a feedback for first try implementation.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -365,24 +384,39 @@ def build_endpoint(service_name, provider_uri=None, config=None):
...
         return f'{provider_uri}/services/{service_name}'
 
     @staticmethod
     def build_encrypt_endpoint(provider_uri=None):
-        return DataServiceProvider.build_endpoint('encrypt', provider_uri)
+        for i in range(0, len(DataServiceProvider.valid_services)):
+            if DataServiceProvider.valid_services[i][0] == 'encrypt':
+                return DataServiceProvider.build_endpoint(DataServiceProvider.valid_services[i][0],
+                                                          DataServiceProvider.valid_services[i][1])
+        # return DataServiceProvider.build_endpoint('encrypt', provider_uri)
```
ssallam, line 395: we need to get the serviceEndpoints dict from the provider, using the root provider endpoint that was added in `https://github.com/oceanprotocol/provider/pull/46`. Then extract the (http-method, url) pair for the specific operation.
All the specific functions here should not be changed, the main change will be in `build_endpoint` function:
example:
```
def build_endpoint(service_name, provider_uri=None, config=None):
    ...
    endpoints_dict = DataServiceProvider.get_service_endpoints()
    method, url = endpoints_dict[service_name]
    return method, f'{provider_uri}/services/{url}'
```

mariacarmina: Ok, ok thank you for suggestion. Firstly, I modified in build_endpoint then I switched to all thos functions and I thounght it wasn't correct. I will refactor those.

mariacarmina: those* thought*

mariacarmina: Actually, it's not needed to have f'{provider_uri}/services/{url}' if I have the url from the endpoint_dict, because the format will add a redundant '/' and will duplicate the path. I think it should be f'{url}'.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -350,6 +351,13 @@ def get_url(config):
     def get_api_version():
         return DataServiceProvider._remove_slash(os.getenv(ENV_PROVIDER_API_VERSION, DataServiceProvider.API_VERSION))
 
+    @staticmethod
+    def get_provider_information():
+        if DataServiceProvider.provider_info is None:
+            DataServiceProvider.provider_info = requests.get('http://localhost:8030')
```
ssallam, line 357: why is this using `localhost`?

mariacarmina: Shoul I get the url from config? I put localhost, because I wanted to be sure that I have the same URL from the browser, too.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -350,6 +351,13 @@ def get_url(config):
     def get_api_version():
         return DataServiceProvider._remove_slash(os.getenv(ENV_PROVIDER_API_VERSION, DataServiceProvider.API_VERSION))
 
+    @staticmethod
+    def get_provider_information():
```
ssallam, line 355: if we only need the serviceEndpoints, let's just return that and rename this to reflect the return value.

ocean_lib/data_provider/data_service_provider.py
ssallam: so everyone using ocean.py will have to run their own provider locally!

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -382,7 +393,7 @@ def build_download_endpoint(provider_uri=None):
 
     @staticmethod
     def build_compute_endpoint(provider_uri=None):
-        return DataServiceProvider.build_endpoint('compute', provider_uri)
+        return DataServiceProvider.build_endpoint('computeStatus', provider_uri)
```
ssallam, line 396: This will not work, we have to replace this with specific compute endpoints. so one for computeStatus, computeStart, ...

mariacarmina: Ok, compute is now the computeStatus and I need to implement for the rest.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -350,6 +351,14 @@ def get_url(config):
     def get_api_version():
         return DataServiceProvider._remove_slash(os.getenv(ENV_PROVIDER_API_VERSION, DataServiceProvider.API_VERSION))
 
+    @staticmethod
+    def get_service_endpoints():
+        if DataServiceProvider.provider_info is None:
+            config = ConfigProvider.get_config()
+            DataServiceProvider.provider_info = requests.get(config.provider_url)
+            DataServiceProvider.provider_info = DataServiceProvider.provider_info.json()
+        return DataServiceProvider.provider_info
```
ssallam, line 360: can this return `DataServiceProvider.provider_info['serviceEndpoints']`?

mariacarmina: ok, sure.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -366,7 +375,10 @@ def build_endpoint(service_name, provider_uri=None, config=None):
         if api_version not in provider_uri:
             provider_uri = f'{provider_uri}/{api_version}'
 
-        return f'{provider_uri}/services/{service_name}'
+        valid_provider_info = DataServiceProvider.get_service_endpoints()
+        method, url = valid_provider_info['serviceEndpoints'][service_name]
```
ssallam, line 379: see comment above, you can remove `['serviceEndpoints']` and maybe rename `valid_provider_info` to `service_endpoints`

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -366,7 +375,10 @@ def build_endpoint(service_name, provider_uri=None, config=None):
...
+        url = url.replace('api/v1/', '')
+        return f'{provider_uri}{url}'
```
ssallam, line 381: the returned value now omits the http `method`, this means we are getting the url from the provider but the method is hard coded.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -80,7 +81,7 @@ def sign_message(wallet, msg, config, nonce=None):
 
     @staticmethod
     def get_nonce(user_address, config):
-        url = DataServiceProvider.build_endpoint('nonce')
+        method, url = DataServiceProvider.build_endpoint('nonce')
```
calina-c, line 84: are you using the `method`? If not, you can replace the variable name with `_` to avoid an unused variable. I am unsure about the business logic here, but maybe you need to check that the nonce endpoint supports GET (with assert, maybe?)

CC: @ssallam any pointers here?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -350,6 +351,17 @@ def get_url(config):
...
+    @staticmethod
+    def get_service_endpoints():
+        """
+        Return the service endpoints from the provider URL.
+        """
+        if DataServiceProvider.provider_info is None:
+            config = ConfigProvider.get_config()
+            DataServiceProvider.provider_info = requests.get(config.provider_url)
+            DataServiceProvider.provider_info = DataServiceProvider.provider_info.json()
+        return DataServiceProvider.provider_info['serviceEndpoints']
```
calina-c, line 363: please add an empty line before the return statement.
I am also confused about L361 and L362 assigning the same variable in consecutive lines. Can you simplify to `requests.get(config.provider_url).json()`?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -360,13 +372,16 @@ def build_endpoint(service_name, provider_uri=None, config=None):
...
 
         api_version = DataServiceProvider.get_api_version()
         if api_version not in provider_uri:
             provider_uri = f'{provider_uri}/{api_version}'
 
-        return f'{provider_uri}/services/{service_name}'
+        service_endpoints = DataServiceProvider.get_service_endpoints()
+        method, url = service_endpoints[service_name]
+        url = url.replace('api/v1/', '')
+        return method, f'{provider_uri}{url}'
```
calina-c, line 384: please add an empty line before the return statement.
Will `api/v1/` always be the replacement? Could we possibly have another format? What will happen in that case?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -382,14 +397,30 @@ def build_download_endpoint(provider_uri=None):
...
+    def build_fileinfo(provider_uri=None):
+        return DataServiceProvider.build_endpoint('fileinfo', provider_uri)
 
     @staticmethod
     def get_initialize_endpoint(service_endpoint):
         parts = service_endpoint.split('/')
         if parts[-2] == 'services':
             base_url = '/'.join(parts[:-2])
-            return f'{base_url}/services/initialize'
+            return "GET", f'{base_url}/services/initialize'
```
calina-c, line 423: why is the `initialize` endpoint handled differently? Can't the logic from fileinfo, startCompute etc. be reused?

ocean_lib/ocean/ocean_assets.py
```diff
@@ -224,7 +224,7 @@ def create(self, metadata: dict, publisher_wallet: Wallet,
         logger.debug('Encrypting content urls in the metadata.')
 
         publisher_signature = self._data_provider.sign_message(publisher_wallet, asset.asset_id, self._config)
-        encrypt_endpoint = self._data_provider.build_encrypt_endpoint(provider_uri)
+        encrypt_endpoint_method, encrypt_endpoint = self._data_provider.build_encrypt_endpoint(provider_uri)
```
calina-c, line 227: please use `_` here and in other places where you don't plan on using the variable name in further lines. This way we can avoid warnings about unused variables.

calina-c (changes requested):
A few changes, but mostly questions I have :)

ocean_lib/data_provider/data_service_provider.py
mariacarmina: 'api/v1/' occurs twice in url field.

calina-c (approved):

ssallam (approved):

kremalicious:
was the closing of this PR intended? It has 2 approvals so was the intention to merge it?

PR #119, opened 2021-02-03
Ocean.py: data_token.,verify_order_tx : should allow access if the sender is either the consumer or the payer of an order

### LINKED ISSUE #116: ocean.py: datatoken.verify_order_tx: should allow access if the sender is either the consumer or the payer of an order
Provider is using this function to check the access rights

Order event definition:  https://github.com/oceanprotocol/contracts/blob/master/contracts/templates/DataTokenTemplate.sol#L33

mariacarmina:
Fixes #116  .

Changes proposed in this PR:

- verify_order_tx function should allow access for both roles that a sender can have: consumer or payer


ocean_lib/ocean/ocean_assets.py
```diff
@@ -351,13 +351,20 @@ def order(self, did: str, consumer_address: str,
         """
         assert service_type or service_index, f'One of service_index or service_type is required.'
         asset = self.resolve(did)
+        print("Asset: ", asset)
```
alexcos20, line 354: could we remove this debug logs ?

alexcos20 (changes requested):

ocean_lib/ocean/ocean_assets.py
mariacarmina: Sure.

alexcos20 (approved):
lgtm.  @ssallam @calina-c ?

ocean_lib/models/data_token.py
```diff
@@ -217,17 +217,18 @@ def verify_order_tx(self, web3, tx_id, did, service_id, amount_base, sender):
 
         # verify sender of the tx using the Tx record
         tx = web3.eth.getTransaction(tx_id)
-        if tx['from'] != sender:
+        if order_log.args.consumer != sender and order_log.args.payer != sender:
```
calina-c, line 220: to avoid and/or ambiguity, can we make this condition: `if sender not in [order_log.args.consumer, order_log.args.payer]`?

ocean_lib/models/data_token.py
```diff
@@ -217,17 +217,18 @@ def verify_order_tx(self, web3, tx_id, did, service_id, amount_base, sender):
...
 
         transfer_logs = self.events.Transfer().processReceipt(tx_receipt)
         receiver_to_transfers = {}
         for tr in transfer_logs:
-            if tr.args.to not in receiver_to_transfers:
-                receiver_to_transfers[tr.args.to] = []
-
-            receiver_to_transfers[tr.args.to].append(tr)
-
+            if order_log.args.consumer == sender or order_log.args.payer == sender:
```
calina-c, line 226: and we can reuse the same condition (without `not` this time, I think) here.

calina-c (changes requested):
Just a small (and unimportant) suggestion to help with readability.

calina-c (approved):

calina-c (approved):

PR #128, opened 2021-02-11
Renamed metadataStoreUri to metadataCacheUri.

### LINKED ISSUE #126: Rename "metadataStoreUri" to "metadataCacheUri"
In the config dict, for ocean-lib-py and ocean-lib-js (and elsewhere), there's a field 'metadataStoreUri' which typically points to aquarius.

That status quo label is there for historical reasons: when aquarius was actually the storage for metadata. But this is no longer the case. Now, the blockchain is the storage and aquarius is simply the cache. 

To rectify this, I recommend renaming to 'metadataCacheUri'.

Tasks:
- [ ] Rename in ocean.py
- [x] Rename in ocean.js
- [x] Rename in react
- [x] Rename in market
- [x] Identify other docs & components that use 'metadataStoreUri', and rename them too

mariacarmina:
Fixes #126 .

Changes proposed in this PR:

- to rename the config dictionary key called metadataStoreUri to metadataCacheUri.

### DIFF
--- a/READMEs/marketplace_flow.md
+++ b/READMEs/marketplace_flow.md
@@ -48,7 +48,7 @@ from ocean_lib.data_provider.data_service_provider import DataServiceProvider
 from ocean_utils.agreements.service_factory import ServiceDescriptor
 config = {
    'network' : os.getenv('NETWORK_URL'),
-   'metadataStoreUri' : os.getenv('AQUARIUS_URL'),
+   'metadataCacheUri' : os.getenv('AQUARIUS_URL'),
    'providerUri' : os.getenv('PROVIDER_URL'),
 }
 ocean = Ocean(config)
@@ -148,7 +148,7 @@ import os
 from ocean_lib.ocean.ocean import Ocean
 config = {
    'network' : os.getenv('NETWORK_URL'),
-   'metadataStoreUri' : os.getenv('AQUARIUS_URL'),
+   'metadataCacheUri' : os.getenv('AQUARIUS_URL'),
    'providerUri' : os.getenv('PROVIDER_URL'),
 }
 market_ocean = Ocean(config)
@@ -197,7 +197,7 @@ import os
 from ocean_lib.ocean.ocean import Ocean
 config = {
    'network' : os.getenv('NETWORK_URL'),
-   'metadataStoreUri' : os.getenv('AQUARIUS_URL'),
+   'metadataCacheUri' : os.getenv('AQUARIUS_URL'),
    'providerUri' : os.getenv('PROVIDER_URL'),
 }
 bob_ocean = Ocean(config)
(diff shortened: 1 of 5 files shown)


trentmc (commented):
I searched for all occurrences of this in the oceanprotocol org. [Here are the results](https://github.com/search?l=&q=org%3Aoceanprotocol+metadataStoreUri+user%3Aoceanprotocol&type=code).

The changes here capture all the occurrences found. With one exception, in the bb repo [here](https://github.com/oceanprotocol/bundesbank/blob/b32b88792d7e8d22927ecb49fc3560f2b05c8e31/py-cli/ocean_cli/util.py). Please check with @ssallam if it needs changing here.

Finally, other ocean.py users' config setups would break when this change is made. To be backwards compatible,  ocean_lib/ocean/ocean.py would need to try using the a config value of "metadataStoreUri" as a fallback. Can probably be handled in one line of code; but also have a unit test to check for both.

READMEs/marketplace_flow.md
```diff
@@ -48,7 +48,7 @@ from ocean_lib.data_provider.data_service_provider import DataServiceProvider
 from ocean_utils.agreements.service_factory import ServiceDescriptor
 config = {
    'network' : os.getenv('NETWORK_URL'),
-   'metadataStoreUri' : os.getenv('AQUARIUS_URL'),
+   'metadataCacheUri' : os.getenv('AQUARIUS_URL'),
```
mariacarmina, line 51: @trentmc, I have updated the 'metadataStoreUri' in READMEs, too.

trentmc: Yes, I acknowledge this, good.

The reason that we need to go through the release process is:

If this is merged *without* a new release, then the READMEs would have `metadataCacheUri` but anyone doing an `import ocean` from the ocean.py pypi library would get the old ocean.py.

ocean_lib/ocean/ocean.py
```diff
@@ -72,10 +72,15 @@ def __init__(self, config=None, data_provider=None):
             except AssertionError:
                 config = Config(os.getenv(ENV_CONFIG_FILE))
                 ConfigProvider.set_config(config)
-
         if isinstance(config, dict):
+            # fallback to metadataStoreUri
+            cache_key = (
```
trentmc, line 77: Can make this a one-liner: 
```
cache_key = "metadataCacheUri" if ("metadataCacheUri" in config) else "metadataStoreUri"
```

trentmc:
I did a first review. Good to see the unit test on both. The rest looks good. I gave a suggested tweak to turn 4 lines into one.

After you make the change, I will do a quick re-review and you can merge, then deploy a new ocean.py to pip.

ocean_lib/ocean/ocean.py
trentmc: I guess you refactored to be across 3 lines. That's ok since both the 1-line and 3-line version align with python idioms.

trentmc (approved):

trentmc:
All good. After the merge please be sure deploy a new ocean.py to pip.

PR #139, opened 2021-02-19
Secure algorithms

### LINKED ISSUE #114: Ocean.py: Secure algorithms
change the structure of trustedAlgorithms in the ddo from :
trustedAlgorithms=['DID1','DID2']
to
publisherTrustedAlgorithms=
{
"did1":{
"filesChecksum":"1234",
"containerSectionChecksum":"7676"
},
"did2":{
"filesChecksum":"1232334",
"containerSectionChecksum":"98787"
}
}

To produce filesChecksum:

sha256(algorithm_ddo.service['metadata'].attributes.encryptedFiles + JSON.Stringify(algorithm_ddo.service['metadata'].attributes.main.files) )

To produce containerSectionChecksum:

sha256(JSON.Stringify(algorithm_ddo.service['metadata'].attributes.main.algorithm.container))

calina-c:
Closes #114 .

Changes proposed in this PR:

- 
- 
-


tests/ocean/test_compute_flow.py
```diff
@@ -156,13 +157,19 @@ def run_compute_test(
...
+            status and status["ok"]
+        ), f"something not right about the compute job, got status: {status}"
+
+        status = ocean_instance.compute.stop(did, job_id, consumer_wallet)
+        print(f"got job status after requesting stop: {status}")
+        assert (
+            status
+        ), f"something not right about the compute job, got status: {status}"
+    else:
+        assert status["status"] == 400
```
calina-c, line 171: I am a big fan of "if" inversion. I'd treat the expect_failure case first, then add a `return` line. I'd then go on with the flow that was previously in place, without the need for an else branch (return makes it moot).

I would also change the assert message to something more suited, e.g. "request expected 400 error but got status:" ...

ocean_lib/ocean/ocean_assets.py
```diff
@@ -319,6 +359,17 @@ def create(
...
+    # I need an update function for trusted_algorithms
+    # that I will call in test_compute_flow.py
+    def update_trusted_algorithms(self, trusted_algorithms=None):
+        services = self._process_service_descriptors(
+            self.service_descriptors,
+            self.metadata_copy,
+            self.provider_uri,
+            self.publisher_wallet,
+            trusted_algorithms=trusted_algorithms,
+        )
```
calina-c, line 371: I think this is a good start, but it is still incomplete. See how services are attached in the create function and try to mimic that. Keep all previous services of the asset in place, but set/change the compute service's trusted algorithms. I think that is the general idea, but more guidance from @ssallam is always welcome! :D

ocean_lib/ocean/ocean_assets.py
```diff
@@ -96,6 +99,38 @@ def _process_service_descriptors(
         if access_service_descriptor:
             _service_descriptors.append(access_service_descriptor)
         if compute_service_descriptor:
+            if trusted_algorithms:
+                trusted_algo_list = []
```
calina-c, line 103: I would separate the creation of trusted_algo_list into a different function. It seems there's too much indentation here.

tests/ocean/test_compute_flow.py
```diff
@@ -215,3 +237,47 @@ def test_compute_multi_inputs():
...
+    # Dataset with compute service
+    compute_ddo = get_registered_ddo_with_compute_service(
+        setup.publisher_ocean_instance,
+        setup.publisher_wallet,
+        trusted_algorithms=[algorithm_ddo.did],
+    )
+    # verify the ddo is available in Aquarius
+    _ = setup.publisher_ocean_instance.assets.resolve(compute_ddo.did)
+
+    # For debugging.
```
calina-c, line 265: Please uncomment this. It is important we test trusted_algorithms in both cases: the successful one, and the unsuccessful one.

.gitignore
```diff
@@ -4,6 +4,7 @@ __pycache__
 *.pyc
 *~
 .eggs/
+artifacts/address.json
```
calina-c, line 7: why was this even in the codebase? Is it supposed to be in the codebase? (Question for Samer, I guess)

calina-c (commented):
It's definitely an improvement and a lot of progress has been done with this task!
At the moment I'm writing this, I don't know if the tests pass, but I know they pass locally using the `ptalgs` image (corresponding to the open PR for security in provider repo). And I know it's been difficult, since you had to add the possibility of computeStart failure, which was not part of the original task (but was necessary to add).

Please take a look at my comments and fix them when possible. Then we only need to add a utilitary function for updating the trusted Algorithms. (please test that as well)

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -655,7 +662,6 @@ def _prepare_compute_payload(
             )
         else:
             payload["algorithmMeta"] = algorithm_meta
-
```
calina-c, line 658: a bit of a nitpick here: please add this line back, keeping only relevant changes.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -64,6 +66,7 @@ def _process_service_descriptors(
         metadata: dict,
         provider_uri: str,
         wallet: Wallet,
+        trusted_algorithms: list,
```
calina-c, line 69: I just noticed this! I remember coding this myself in the experimental commit, but it was probably overwritten in the meantime. The default value should be None here (and probably no type hinting) See: https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

tests/ocean/test_compute_flow.py
```diff
@@ -215,3 +237,47 @@ def test_compute_multi_inputs():
...
+    )
+    _ = setup.publisher_ocean_instance.assets.resolve(algorithm_ddo_v2.did)
+
+    # Dataset with compute service
+    compute_ddo = get_registered_ddo_with_compute_service(
+        setup.publisher_ocean_instance,
+        setup.publisher_wallet,
+        trusted_algorithms=[algorithm_ddo.did],
+    )
+    # verify the ddo is available in Aquarius
```
calina-c, line 262: please copy this comment over all the similar lines, for clarity

calina-c:
Also, until we merge the branch in provider, you can set the provider version to `ptAlgs` in tox (?)

.gitignore
mariacarmina: Btw, I modified an address right there and I wanted to push the modifications `address.json` file appeared as modified and ready to be committed so I tried to put that in `.gitignore`.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -96,6 +124,14 @@ def _process_service_descriptors(
         if access_service_descriptor:
             _service_descriptors.append(access_service_descriptor)
         if compute_service_descriptor:
+            if trusted_algorithms:
```
calina-c, line 127: move the if too;
the function name add_trusted_algorithms is ambiguous. It does not ADD anything, only if you move the part that modifies the service descriptor to that function. Otherwise it formats publisher trusted algorithms, and should be named format_publisher_trusted_algorithms and moved into a utils file (no need for "self" at all). I think that is the best course of action is the second one, to avoid name clashes with the utilitary function.

calina-c (commented):
Just a few left/new.

.gitignore
calina-c: @mariacarmina please remove this from the commit. Trent is already working on deleting this file altogether.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -334,6 +334,24 @@ def create(
...
+
+        if compute_service_descriptor:
+            if trusted_algorithms:
+                trusted_algorithms_list = create_publisher_trusted_algorithms(
+                    self, trusted_algorithms=trusted_algorithms
+                )
+                compute_service_descriptor[1]["attributes"]["main"]["privacy"] = {
+                    "publisherTrustedAlgorithms": trusted_algorithms_list
+                }
+        self.update(asset=asset, publisher_wallet=publisher_wallet)
```
mariacarmina, line 357: Not sure here, when I call `update` function if I used the correct object for `asset`. Do I need to use `self` for asset field instead of the parameter `asset`?

mariacarmina (commented):
I think in `barge`, the logs for `aquarius` look good, but I don't understand the error of `NoneType` that I get when I run `pytest`, because I tried to see if there is a `NoneType` object, but nothing seems to be wrong where the actual error points. I will attach the screenshots of my local run.
aqua
failure

ocean_lib/ocean/ocean_assets.py
```diff
@@ -113,6 +123,7 @@ def create(
         owner_address: str = None,
         data_token_address: str = None,
         provider_uri=None,
+        trusted_algorithms=None,
```
alexcos20, line 126: Please, do not add custom parameters to create !
If a service requires custom paramaters, add them when you create that service.
look at https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts#L258

if next week we are going to add a service.type = XXX , are we going to change asset.create again?

alexcos20 (changes requested):

ocean_lib/ocean/ocean_assets.py
mariacarmina: I think @calina-c  did this part to help me and I chose to do the `update` part. Did you find something suspicious at the error @alexcos20?

mariacarmina: I will take a look on create part too.

calina-c: @alexcos20 It is true, I added this part because this is what I understood from the call we had 3-4 weeks ago with you and Samer. I asked for a review on my initial changes [(the commit)](https://github.com/oceanprotocol/ocean.py/pull/139/commits/6d15e58d034af20bda400848a2af2438ea2f2d20) and whether this was the right direction to go, but I had no answer then.

So, basically, when testing, should we create the asset without the extra parameter, and apply the update_trusted_algorithms to it? Would that work?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -221,6 +221,7 @@ def start_compute_job(
         output: dict = None,
         input_datasets: list = None,
         job_id: str = None,
+        raw_response: bool = False,
```
ssallam, line 224: I think it's better to not add unnecessary argument. We can just always return the `raw_response`. then the caller can take the job_id from that. Of course we have to update all existing calls to this method.

ocean_lib/ocean/ocean_compute.py
```diff
@@ -192,6 +192,7 @@ def start(
         algorithm_data_token: str = None,
         output: dict = None,
         job_id: str = None,
+        raw_response: bool = False,
```
ssallam, line 195: same as above, don't need additional parameter, just return the raw response always

ssallam: but this would be a breaking change

mariacarmina: That's why I didn't want to cause another breaking change and I add an additional parameter instead.

ocean_lib/ocean/ocean_assets.py
ssallam: Alex's comment is still valid, we shouldn't add more parameters to this `assets.create`.
Instead, the trusted algorithms can be added to the attributes in the service descriptor of the compute service. 
This means that the caller has to create a computer service descriptor to be able to set trusted algorithms. Otherwise, the asset can be updated with the trusted algorithms list after publishing.

tests/ocean/test_compute_flow.py
```diff
@@ -217,3 +241,101 @@ def test_compute_multi_inputs():
...
+    _ = setup.publisher_ocean_instance.assets.resolve(compute_ddo.did)
+    compute_ddo.update_trusted_algorithms(trusted_algorithms=[algorithm_ddo.did])
```
ssallam, line 270: The list of trusted algorithms is not just the algorithm did. Need to have the other hash values that are created in the helper function.

alexcos20 (approved):

PR #143, opened 2021-02-22
fix optional argument in pay_for_service

### LINKED ISSUE #141: quickstart fails using pay_for_service(); recent changes broke backwards compatibility
**Describe the bug**

In mid Jan I revised the quickstarts to help make them easier to use. They all worked then. 

Today, I decided to see how they do now. Everything went fine until step 7 of the final quickstart flow - the marketplace quickstart. 

**To Reproduce**
1. Go to marketplace quickstart README
2. Go through the steps. It will fail at step 7 with:
```python
>>> order_tx_id = bob_ocean.assets.pay_for_service(
...     quote.amount, quote.data_token_address, asset.did, service.index, fee_receiver, bob_wallet)
TypeError: pay_for_service() missing 1 required positional argument: 'consumer'
```

**Expected behavior**
1. It doesn't fail. I can complete my quickstart.
2. It doesn't break the interface promise of ocean.py

More on (2): this is a breaking change to the ocean.py interface. Obviously this is a big no-no; we should not do breaking changes to the interface unless it's absolutely necessary as part of a bug fix, or we're going to a new version (rare). In this case it's clear that neither were the case.

**More info**

[Here's the file in mid-Jan](https://github.com/oceanprotocol/ocean.py/blob/5e1a29393f4cacc06c38af05f942cb291192a25a/ocean_lib/ocean/ocean_assets.py). Note in `pay_for_service()` that it does not have a parameter `consumer`.

[Here's the status quo file](https://github.com/oceanprotocol/ocean.py/blob/master/ocean_lib/ocean/ocean_assets.py). Note  in `pay_for_service()` that it *does* not have a parameter `consumer`. According to the docstring it's supposed to default to the payer. But it only does that if `None` is specified explicitly specified.
```python
 def pay_for_service(
...
        from_wallet: Wallet,
        consumer: str,
    ) -> str:
        """
...
        :param from_wallet: Wallet instance
        :param consumer: str the address of consumer of the service, defaults to the payer (the `from_wallet` address)
...
        """
...
        if consumer is None:
            consumer = from_wallet.address
...
```
   
The error was introduced in [this commit](https://github.com/oceanprotocol/ocean.py/commit/a6a76dc38685c55c44cdea7d46a3758725e23ae5#diff-b085f9c75f577fac9b35d97d552e9d70b034711f7aab57279bd0cd72412c3446) of Feb 8.

**TODOs to fix this issue**
- [x] Change the quickstart so at least *it's* not broken before the other stuff is done
- [x] Write / ensure there's a unit test to capture behavior when `consumer` is specified and non-`None`. It should pass.
- [x] Write / ensure there's a unit test to capture behavior when `consumer` is specified and `None`. It should pass.
- [x] Write a unit test "C" for when `consumer` is not specified at all. It should fail.
- [x] Fix the bug. Now all three tests should pass. Commit.
- [ ] Deploy new version of ocean.py

alexcos20:
Closes #141.

### DIFF
--- a/READMEs/marketplace_flow.md
+++ b/READMEs/marketplace_flow.md
@@ -35,7 +35,7 @@ npm install
 npm start
 ```
 
-Finally, check out the market app as a webapp, at `http://localhost:8000`. 
+Finally, check out the market app as a webapp, at http://localhost:8000. 
 
 ## 2. Alice publishes data asset (including metadata)
 
--- a/READMEs/marketplace_flow.md
+++ b/READMEs/marketplace_flow.md
@@ -259,7 +259,7 @@ service = asset.get_service(ServiceTypes.ASSET_ACCESS)
 #order the asset, and send over the datatoken
 quote = bob_ocean.assets.order(asset.did, bob_wallet.address, service_index=service.index)
 order_tx_id = bob_ocean.assets.pay_for_service(
-    quote.amount, quote.data_token_address, asset.did, service.index, fee_receiver, bob_wallet)
+    quote.amount, quote.data_token_address, asset.did, service.index, fee_receiver, bob_wallet, None)
 print(f"order_tx_id = '{order_tx_id}'")
 ```
 
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -444,7 +444,7 @@ class OceanAssets:
         service_id: int,
         fee_receiver: str,
         from_wallet: Wallet,
-        consumer: str = None
+        consumer: str = None,
     ) -> str:
         """
         Submits the payment for chosen service in DataTokens.
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -444,7 +444,7 @@ class OceanAssets:
         service_id: int,
         fee_receiver: str,
         from_wallet: Wallet,
-        consumer: str,
+        consumer: str = None
     ) -> str:
         """
         Submits the payment for chosen service in DataTokens.
--- a/tests/ocean/test_market_flow.py
+++ b/tests/ocean/test_market_flow.py
@@ -6,6 +6,7 @@ import os
 from ocean_lib.assets.asset import Asset
 from ocean_utils.agreements.service_agreement import ServiceAgreement
 from ocean_utils.agreements.service_types import ServiceTypes
+from tests.resources.ddo_helpers import get_metadata, get_registered_ddo
 from tests.resources.helper_functions import (
     get_another_consumer_ocean_instance,
     get_another_consumer_wallet,
@@ -15,7 +16,6 @@ from tests.resources.helper_functions import (
     get_publisher_wallet,
     mint_tokens_and_wait,
 )
-from tests.resources.ddo_helpers import get_registered_ddo, get_metadata
 
 
 def test_market_flow():
@@ -62,7 +62,6 @@ def test_market_flow():
         service.index,
         "0xF9f2DB837b3db03Be72252fAeD2f6E0b73E428b9",
         consumer_wallet,
-        consumer_wallet.address,
     )
     ######
     # Download the asset files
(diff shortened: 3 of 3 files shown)


calina-c (approved):

trentmc:
There is a checklist in the issue itself (#141). Not all items in this checklist are complete. When they are (except deploy), I will be happy to approve. But not in the meantime. We *need* to improve our game in SW engineering.

trentmc (changes requested):
Needs to tick off all the boxes described in the ticket before this can be approved. Currently it doesn't. Eg needs more unit tests.

alexcos20:
I agree.   market tests should cover the missing consumer parameter

trentmc:
> No need for the 3rd test

Nope. We should test all three cases, in ocean.py, as the ticket describes: non-`None` input, `None` input, no input at all (for value of `consumer` parameter).

> market tests should cover the missing consumer parameter (user is the consumer) (missing consumer and consumer=None)

This is a good idea. To be clear, is this for `market` repo?

> compute tests should cover the additional consumer parameter (because Ctd is the consumer, and not the user)

This is a good idea too. What repos do you mean for this? ocean.py, provider.py, other?

alexcos20:
no, I was referring to:
- market tests (test_market_flow.py) , which should handle no consumer and consumer = None  scenario
- compute tests (test_compute_flow.py) should handle consumer = address (not None).

calina-c:
> * market tests (test_market_flow.py) , which should handle no consumer and consumer = None  scenario
test_market_flow seems to have a consumer, but it is the same as the consumer wallet address. https://github.com/oceanprotocol/ocean.py/blob/fcf28053193deaa3a3b228e2587bbb2b1efc0bb7/tests/ocean/test_market_flow.py#L65

My suggestion is to consider the consumer as the wallet.address if the address param is None. I'll adjust the tests accordingly.

> * compute tests (test_compute_flow.py) should handle consumer = address (not None).
This seems fine: https://github.com/oceanprotocol/ocean.py/blob/fcf28053193deaa3a3b228e2587bbb2b1efc0bb7/tests/ocean/test_compute_flow.py#L98

alexcos20:
It's already handled here: https://github.com/oceanprotocol/ocean.py/blob/aaf9691fc3162b60f05a331bd608c683ba40bcd9/ocean_lib/ocean/ocean_assets.py#L477

calina-c:
@trentmc tested with and without Alex's fix. All results are as expected and unit tests cover the cases for parameter missing (test market flow) and existing (test compute flow). Without Alex's fix, the new test fails, as expected.

I agree there is no need for the 3rd unit test (explicitly passing `None`) since absent and None are now equivalent through default parameters.

trentmc:
> I agree there is no need for the 3rd unit test (explicitly passing None) since absent and None are now equivalent through default parameters.

Yes it's like that now that the fix is done. But it wasn't like that before the fix. Hence why you need all three tests. I will continue to be adamant about this, because with every issue we should consider it essential to have a test that captures the breakage.

> @alexcos20 I agree no need for a major version, but shouldn't we bump the minor one at least?

I agree too that no need for a major version. I always saw it this way; sorry if I wasn't clear in the initial description.

It should be considered a small bugfix release. Not E.g. in "x.y.z" it should be a change to the z.

alexcos20:
yes, just bump patch version   (./bumpversion.sh patch)

calina-c:
@trentmc I added the explicitly None test. Let me know if it needs anything else. I'll add the bump after confirming.

trentmc:
> @trentmc I added the explicitly None test. 

Thank you.

> Let me know if it needs anything else.

The tests violate the DRY principle. That is, too much repeated code. It's gonna be hard to maintain. There should be some way to test for this with just a few lines of code. Some possible ways:
1. parameterize `test_market_flow` (it's a trick in pytest). 
1. build up some of object values in conftest.py, then have different unit tests that check various settings
1. create a new directory `tests/models/assets` and start to flesh out proper unit tests for the `assets` module. It's honestly really silly that the only place we tests assets is on a system-ish level test.

The best option is (3). But that's a lot of work for this issue. My suggestion: do (1) for this. I just created #149 for (3).

PR #183, opened 2021-03-02
Codacy fixes

### LINKED ISSUE #174: Get all codacy issues passing

Codacy does linting other tests. 
- It auto-runs remotely. 
- And we can now run locally too, as of #164 :) [Here's the PR](https://github.com/oceanprotocol/ocean.py/pull/169/files).
 
[This Codacy ocean.py dashboard](https://app.codacy.com/gh/oceanprotocol/ocean.py/dashboard) reports *many* issues in ocean.py.
- One subset is from [this PR on developers.md](https://github.com/oceanprotocol/ocean.py/pull/145) leading to [these issues](https://app.codacy.com/gh/oceanprotocol/ocean.py/pullRequest?prid=6996426). 
- But that's just a subset. We should get everything to work.

- [ ] pass all codacy tests, both local *and* remote

calina-c:
Towards fixing #174 (but not completely there yet).


trentmc (changes requested):
Please merge in most recent changes from trunk. Then ping me for another review. Thanks!

tests/ocean/test_ocean_assets.py
```diff
@@ -93,7 +93,6 @@ def _get_num_assets(_minter):
     assert ddo.metadata["main"]["name"] == _name
     with pytest.raises(ValueError):
         ocn.assets.update(ddo, bob)
-        # assert False, "this asset update should fail, but did not."
```
trentmc: Why is this removed?

Perhaps we should go in the other direction: uncomment the line and to revive the effectiveness of the test?

calina-c: pytest.raises will take care of that, I've tested it both ways. If a ValueError is not thrown in this block, the test fails.

calina-c: and it's consistent with previous blocks. This was the only one done using try/except.

README.md
```diff
@@ -44,19 +44,19 @@ This is in beta state and you can expect running into problems. If you run into
 
 ### Simple Flow
 
-[Publish your first datatoken](READMEs/datatokens_flow.md) - connect to Ethereum, create an Ocean instance, and publish.
```
trentmc, line 47: While codacy wants us to go to files with `-` instead of `_`, and therefore requiring `\` handler before every `-`. This makes the code _harder_ to read, the opposite of what codacy is supposed to achieve. 

Also note that python convention itself is for .py filenames to be snake case (include `_`) rather than camel case.

We should choose for ourselves, which we want:
1. Write a Codacy instruction to ignore the rule. Pro: better readability, still in line with py 
2. Follow this Codacy rule of `_` for .md files. Pros/cons: opposite of above

Thoughts?

calina-c: while python recommends snake case, .py files are not "linked" as an URL anywhere. These changes affect only markdown files, and are recommended by Codacy, but also here: https://github.com/bartvandebiezen/file-name-conventions/blob/master/README.md

see:
```
Never use spaces or underscores. Spaces are converted to %20 in URLs or can break an URL when shared. Underscores are difficult to see when the file name is displayed as an underlined link. Although the use of underscores does not impact your ranking that much, Google advices not to use underscores.
```

Also, the Google link from that readme: https://www.youtube.com/watch?v=AQcSFsQyct8

Markdown file names should conform to "linkable"/SEO-compliant etc. rules, not Python executable/source code file names.

calina-c: Ref. SEO -> if we want a search "get tokens OCEAN" to point to our github docs on getting ocean tokens, "get-tokens-OCEAN.md" is better than "get_tokens_OCEAN.md"

trentmc: Thanks very much for this info, these are great points! OK this is fine then.

ocean_lib/models/bpool.py
```diff
@@ -426,15 +426,9 @@ def exitswapExternAmountOut(
     def totalSupply(self) -> int:
         return self._ccontract.totalSupply()
 
-    def balanceOf(self, whom_address: str) -> int:
```
trentmc, line 429: Why is `balanceOf()` and `approve()` removed? 

Notes: BPool class has ERC20 functionality, starting at the line `# ==== Balancer Pool as ERC20`. I implemented them here rather than relying on its parent class BToken's implementation, on purpose.

calina-c: It was the exact same code as in the parent class, only different variable names, so I removed the lines. Should I add them back?

trentmc: It's not the exact same code: they have different variable names for the concise contract. `_ccontract` vs `contract_concise`.

Yes, add them back please.

There is probably a small opportunity for cleanup here. But it's a bit dangerous, so I recommend keeping them apart for now. This code has been pretty rock solid with few changes since release, lots of unit tests, and quite clean, so I don't see much need to fiddle with it.

One final thought: the unit tests should have picked up an error after you deleted the lines. Did they not? If no, it points to that we need better tests yet. (And we know that! Right now even though this is the most thorough testing in all of ocean.py, they're still only mostly happy-path tests.)

trentmc (approved):

Issue #113, opened 2021-01-19
development instructions are super confusing


kremalicious:
**Problem**

The current instructions provide a terrible onboarding experience for both (a) devs just using ocean.py and (b) devs working to improve ocean.py code itself (in developers.md). 

Use case: I'm an external developer and want to do a tiny code change and want to test my change locally. developers.md tells me to clone 3 repos and in the end no tests can run locally.

For testing flow, compare with this for inspiration https://github.com/oceanprotocol/ocean.js#-testing.

**This needs:**

For developers just using ocean.py (via import):
- [x] Get them to successfully go through "simple" and "marketplace" quickstarts in a streamlined fashion as possible. It's ok to use barge here. (How: use barge for both, and move the other stuff until after.) [Done in https://github.com/oceanprotocol/ocean.py/commit/a246fe4ee5515bb63d457660df9e71554161678b]
- [x] And then provide them with ways to understand how to do all the things towards their own app's production usage. That is: work with networks & services that aren't just local, e.g. to use rinkeby & provider/aquarius for rinkeby. And, use test OCEAN. And, understand how config params are set. [Already done via the "learn more" section]

For developers working to improve the ocean.py code itself:
- [x] for developers of ocean.py, figure out most minimal instructions to get tests running locally
- [x] The current development instructions provide 2 paths. Make it 1.
- [x] 1 simple way of running tests locally against locally running components. Please choose `barge` and not manual cloning/copying
- [x] same behavior between instructions and Travis (`tox` on Travis, `pytest` locally, why the difference?)

----
Related: #181: 'example' folder doesn't do much; either delete or flesh out

kremalicious:
@alexcos20 @ssallam let's start by first figuring out the most basic way to run the tests locally. What else is required to make this succeed?

```
# 1st terminal
git clone https://github.com/oceanprotocol/barge
cd barge
./start_ocean.sh --no-dashboard

# 2nd terminal
tox
```

@mariacarmina what do you get when you run this?

trentmc:
I'm glad to see this as an issue 👍👍

mariacarmina:
> @alexcos20 @ssallam let's start by first figuring out the most basic way to run the tests locally. What else is required to make this succeed?
> 
> ```shell
> # 1st terminal
> git clone https://github.com/oceanprotocol/barge
> cd barge
> ./start_ocean.sh --no-dashboard
> 
> # 2nd terminal
> tox
> ```
> 
[...]

PR #184, opened 2021-03-02
issue113 dev instrs

trentmc:
Towards #113. Focus on user of ocean-lib, versus an improver of it.

Changes proposed in this PR:
- WIP, please don't review yet :)


trentmc:
Notes from initial commit:

First cut improving main README (reorg) and datatokens_flow.md. Just starting on marketplace_flow.md. However, errors occur in datatokens_flow.md, I think this has uncovered bugs, I need to resolve them before I go further here.

trentmc:
Here's an annotated version of `contract_handler.py::_load`:
```python

    @staticmethod
    def _load(contract_name, address=None):
        """Retrieve the contract instance for `contract_name` that represent the smart
        contract in the ethereum network.

        :param contract_name: str name of the solidity smart contract.
        :param address: hex str -- address of smart contract
        :return: web3.eth.Contract instance
        """        
        assert (
            ContractHandler.artifacts_path is not None
        ), "artifacts_path should be already set."
        contract_definition = ContractHandler.read_abi_from_file(
            contract_name, ContractHandler.artifacts_path
        )

        #Debugging below is
        # from the modified datatokens readme,
        # final line, calling ocean.create_data_token
        # Before it gets here, I get a log message:
        # "WARNING:root:No logs where found for tx 0xe62c75b18ecdfd0f928eef3564f3573208c37c2e519cc1b456db6217a4d5b6bd."

        #Then I only got PDB to stop when
        # contract_name = 'DataTokenTemplate' (not 'DTFactory')
        # 

        # (Pdb) address
        # ''
        # (Pdb) "address" in contract_definition
        # False
                    
        if not address and "address" in contract_definition: #False, so skipped
            address = contract_definition.get("address")
            assert address, "Cannot find contract address in the abi file."
            address = Web3.toChecksumAddress(address)

        # (Pdb) address
        # ''
        
        contract = Web3Provider.get_web3().eth.contract(
        )

        #(Pdb) ContractHandler._contracts[contract_name]
        #*** KeyError: 'DataTokenTemplate'

        ContractHandler._set(contract_name, contract)

        # (Pdb) ContractHandler._contracts[(contract_name, address)]
        # *** KeyError: ('DataTokenTemplate', '')
        # (Pdb) ContractHandler._contracts[contract_name]
        # (<class 'web3.utils.datatypes.Contract'>, <web3.contract.ConciseContract object at 0x7fbb3f777c70>)
        
        # Here's the bug!! The return value is always the ordinal-2 key,
        # but if address is still '' then it should use ordinal-1 key
        #
        # That will fix the error here. However the final line of quickstart:

        # >>> print(f"created datatoken: done. Its address is {datatoken.address}")
        # created datatoken: done. Its address is None

        # So maybe there are two bugs.
        # (1) Return value should handle when address=''.
        #     The fix for one is below (or assert != ''?)
        #     Either way, I should write a unit test. 
        # (2) Why is address='' in the first place, and warning "No logs where found for tx .." ?

        if address == '': #FIXED with if/else here
            return ContractHandler._contracts[contract_name]
        else:
            return ContractHandler._contracts[(contract_name, address)]
```

PR #190, opened 2021-03-03
Codacy fixes 2nd

calina-c:
Towards #174.

### DIFF
--- a/examples/__init__.py
+++ b/examples/__init__.py
@@ -3,4 +3,5 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+"""Initialises examples package."""
 from .example_config import ExampleConfig  # noqa
--- a/ocean_lib/assets/__init__.py
+++ b/ocean_lib/assets/__init__.py
@@ -1,4 +1,0 @@
-#
-# Copyright 2021 Ocean Protocol Foundation
-# SPDX-License-Identifier: Apache-2.0
-#
--- a/ocean_lib/ocean/__init__.py
+++ b/ocean_lib/ocean/__init__.py
@@ -1,4 +1,0 @@
-#
-# Copyright 2021 Ocean Protocol Foundation
-# SPDX-License-Identifier: Apache-2.0
-#
--- a/ocean_lib/web3_internal/web3_overrides/contract.py
+++ b/ocean_lib/web3_internal/web3_overrides/contract.py
@@ -19,7 +19,7 @@ class CustomContractFunction:
         Customize calling smart contract transaction functions to use `personal_sendTransaction`
         instead of `eth_sendTransaction` and to estimate gas limit.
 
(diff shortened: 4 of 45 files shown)


ocean_lib/assets/__init__.py
```diff
@@ -1,4 +0,0 @@
-#
```
trentmc, line 1: Why was this deleted? I know that for pytest (and maybe other things) you need an __init__.py file.

ocean_lib/ocean/__init__.py
```diff
@@ -1,4 +0,0 @@
-#
```
trentmc, line 1: There are unit tests as subdirectories of this. So I'm pretty sure that this one is needed in order for pytest. (But I could be wrong! Maybe check)

examples/__init__.py
```diff
@@ -3,4 +3,5 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+"""Initialises examples package."""
```
trentmc, line 6: What's with all the "initialize" and "initializes" comments? Redundant because it's obvious imo. Therefore violates DRY

ocean_lib/web3_internal/web3_overrides/contract.py
```diff
@@ -11,13 +11,15 @@
...
         self._contract_function = contract_function
 
     def transact(self, transaction=None):
         """
         Customize calling smart contract transaction functions to use `personal_sendTransaction`
-        instead of `eth_sendTransaction` and to estimate gas limit. This function
-        is largely copied from web3 ContractFunction with important addition.
+        instead of `eth_sendTransaction` and to estimate gas limit.
+
+        This function is largely copied from web3 ContractFunction with important addition.
```
trentmc, line 22: "with" -> "with an"

trentmc (changes requested):
The comments give specific suggestions. Big picture: (a) why delete the __init__.py's (I am somewhat certain they are needed for testing) (b) why a bunch of redundant comments like "initialize __" ?

calina-c:
@trentmc:
(a) not all, just the empty ones. I actually wanted to discuss with the team. An empty__init__.py is not required anymore to signal namespaces. (since Python 3.4, and we are using 3.8 on all repos). However, is there alternately something else that could break?
(b) added the initialise docstrings because Codacy was complaining about missing docstrings for __init__ function. Even if it's trivial just saying that it initialises the object, it's better to have one. i.e. for documentation generators, IDEs that plug into the docstrings etc. I'd rather add a few trivialities and encourage docstrings. Or we can work through the list and add more meaningful documentation.

trentmc (approved):

calina-c:
For extra safety reasons, on Samer's suggestions, I will run a local build to check the imports.

calina-c:
Seems ok. Imports work.

PR #193, opened 2021-03-04
More Codacy Fixes (part 3)

calina-c:
Towards #174.


calina-c:
We should add D213 and D211 to the ignore list, but it is not possible yet. I will experiment on another branch.

bumpversion.sh
```diff
@@ -30,12 +30,12 @@ fi
 
 if [[ $2 == '--tag' ]]; then
   if git branch --contains $(git rev-parse --verify HEAD) | grep -E 'main'; then
-    bumpversion --tag --commit $1
```
trentmc, line 33: Did you test that bumpversion.sh works with these changes? Since it won't show up in unit tests

trentmc (approved):

PR #198, opened 2021-03-05
Codacy batch 4

calina-c:
Towards #174.


trentmc (dismissed):

trentmc (approved):
I don't really know what I'm looking at in the prospector code, but I know the context due to the slack conversation so I will approve here:) And I've already approved the other code.

PR #215, opened 2021-03-10
Complete the file_path returned from assets.download()

### LINKED ISSUE #142: file_path returned from assets.download() is not complete
**Describe the bug**
When making a call to `ocean.assets.download()`, I do not get the full path of the file that I just downloaded. I get everything except for the last part, e.g. the part with `datafile.0x90ba155..`.

**To Reproduce**

Fortunately it's easy! It appears in step 7 of the marketplace quickstart.

1. Go to marketplace quickstart
2. Do it all, until you finish the line `print(f"file_path = '{file_path}'") #e.g. datafile.0xAf07...`
3. Look at the printout. It doesn't include the part `datafile.0xAf07...`. 

Here's an example of my interactive session and printout, in Python console:
```python
>>> file_path = bob_ocean.assets.download(
...     asset.did, 
...     service.index, 
...     bob_wallet, 
...     order_tx_id, 
...     destination='./' 
... )
signing message with nonce 0: did:op:90ba155fe16404b17dd770f704C2273B9228E847, account=0x15f8a84B184A62bD3fC5Ad575F85560C72BAFB24
>>> print(f"file_path = '{file_path}'") #e.g. datafile.0xAf07...
file_path = '/home/trentmc/code/test1'
>>> file_path
'/home/trentmc/code/test1'
```

**Expected behavior**
`file_path` should have included `datafile.0x...`. 

For example, in Python console:
```python
 >>> file_path
'/home/trentmc/code/test1/datafile.0x90ba155fe16404b17dd770f704C2273B9228E847.3
```

And, inside that path, there should be the file. (This *is* the case already, we just have to make sure that the bug fix maintains this~). For example, in console:
```console
cd /home/trentmc/code/test1/datafile.0x90ba155fe16404b17dd770f704C2273B9228E847.3
cat branin.arff #should output the branin file just downloaded
```

**To fix**

- [ ] Write a unit test to reproduce the bug. When it runs, the test should fail
- [ ] Fix the bug, merge in the PR
- [ ] Deploy new version to pip

mariacarmina:
Fixes #142  .

Changes proposed in this PR:

- complete the file path returned from `asset_downloader`


ocean_lib/models/data_token.py
```diff
@@ -262,7 +262,7 @@ def verify_order_tx(self, web3, tx_id, did, service_id, amount_base, sender):
             target_amount = target_amount - order_log.args.marketFee
 
         # verify sender of the tx using the Tx record
-        tx = web3.eth.getTransaction(tx_id)
+        tx = web3.eth.getTransaction(tx_id) ###  TODO
```
trentmc, line 265: What does the "TODO" mean here?

ocean_lib/assets/asset_downloader.py
```diff
@@ -84,4 +84,4 @@ def download_asset_files(
             i,
         )
 
-    return destination
+    return asset_folder, destination ### verify if asset_folder is a file => append to destination
```
trentmc, line 87: Ideally we don't change the interface, because that's a breaking change to the API. For that you normally need a major version change. Why not just give the path promised by the interface?

ocean_lib/assets/asset_downloader.py
```diff
@@ -55,7 +55,7 @@ def download_asset_files(
     if not os.path.exists(destination):
         os.mkdir(destination)
 
-    asset_folder = os.path.join(
+    asset_folder = os.path.join( ### TODO
```
trentmc, line 58: Why "TODO"?

calina-c:
Hi all! Do you think you guys can make some improvements to the code coverage as well? The file you worked on, asset_downloader.py, has a 78% coverage, and we'd like to bring it up to at least 90%. Would you like to cover some more cases, since you are working on this?

To get the code coverage, run a full `tox` then `coverage report -m -i`. It will give you the coverage percentage for each file and the list of uncovered lines. You can work out from those missing lines which test cases are not covered. Add the 
needed tests, re-run and re-check. You should be good to go.

Tests also give you insight into what needs to be modified, e.g. one of the lines I see are untested (on my branch) is the creation of the directory itself, which happens when it doesn't exist. Maybe that is why it's not creating the dir, and returns a parent dir instead?

Anyway, feel free to leave this work for me if you think that is best. Let me know if you want to handle it, or whether you want me to do this when you're done here.

calina-c (approved):
Clean and elegant. I like it, good job!

mariacarmina:
> Hi all! Do you think you guys can make some improvements to the code coverage as well? The file you worked on, asset_downloader.py, has a 78% coverage, and we'd like to bring it up to at least 90%. Would you like to cover some more cases, since you are working on this?
> 
> To get the code coverage, run a full `tox` then `coverage report -m -i`. It will give you the coverage percentage for each file and the list of uncovered lines. You can work out from those missing lines which test cases are not covered. Add the
> needed tests, re-run and re-check. You should be good to go.
> 
> Tests also give you insight into what needs to be modified, e.g. one of the lines I see are untested (on my branch) is the creation of the directory itself, which happens when it doesn't exist. Maybe that is why it's not creating the dir, and returns a parent dir instead?
> 
> Anyway, feel free to leave this work for me if you think that is best. Let me know if you want to handle it, or whether you want me to do this when you're done here.

Sure @calina-c . Good practice is key.

ocean_lib/assets/test/test_asset_downloader.py
```diff
@@ -0,0 +1,103 @@
...
+            publisher,
+            "test_destination",
+            ddo.data_token_address,
+            "test_order_tx_id",
+            data_provider,
+            index
+        )
+
+
+def test_ocean_assets_download_destination_file(publisher_ocean_instance, metadata):
```
mariacarmina, line 85: I still have a uncovered part by this test (line `55` in `asset_downloader`). What am I missing here?

calina-c: the only missing test is the one where the os path does not exist, e.g. "examples/nofolderhere". I am satisfied with the current implementation, but if you want to add the extra test, that would be great. (btw please remember to delete the created folder at the end of the test, if you choose to implement it)

ocean_lib/models/data_token.py
```diff
@@ -262,12 +262,11 @@ def verify_order_tx(self, web3, tx_id, did, service_id, amount_base, sender):
             target_amount = target_amount - order_log.args.marketFee
 
         # verify sender of the tx using the Tx record
-        tx = web3.eth.getTransaction(tx_id)
         if sender not in [order_log.args.consumer, order_log.args.payer]:
             raise AssertionError(
                 "sender of order transaction is not the consumer/payer."
             )
-
+        tx = web3.eth.getTransaction(tx_id)
```
calina-c, line 269: why is this added here? I see no usage of it later.

ocean_lib/assets/test/test_asset_downloader.py
```diff
@@ -0,0 +1,102 @@
...
+            ddo,
+            publisher,
+            "test_destination",
+            ddo.data_token_address,
+            "test_order_tx_id",
+            data_provider,
+        )
+
+
+def test_ocean_assets_download_indexes(publisher_ocean_instance, metadata):
```
calina-c, line 37: Please add a docstring to this test as well.

calina-c (changes requested):
Just minor changes :) Thank you!

ocean_lib/models/data_token.py
mariacarmina: Saved some wrong modification that time. Now refactored it.

ocean_lib/assets/test/test_asset_downloader.py
mariacarmina: Actually the uncovered part is about using already an existing directory to place the downloads. And it is still uncovered :-1:

calina-c (dismissed):
Good job!

calina-c:
I tested locally, and we need to fix some gitignore stuff :)

mariacarmina (commented):
97% code coverage on `asset_downloader.py.`

calina-c (dismissed):

tests/resources/downloads/__init__.py
```diff
@@ -0,0 +1,5 @@
+#
```
trentmc, line 1: There's likely a cleaner way than this: use pytest built-in functionality for temporary directories.

Details: https://docs.pytest.org/en/stable/tmpdir.html

Then you won't need tests/resources/downloads, or adding the extra .gitignore line.

.gitignore
```diff
@@ -17,3 +17,4 @@ ocean_lib.egg-info/
 .tox/
 consume-downloads/
 coverage.xml
+tests/resources/downloads/datafile.*
```
trentmc, line 20: Likely won't need this, see my comment below about pytest functionality for temporary directories.

trentmc (changes requested):
Getting close! Hopefully my comments will help make the changes smaller yet.

tests/resources/downloads/__init__.py
calina-c: thank you, Trent! I suggested this implementation and did a similar thing on the 3rd code coverage batch. I've fixed it there. @mariacarmina sorry for misleading you, Trent's suggestion is much better.

trentmc (approved):

Issue #202, opened 2021-03-05
[EPIC] singletons for Web3Provider and more is surprising, non-standard, and won't work for multi-chain


trentmc:
**Is your feature request related to a problem? Please describe.**

Status quo: ocean.py has these singletons, via class-level attributes:
1. Web3Provider. Pointer to web3 provider: `ocean_lib/web3_internal/web3_provider.py:Web3Provider` with attribute `_web3` stored at _class_ level. Which makes it a singleton.
2. ContractHandler. The class `ocean_lib/web3_internal/contract_handler.py:ContractHandler` has attribute `_contracts` which caches contracts. It also has `artifacts_path`. 
3. ConfigProvider. The class  `ocean_lib/config_provider.py:ConfigProvider` with attribute `_config` which caches config params including pointer to the network. It's basically like having a bunch of global variables; though at least it's in one place.

Issues with this:
- It's non-standard in web3. The main web3 python tool, web.py, allows a separate object for each web3 connection. Same with brownie.
- The behavior is surprising. You don't really know that it's there unless you know. 
- It won't work for cross-chain, such as Eth mainnet + matic at once. And that is extremely useful behavior for Ocean as we support cross-chain, not to mention other dev users who want to do things like cross-chain arbitrage.
- It gets in the way of unit tests, because right now `./tests/conftest.py::setup_all()` instantiates a local singleton all the time. So you can't easily run tests for more specific behavior with these side effects happening

Note: we _do_ use the brownie pattern of passing in `wallet` as the last parameter into methods that do a blockchain tx. That's good! :) 

**Describe the solution you'd like**

Behave like web3.py and brownie. That is: each ocean = Ocean() object has its own connection to a chain. For example, a single py file with:
```python
ocean_matic = Ocean(matic_config)
ocean_ethmainnet = Ocean(ethmainnet_config)
```

General Strategy:
- Web3Provider: make it an object instance (not singleton)
- ContractHandler: make it an instance instance (not singleton)
- ConfigProvider: make it an instance, or move some of its attributes into instances (eg network url). We'll try different things and see what works best.
- Note: in sw eng, there's a natural tension of providing global variables for convenience, vs tighter scoping for maintainability. A useful middle ground is to have a "context" object that gets passed around (ref Osterhout). "Make it an object" above follows this. We could also have a _single_ object that holds each of these as sub-objects if we like. Let's see what looks best.

trentmc:
[...]

PR #222, opened 2021-03-12
Issue185 fail when addresses not specified

### LINKED ISSUE #185: create_data_token() fail when addresses not specified
## Describe the bug

The following configuration fails:
- using barge
- import ocean.py (vs git clone)
- no config.ini file, just `NETWORK_URL=ganache`. Note that location of addresses artifacts is _not_ specified, with the expectation that there be good defaults.
- trying to do simple quickstart functionality 

This was first discovered during work on #113, in [this PR](https://github.com/oceanprotocol/ocean.py/pull/184). With the configuration above, it fails on the first real usage of ocean.py functionality: the call to `ocean.create_data_token()`.

There are two issues:
1. It logs a warning: `WARNING:root:No logs were found for tx 0xe62c7...`
2. `contract_handler.py::_load()` gives an error: `KeyError: ('DataTokenTemplate', '')`, which seems to be caused by the `address` value staying at `''`. 

Issue (1) is a sign that it can't see the smart contract addresses. Issue (2) is likely a result of (1).

Related: 
- #205 create_data_token() fail when using default CONFIG_FILE value.
- [multi-repo#83](https://github.com/oceanprotocol/multi-repo-issue/issues/83) ocean-contracts is hard to use in python, including in ocean.py

## To Reproduce

(This is like developers.md, but replacing `git clone` of ocean.py to `pip install` of it.)

**Run the services.** In a new console:
```console
#grab repo
git clone https://github.com/oceanprotocol/barge
cd barge

#clean up old containers (to be sure)
docker system prune -a --volumes

#run barge with provider on
./start_ocean.sh  --with-provider2
```

**Install the library; set envvars; go into py.** In a new console:
```console
#Initialize virtual environment and activate it.
python -m venv venv
source venv/bin/activate

#Install the ocean.py library
pip install ocean-lib

#set envvars
export TEST_PRIVATE_KEY1=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58
export NETWORK_URL=ganache

#go into python
python
```

**Publish datatokens.** Within the py console:
```python
import os

private_key = os.getenv('TEST_PRIVATE_KEY1')
config = {'network': os.getenv('NETWORK_URL')}
ocean = Ocean(config)

print("create wallet: begin")
wallet = Wallet(ocean.web3, private_key=private_key)
print(f"create wallet: done. Its address is {wallet.address}")

print("create datatoken: begin.")
assert datatoken.address not in [None, '']
```

It fails on the last line. The second-last line gives a warning

## Expected behavior
It should create the datatokens successfully (no warnings or errors), while still using barge. 

**TODOs**

- [ ] Write a unit test to capture the KeyError issue in `contract_handler.py::_load()`; then update `_load()` to handle it better. 
- [ ] Get to the bottom of what's causing the warning; write a test; fix it
- [ ] Re-deploy

## Traceback-y stuff

Notes from initial commit:

First cut improving main README (reorg) and datatokens_flow.md. Just starting on marketplace_flow.md. However, errors occur in datatokens_flow.md, I think this has uncovered bugs, I need to resolve them before I go further here.

----
How I got the errors

* used the flow as described in the updated datatokens readme
* this used the most recent release of ocean-lib, which went into my local venv. 
* I debugged with pdb breakpoints in the venv. `~/code/play/venv/lib/python3.8/site-packages/ocean_lib/web3_internal/contract_handler.py`

Here's `contract_handler.py::_load` annotated to point out what I learned, and the main issues:
```python

    @staticmethod
    def _load(contract_name, address=None):
        """Retrieve the contract instance for `contract_name` that represent the smart
        contract in the ethereum network.

        :param contract_name: str name of the solidity smart contract.
        :param address: hex str -- address of smart contract
        :return: web3.eth.Contract instance
        """        
        assert (
            ContractHandler.artifacts_path is not None
        ), "artifacts_path should be already set."
        contract_definition = ContractHandler.read_abi_from_file(
            contract_name, ContractHandler.artifacts_path
        )

        #Debugging below is
        # from the modified datatokens readme,
        # final line, calling ocean.create_data_token
        # Before it gets here, I get a log message:
        # "WARNING:root:No logs were found for tx 0xe62c75b18ecdfd0f928eef3564f3573208c37c2e519cc1b456db6217a4d5b6bd."

        #Then I only got PDB to stop when
        # contract_name = 'DataTokenTemplate' (not 'DTFactory')
        # 

        # (Pdb) address
        # ''
        # (Pdb) "address" in contract_definition
        # False
                    
        if not address and "address" in contract_definition: #False, so skipped
            address = contract_definition.get("address")
            assert address, "Cannot find contract address in the abi file."
            address = Web3.toChecksumAddress(address)

        # (Pdb) address
        # ''
        
        contract = Web3Provider.get_web3().eth.contract(
        )

        #(Pdb) ContractHandler._contracts[contract_name]
        #*** KeyError: 'DataTokenTemplate'

        ContractHandler._set(contract_name, contract)

        # (Pdb) ContractHandler._contracts[(contract_name, address)]
        # *** KeyError: ('DataTokenTemplate', '')
        # (Pdb) ContractHandler._contracts[contract_name]
        # (<class 'web3.utils.datatypes.Contract'>, <web3.contract.ConciseContract object at 0x7fbb3f777c70>)
        
        # Here's the bug!! The return value is always the ordinal-2 key,
        # but if address is still '' then it should use ordinal-1 key
        #
        # That will fix the error here. However the final line of quickstart:

        # >>> print(f"created datatoken: done. Its address is {datatoken.address}")
        # created datatoken: done. Its address is None

        # So maybe there are two bugs.
        # (1) Return value should handle when address=''.
        #     The fix for one is below (or assert != ''?)
        #     Either way, I should write a unit test. 
        # (2) Why is address='' in the first place, and warning "No logs where found for tx .." ?

        if address == '': #FIXED with if/else here
            return ContractHandler._contracts[contract_name]
        else:
            return ContractHandler._contracts[(contract_name, address)]
```

trentmc:
Towards fixing #185. I couldn't easily fix everything just yet because there are some more fundamental issues that are better to solve first: multi-repo-issue#83, #202, #227. When those get done, I will revisit #185, hopefully by then it will become trivial:)

Changes proposed in this PR:
- when a datatoken is created via `ocean::create_data_token()`, it flags an error if the blockchain tx was not successful. (This greatly helps debugging. We will get more general via #227: make calls blocking.) 
- can disable call to `conftest.py::setup_all()` via a pytest fixture on that method, plus on a given unit test, have  `@pytest.mark.nosetup_all` decorator. This is what led to the update on pytest.ini.
- further split the unit test files for contract_handler.py.
- add explicit tests for #185: test_issue185_unit() and test_issue185_system().
- contract_handler.py documentation. Better document the attribute `_contracts`, and how 2 use cases of contract deployment are handled. (Deploy once, vs many times).
- contract_handler.py::_set() - refactor for readability, add preconditions and aggressive postconditions
- contract_handler.py::_load() - better document. Extra assert that address is not None by a certain point (catches badness). Extra assert that *contract.address* is not None (work around a web3.py bug). Do *not* return values, to better orthogonalize behavior. Aggressive postconditions to see everything is kosher
- contract_handler.py::_get() - refactor for readability, and update handling now that _load() doesn't return a value
-aggressive postconditions to see everything is kosher


ocean_lib/web3_internal/contract_handler.py
```diff
@@ -52,13 +62,42 @@ def set_artifacts_path(artifacts_path):
...
+        :return: tuple of (contract, concise_contract)
+        """
         if address:
-            return ContractHandler._contracts.get(
-                (name, address)
-            ) or ContractHandler._load(name, address)
-        return ContractHandler._contracts.get(name) or ContractHandler._load(
-            name, address
-        )
+            tup = ContractHandler._contracts.get((name, address))
```
calina-c, line 73: I think you can set an input_tup as different, only. e.g.
input_tup = (name, address) if address else (name)

and reuse the same contents from the if/else only with a different input. It would reduce duplication.

ocean_lib/web3_internal/contract_handler.py
```diff
@@ -136,14 +191,32 @@ def _load(contract_name, address=None):
...
+            contract.address = address
+        assert contract.address is not None
+
         ContractHandler._set(contract_name, contract)
-        return ContractHandler._contracts[(contract_name, address)]
+
+        # postconditions
+        name = contract_name
+        (contract1, concise_contract1) = ContractHandler._contracts[name]
+        assert contract1 is not None
```
calina-c, line 210: these assertions seem duplicated from _set(). Is there any way we can extract them into a function?

ocean_lib/web3_internal/test/test_contract_handler_load.py
```diff
@@ -0,0 +1,133 @@
...
+    setup_issue_185(monkeypatch)
+
+    # ensure that conftest::setup_all() was not called
+    assert ConfigProvider._config is None
+    assert Web3Provider._web3 is None
+    assert ContractHandler._contracts == dict()
+    assert ContractHandler.artifacts_path is None
+
+    # actual test. Imports only come now, to avoid setting class-level attributes
+    from ocean_lib.ocean.ocean import Ocean
```
calina-c, line 72: there has to be a workaround for this. isort automatically moves these to the top of the file and we'll have pre-commit issues. The simplest would be to set the isort skip flag, but I think this design is not that intuitive anyway. Maybe there is a way to reimport/refresh an imported class?

ocean_lib/web3_internal/test/test_contract_handler_load.py
```diff
@@ -0,0 +1,133 @@
...
+    ocean = Ocean(config)
+
+    wallet = Wallet(ocean.web3, private_key=private_key)
+
+    # this failed before the fix
+    datatoken = ocean.create_data_token("Dataset name", "dtsymbol", from_wallet=wallet)
+    assert datatoken is not None
+
+
+def setup_issue_185(monkeypatch):
```
calina-c, line 119: Can you please rename this to be more suggestive? I'm not hyped about using issue names inside the repo itself, but even if we do in comments, function names should not contain issue numbers, in my opinion.

pytest.ini
```diff
@@ -3,6 +3,10 @@
 filterwarnings =
     ignore::DeprecationWarning
 
+markers =
+    nosetup_all: do not call setup_all
```
calina-c, line 7: will this affect the entire codebase, or just when specifically asked for?

calina-c (changes requested):

ocean_lib/web3_internal/contract_handler.py
trentmc: Good point. Fixed in 0a5976a

ocean_lib/web3_internal/contract_handler.py
trentmc: Good point. Fixed in 83c7a5b

ocean_lib/web3_internal/test/test_contract_handler_load.py
trentmc: Personal opinion on isort: it causes more trouble than it's worth, everywhere. Eg we can't group imports according to if 3rd party vs our own. We should strongly consider removing it. (I still like black and _some_ code style checker like codacy. But isort, no.)

> Maybe there is a way to reimport/refresh an imported class?

I looked around. It's all super messy.

But we have a specific answer for our purposes: #202 "Web3 provider as singleton is surprising, non-standard, and won't work for multi-chain". This will solve it definitively. And the code to test issue185 will either be simpler in setup, or get deprecated entirely. Hence why _this_ PR is not a full solution to #185, just a partial one to ratchet up quality; and I'm going for #202 next (and #83).

Summary of recommendation here: ignore

pytest.ini
trentmc: > will this affect the entire codebase, or just when specifically asked for?

The latter. To invoke it for a given unit test, add decorator `@pytest.mark.nosetup_all` to the top of the unit test. (I will put that into the final squash-merged description too.)

ocean_lib/web3_internal/test/test_contract_handler_load.py
trentmc: Point taken.

I have added a docstring to give more context in a913fa4. 

Re function names with numbers, here's my thoughts:
- For non-test code: fully agree. Don't ever have issue names in the methods there
- For test code: it depends. For simple tests without much context needed, don't have. But for more complex issues that have eg a nuanced setup, it can be very helpful: rather than having 10 lines of context in the code, it points to the issue itself. That's what I did here. 

How does that sound?

calina-c (approved):

Issue #217, opened 2021-03-11
Improve ocean.py test coverage to 90%


calina-c:
Files that need more coverage:
- [x] ocean_lib/assets/asset_downloader.py -> handled as part of https://github.com/oceanprotocol/ocean.py/pull/215
- [x] ocean_lib/config.py -> got to 89% but some logic will change soon
- [x] ocean_lib/data_provider/data_service_provider.py
- [x] ocean_lib/example_config.py

## Models 
- [x] ocean_lib/models/algorithm_metadata.py
- [x] ocean_lib/models/bpool.py
- [x] ocean_lib/models/compute_input.py
- [x] ocean_lib/models/data_token.py
- [x] ocean_lib/models/dtfactory.py

## Ocean
- ocean_lib/ocean/constants.py -> to be left for further issues with typeChecking
- [x] ocean_lib/ocean/ocean_assets.py
- [x] ocean_lib/ocean/ocean_compute.py
- ocean_lib/ocean/ocean_pool.py -> soon to be deprecated per #239.

## Web3 Internal
- [x] ocean_lib/web3_internal/account.py
- [x] ocean_lib/web3_internal/contract_base.py
- [x] ocean_lib/web3_internal/event_filter.py
- [ ] ocean_lib/web3_internal/event_listener.py
- [x] ocean_lib/web3_internal/exceptions.py
- [x] ocean_lib/web3_internal/utils.py
- [x] ocean_lib/web3_internal/web3_overrides/contract.py
- [x] ocean_lib/web3_internal/web3_provider.py
- [x] ocean_lib/web3_internal/web3helper.py

PR #224, opened 2021-03-16
Issue 217 test coverage 2

calina-c:
Towards #217.

### DIFF
--- a/ocean_lib/models/data_token.py
+++ b/ocean_lib/models/data_token.py
@@ -304,20 +304,23 @@ class DataToken(ContractBase):
     def token_balance(self, account: str):
         return from_base_18(self.balanceOf(account))
 
+    def _get_url_from_blob(self, int_code):
+        try:
+            url_object = json.loads(self.blob())
+        except json.decoder.JSONDecodeError:
+            return None
+
+        assert (
+            url_object["t"] == int_code
+        ), "This datatoken does not appear to have a direct consume url."
+        return url_object["url"] if "url" in url_object else None
+
     def get_metadata_url(self):
         # grab the metadatastore URL from the DataToken contract (@token_address)
-        url_object = json.loads(self.blob())
-        assert (
-            url_object["t"] == 1
-        ), "This datatoken does not appear to have a metadata store url."
-        return url_object["url"]
+        return self._get_url_from_blob(1)
 
     def get_simple_url(self):
-        url_object = json.loads(self.blob())
-        assert (
-            url_object["t"] == 0
-        ), "This datatoken does not appear to have a direct consume url."
-        return url_object["url"]
+        return self._get_url_from_blob(0)
 
     # ============================================================
     # Token transactions using amount of tokens as a float instead of int
--- a/ocean_lib/models/data_token.py
+++ b/ocean_lib/models/data_token.py
@@ -17,6 +17,7 @@ from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_provider import Web3Provider
 from ocean_utils.http_requests.requests_session import get_requests_session
 from web3 import Web3
+from web3.exceptions import MismatchedABI
 from web3.utils.events import get_event_data
 from websockets import ConnectionClosed
 
@@ -38,8 +39,9 @@ class DataToken(ContractBase):
     MAX_MARKET_FEE_PERCENTAGE = 0.001
 
     def get_event_signature(self, event_name):
-        e = getattr(self.events, event_name)
-        if not e:
+        try:
+            e = getattr(self.events, event_name)
+        except MismatchedABI:
             raise ValueError(
                 f"Event {event_name} not found in {self.CONTRACT_NAME} contract."
             )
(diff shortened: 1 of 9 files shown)


ocean_lib/models/test/test_datatoken.py
```diff
@@ -33,13 +35,48 @@ def test_ERC20(alice_ocean, alice_wallet, alice_address, bob_wallet, bob_address
     assert from_base_18(token.balanceOf(alice_address)) == 98.0
     assert from_base_18(token.balanceOf(bob_address)) == 2.0
 
+    # test status functions
+    assert token.totalSupply() == 100_000_000_000_000_000_000
+    assert token.cap() == 1_000_000_000_000_000_000_000
+    assert token.datatoken_name() == "DataToken1"
+    block = alice_ocean.web3.eth.blockNumber
+    token_info = token.get_info(
```
trentmc, line 43: The name of the test is test_ERC20. To test ERC20 functionality.

get_info() is not ERC20 compliant. So it shouldn't be part of "test_ERC20", make it a new test.

BTW I just realized that a bunch of ocean.py datatoken interface isn't compliant :/ So I just created an issue for it. https://github.com/oceanprotocol/ocean.py/issues/228

For the purposes of _this_ PR, we don't have to fix that issue of course. To handle this comment, simply create another test method like "def test_get_info" and move its stuff there.

ocean_lib/models/test/test_datatoken.py
```diff
@@ -33,13 +35,48 @@ def test_ERC20(alice_ocean, alice_wallet, alice_address, bob_wallet, bob_address
...
+    token = alice_ocean.create_data_token(
+        "DataToken1", "DT1", alice_wallet, blob=json.dumps(blob_dict)
+    )
+    assert token.blob() == json.dumps(blob_dict)
+    assert token.get_simple_url() == "http://tblob/"
+
+    blob_dict = {"t": 1, "url": "http://tblob/", "foo": "bar"}
     token = alice_ocean.create_data_token(
-        "DataToken1", "DT1", alice_wallet, blob="foo_blob"
+        "DataToken1", "DT1", alice_wallet, blob=json.dumps(blob_dict)
```
trentmc, line 69: blob itself should work whether it's a dict or not.

I'd suggest to keep the original test as-is, and instead have another new test that tests a structured blob with the url.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -494,7 +494,6 @@ def pay_for_service(
                 amount_base,
                 from_wallet.address,
             )
-            return tx_hash
```
trentmc, line 497: This doesn't look like a "test coverage". And I don't see any context / motivation of why to change it.

I recommend putting this in a different PR. (Quickly create a new issue with the motivation, then branch, then put these changes in, merge it to main, then merge from main to this branch.)

ocean_lib/models/test/test_datatoken.py
calina-c: I have kept the original test, it's a few lines lower.

tests/integration/test_compute_flow.py
```diff
@@ -128,6 +129,16 @@ def run_compute_test(
     print(f"got job status after requesting stop: {status}")
     assert status, f"something not right about the compute job, got status: {status}"
 
+    if restart_and_result:
+        # TODO
```
trentmc, line 133: What's the time taken TODO this? Remember, there will be 10-100 more people reading this than your writing this. So if it just takes a few min to get this test going, I strongly recommend doing so.

trentmc (changes requested):
In general looks good. My comments point to suggested changes.

tests/integration/test_compute_flow.py
calina-c: no, as I explained on Slack it will take much longer, since a change in restart will most certainly break the interface and is more complicated. Restart does not work right now, so I can't test it.

trentmc: Thanks. I just saw the slack msg. OK. Then perhaps rather than say "TODO" comment it out and say "will be deprecated soon" or something.

ocean_lib/ocean/ocean_assets.py
calina-c: well, finishOrder is never used and never tested. The change here is just to delay the return until we finish the Order.

trentmc: OK to leave it here. Be sure to mention the "why" in your message in squash merge.

trentmc (dismissed):

tests/integration/test_compute_flow.py
calina-c: fyi I created this ticket as a follow-up: #232

ocean_lib/models/data_token.py
```diff
@@ -304,20 +306,23 @@ def download(self, wallet: Wallet, tx_id: str, destination_folder: str):
...
+    def _get_url_from_blob(self, int_code):
+        try:
+            url_object = json.loads(self.blob())
+        except json.decoder.JSONDecodeError:
+            return None
+
+        assert (
+            url_object["t"] == int_code
+        ), "This datatoken does not appear to have a direct consume url."
+        return url_object["url"] if "url" in url_object else None
```
ssallam, line 318: url_object.get("url") should do it

ocean_lib/ocean/ocean_assets.py
ssallam: please remove `finishOrder`, it is meant to be used by the provider, not the consumer.

calina-c: done!

ssallam (approved):
LGTM.

PR #233, opened 2021-03-17
Issue 217 test coverage 3

calina-c:
Towards #217.

This batch covers files under the `models` folder.

### DIFF
--- a/ocean_lib/models/dtfactory.py
+++ b/ocean_lib/models/dtfactory.py
@@ -36,10 +36,13 @@ class DTFactory(ContractBase):
     def get_token_minter(self, token_address):
         """Retrieves token minter.
 
-        This function is deprecated and only kept for maintainability purposes."""
+        This function will be deprecated in the next major release.
+        It's only kept for backwards compatibility."""
         from ocean_lib.models.data_token import DataToken  # isort:skip
 
-        return DataToken(address=token_address).contract_concise.minter()
+        dt = DataToken(address=token_address)
+
+        return dt.contract_concise.minter()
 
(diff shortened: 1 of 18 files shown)


ocean_lib/models/dtfactory.py
```diff
@@ -14,50 +13,36 @@ class DTFactory(ContractBase):
...
-            "TokenRegistered", event, filter_params, from_block=0, to_block="latest"
-        )
-        logs = event_filter.get_all_entries(max_tries=10)
-        for log in logs:
-            # assert log.args.tokenAddress == token_address
-            if log.args.tokenAddress == token_address:
-                return log.args.registeredBy
+        """Retrieves token minter.
+
+        This function is deprecated and only kept for maintainability purposes."""
```
trentmc, line 39: Suggest replace "This function .. purposes" with 
"This function will be deprecated in the next major release. It's only kept for backwards compatibility".

ocean_lib/models/dtfactory.py
```diff
@@ -14,50 +13,36 @@ class DTFactory(ContractBase):
...
+        from ocean_lib.models.data_token import DataToken  # isort:skip
 
-        return None
+        return DataToken(address=token_address).contract_concise.minter()
```
trentmc, line 42: Too many things going on in one line. Hurts readability. Suggested:
```python
dt = DataToken(address=token_address)
return dt.contract_concise.minter()
```

trentmc (changes requested):
Generally good! A couple small suggestions, then good to go.

trentmc (approved):

PR #237, opened 2021-03-18
Check if ocean.py is compatible with aqua 2.2.5

### LINKED ISSUE #236: Is ocean.py compatible with aquarius 2.2.5 ?
As we started rolling out aqua 2.2.5 (already deployed on ropsten), we should make sure that ocean.py is compatible)

alexcos20:
closes #236


calina-c:
superseded by #247

PR #243, opened 2021-03-19
Issue 200 - fake OCEAN

### LINKED ISSUE #200: For users of ocean-lib with ganache, getting fake OCEAN adds friction
I just streamlined the README quickstarts for simple flow and marketplace flow, as part (a) in #113. They're much simpler than before.

But the marketplace flow still has a specific pain point: getting fake OCEAN into ganache. To do so, the user needs to `git clone` the ocean.py repo then run `deploy.py ganache` 

It would be way better if we could remove that friction. Perhaps when barge starts, it not only gives ETH to those initial accounts, it also gives OCEAN (using `deploy.py`)?

Something more streamlined. 

**Candidate approaches**
1. in barge, have an extra option to run "python deploy.py ganache"
2. when ocean-contracts deploys, it does the core of what deploy.py does: (a) deploys a datatoken contract called OCEAN, and (b) updates the artifacts/address.json file (c) mint & transfer OCEAN tokens to the 2 test accounts
3. add (many) lines to the top of quickstart where Alice creates her own datatoken with symbol OCEAN, then sends some to Bob
4. add a single line to the top of quickstart like `create_fake_OCEAN()` and then ocean.py does all the work
5. do nothing

**Recommended approach**
We're going with 3 or 4, assuming complexity reduction is worth it compared to 5

trentmc:
Fixes #200. Specific changes:
- Rename command-line script `./deploy.py` to ` ./deploy_fake_OCEAN.py`. Deprecate anything that isn't for deploying OCEAN (it's obsolete).
- Move all functionality from `./deploy_fake_OCEAN.py` into a new module `./ocean_lib/ocean/deploy.py` with function `deploy_fake_OCEAN()`. 
- In marketplace quickstart, remove _20_-line section "Give Alice & Bob fake OCEAN", and replace with _2_-line section containing an import and a call to `deploy_fake_OCEAN()`
- And create a new test module for `deploy_fake_OCEAN()`


trentmc:
I merged sans review because I needed this out of my head to move on to other code. Balance of quality vs speed, but since I'm working on tickets for quality, it's all around optimizing quality :P

PR #247, opened 2021-03-22
Check aqua version compatibility.

### LINKED ISSUE #236: Is ocean.py compatible with aquarius 2.2.5 ?

calina-c:
Fixes #236.


ssallam (approved):
LGTM.

PR #263, opened 2021-04-06
Check files before publish.

### LINKED ISSUE #235: Check file info before publishing
While publishing an asset, ocean.py should use the fileinfo endpoint on provider, in order to make sure that your asset's files are reachable.  (ocean.js is already doing that)

calina-c:
Closes #235.

Changes proposed in this PR:

- check URLs are accessible before publishing metadata

### DIFF
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -11,6 +11,7 @@ import re
 from collections import namedtuple
 from json import JSONDecodeError
 
+import requests
 from ocean_lib.config_provider import ConfigProvider
 from ocean_lib.data_provider.exceptions import InvalidURLException
 from ocean_lib.enforce_typing_shim import enforce_types_shim
@@ -632,6 +633,14 @@ class DataServiceProvider:
                 f"Error invoking http method {method}: args={str(args)}, kwargs={str(kwargs)}"
             )
             raise
+
+    @staticmethod
+    def check_single_file_info(file_url, provider_uri=None):
+        _, endpoint = DataServiceProvider.build_fileinfo(provider_uri)
+        data = {"url": file_url}
+        response = requests.post(endpoint, json=data).json()
+        for file_info in response:
+            return file_info["valid"]
 
 
 def urljoin(*args):
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -10,8 +10,8 @@ import os
 import re
 from collections import namedtuple
 from json import JSONDecodeError
-
 import requests
+
 from ocean_lib.config_provider import ConfigProvider
 from ocean_lib.data_provider.exceptions import InvalidURLException
 from ocean_lib.enforce_typing_shim import enforce_types_shim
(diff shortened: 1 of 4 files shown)


ocean_lib/data_provider/data_service_provider.py
```diff
@@ -11,6 +11,7 @@
 from collections import namedtuple
 from json import JSONDecodeError
 
+import requests
```
trentmc, line 14: This should be part of the imports in the 3rd party libraries above. Eg remove the space above this line, and add a space below, and you're good.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -633,6 +634,14 @@ def _http_method(method, *args, **kwargs):
             )
             raise
 
+    @staticmethod
+    def check_single_file_info(file_url, provider_uri=None):
+        _, endpoint = DataServiceProvider.build_fileinfo(provider_uri)
+        data = {"url": file_url}
+        response = requests.post(endpoint, json=data).json()
```
trentmc, line 641: A "check" function should be doing reads, not writes. Yet by calling HTTP POST, it's doing a write. Why not use HTTP GET?

ocean_lib/ocean/ocean_assets.py
```diff
@@ -163,6 +166,13 @@ def create(
             logger.error(msg)
             raise ValueError(msg)
 
+        urls = [item["url"] for item in metadata["main"]["files"]]
+        for url in urls:
+            if not DataServiceProvider.check_single_file_info(url, provider_uri):
+                msg = f"Some files can not be accessed: {url}."
```
trentmc, line 172: It may be a file, or it may be another service. And, you only know right now that this _single_ service can't be accessed, not >1. Perhaps better phrasing: "The following service could not be accessed: {url}."

trentmc (changes requested):
3 comments: 2 small, one will take a bit more work but likely small too (HTTP POST->GET).

ocean_lib/data_provider/data_service_provider.py
calina-c: This was discussed on implementation.
1. We need a payload since sometimes we are requesting data on ddo's (encrypted), otherwise on raw URLs. This can not be achieved with GET as clearly as with POST.
2. This is not an idempotent request, which is a prerequisite for GET. We can not guarantee the same status of the file URL: it might be deleted, moved or otherwise inaccessible to the exterior (which is exactly what we are checking). As far as I know, POST is ok for requests that are not idempotent.

trentmc: OK! Thanks for the clear answer. When you do the squash & merge, please give the info you gave here, it will help.

ocean_lib/ocean/ocean_assets.py
trentmc: I just checked. It now says "The URL of this file can not be accessed: {url}".

That fixes my second concern. 

My first concern still needs addressing: It may be a file, or it may be another service.

trentmc (changes requested):
One thing left, need to generalize "file" -> "service"

trentmc (approved):

PR #271, opened 2021-04-09
fix issue 270 set_config first AssertionError in simple import 

### LINKED ISSUE #270: "set_config first" AssertionError in simple Ocean import
### Describe the bug
When trying to do

```python
```

I get:
```console
  File "/home/trentmc/code/ocean.py/ocean_lib/enforce_typing_shim.py", line 18, in enforce_types_shim
    if not ConfigProvider.get_config()["util"].getboolean("typecheck"):
  File "/home/trentmc/code/ocean.py/ocean_lib/config_provider.py", line 17, in get_config
    raise AssertionError("set_config first.")
```

### To Reproduce

Outline: 
- just like developers.md: (1) install dependencies, (2) run the services, (3) set up contracts like in developers.md. 
- then (4) try to do a couple imports, and it will fail.

Details:

**1. Do Install**

In a new console that we'll call the _work_ console (as we'll use it later):

```console
#clone the repo and enter into it
git clone https://github.com/oceanprotocol/ocean.py
cd ocean.py

#Install OS dependencies
sudo apt-get install -y python3-dev gcc python-pytest

#Initialize virtual environment and activate it.
python -m venv venv
source venv/bin/activate

#Install modules in the environment.
pip install -r requirements_dev.txt
```

***2. Run barge services***

In a new console:

```console
#grab repo
git clone https://github.com/oceanprotocol/barge
cd barge

#clean up old containers (to be sure)
docker system prune -a --volumes

#run barge: start ganache, Provider, Aquarius; deploy contracts; update ~/.ocean
./start_ocean.sh  --with-provider2
```

***3. Set up contracts***

In work console:

```console
#specify config file as an envvar
export CONFIG_FILE=config.ini

#set private keys of two accounts
export TEST_PRIVATE_KEY1=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58
export TEST_PRIVATE_KEY2=0xef4b441145c1d0f3b4bc6d61d29f5c6e502359481152f869247c7a4244d45209

#deploy new OCEAN token; update ~/.ocean/ocean-contracts/artifacts/address.json; send OCEAN to accounts
./deploy_fake_OCEAN.py
```

**4. Start to do stuff in python**

In Python console, try the following:
```python
import os
from ocean_lib.config import Config
```

It fails at the last line, with the following traceback.
```console
>>> from ocean_lib.ocean.ocean import Ocean
  File "/home/trentmc/code/ocean.py/ocean_lib/ocean/ocean.py", line 13, in <module>
    from ocean_lib.data_provider.data_service_provider import DataServiceProvider
  File "/home/trentmc/code/ocean.py/ocean_lib/data_provider/data_service_provider.py", line 19, in <module>
    from ocean_lib.web3_internal.utils import add_ethereum_prefix_and_hash_msg
  File "/home/trentmc/code/ocean.py/ocean_lib/web3_internal/utils.py", line 85, in <module>
    def privateKeyToAddress(private_key: str) -> str:
  File "/home/trentmc/code/ocean.py/ocean_lib/enforce_typing_shim.py", line 18, in enforce_types_shim
    if not ConfigProvider.get_config()["util"].getboolean("typecheck"):
  File "/home/trentmc/code/ocean.py/ocean_lib/config_provider.py", line 17, in get_config
    raise AssertionError("set_config first.")
AssertionError: set_config first.
```

Notes: 
* config.ini has default values, including `typecheck = True`. The error above still happens if `typecheck = False`.
* #37 "Revive dynamic type checking" was recently merged in, changes there likely led to this
* I encountered this issue while trying to do #205.

trentmc:
Fixes #270.

Changes proposed in this PR:
- made `enforce_types_shim()` robust to being called even before `ConfigProvider.set_config()` is called, which can easily happen in practice

### DIFF
--- a/ocean_lib/enforce_typing_shim.py
+++ b/ocean_lib/enforce_typing_shim.py
@@ -18,11 +18,11 @@ def setup_enforce_typing_shim():
 def enforce_types_shim(func):
     try:
         c = ConfigProvider.get_config()
-        typecheck = c["util"].getboolean("typecheck")
-
     except AssertionError:  # handle if ConfigProvider.set_config() not done yet
-        val = config_defaults["util"][NAME_TYPECHECK]
-        typecheck = distutils.util.strtobool(val)
+        c = config_defaults
+
+    val = c["util"][NAME_TYPECHECK]
+    typecheck = distutils.util.strtobool(val)
 
     if typecheck:
         return enforce_types(func)
--- a/ocean_lib/enforce_typing_shim.py
+++ b/ocean_lib/enforce_typing_shim.py
@@ -18,12 +18,11 @@ def enforce_types_shim(func):
     try:
         c = ConfigProvider.get_config()
         typecheck = c["util"].getboolean("typecheck")
-        
-    except AssertionError: #handle if ConfigProvider.set_config() not done yet
+
+    except AssertionError:  # handle if ConfigProvider.set_config() not done yet
         typecheck = config_defaults["util"][NAME_TYPECHECK]
 
     if typecheck:
         return enforce_types(func)
     else:
         return func
-
--- a/ocean_lib/enforce_typing_shim.py
+++ b/ocean_lib/enforce_typing_shim.py
@@ -26,5 +26,4 @@ def enforce_types_shim(func):
 
     if typecheck:
         return enforce_types(func)
-    else:
-        return func
+    return func
--- a/ocean_lib/enforce_typing_shim.py
+++ b/ocean_lib/enforce_typing_shim.py
@@ -5,7 +5,7 @@
 import os
 
 from enforce_typing import enforce_types
-from ocean_lib.config import Config
+from ocean_lib.config import Config, config_defaults, NAME_TYPECHECK
 from ocean_lib.config_provider import ConfigProvider
 from ocean_lib.ocean.env_constants import ENV_CONFIG_FILE
 
@@ -15,6 +15,15 @@ def setup_enforce_typing_shim():
 
 
 def enforce_types_shim(func):
-    if not ConfigProvider.get_config()["util"].getboolean("typecheck"):
+    try:
+        c = ConfigProvider.get_config()
+        typecheck = c["util"].getboolean("typecheck")
+        
+    except AssertionError: #handle if ConfigProvider.set_config() not done yet
+        typecheck = config_defaults["util"][NAME_TYPECHECK]
+
+    if typecheck:
+        return enforce_types(func)
+    else:
         return func
-    return enforce_types(func)
+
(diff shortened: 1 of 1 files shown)


ocean_lib/enforce_typing_shim.py
```diff
@@ -15,6 +16,15 @@ def setup_enforce_typing_shim():
...
+        c = ConfigProvider.get_config()
+    except AssertionError:  # handle if ConfigProvider.set_config() not done yet
+        c = config_defaults
+
+    val = c["util"][NAME_TYPECHECK]
+    typecheck = distutils.util.strtobool(val)
+
+    if typecheck:
+        return enforce_types(func)
+    else:
```
calina-c, line 29: we don't need the else branch. If the return inside the "if" branch is activated, there is no way of getting inside the "else" branch. (you can just remove the "else" and unindent `return func`)

ocean_lib/enforce_typing_shim.py
```diff
@@ -15,6 +16,15 @@ def setup_enforce_typing_shim():
 
 
 def enforce_types_shim(func):
-    if not ConfigProvider.get_config()["util"].getboolean("typecheck"):
+    try:
+        c = ConfigProvider.get_config()
+    except AssertionError:  # handle if ConfigProvider.set_config() not done yet
```
calina-c, line 21: can this AssertionError mean something else? Maybe it can be thrown for "incorrect" values as well as "missing" ones. When do we want to override with the defaults? Should we warn the end user in some way?

calina-c (changes requested):

ocean_lib/enforce_typing_shim.py
trentmc: I'm fine with this suggestion, I don't know the most pythonic way. Changed.

ocean_lib/enforce_typing_shim.py
trentmc: Right now config_provider.ConfigProvider has the following. 
```python
    def get_config():
        """Get a Config instance."""
        if not ConfigProvider._config:
            raise AssertionError("set_config first.")
        return ConfigProvider._config
```

We could make a more specific label than `AssertionError` or get fancier in how to handle it. However soon I will work on #202 "singletons for Web3Provider and more is surprising, non-standard, and won't work for multi-chain". In that, ConfigProvider is gonna get revamped anyway, and almost certainly change the function above. 

Therefore for _this_ issue I just want to do something surgical and move on. How does that sound?

calina-c: Great! I will approve and let's move on. If it's documented in other issues, that's all I need to hear 👍

calina-c (approved):

PR #272, opened 2021-04-09
Issue205 default config problems

### LINKED ISSUE #205: create_data_token() fail when using default CONFIG_FILE value, looks like artifacts issue
**Describe the bug**

The following configuration fails:
- using barge
- import ocean.py (vs git clone)
- a config.ini file that includes location of  addresses artifacts is _not_ specified
- exported CONFIG_FILE ennvar as config.ini
- in py code, do _not_ specify the location of config file (it's supposed to use CONFIG_FILE envvar)
- trying to do simple quickstart functionality 

Error: 
```console
  File "/home/trentmc/code/ocean.py/ocean_lib/ocean/ocean.py", line 157, in create_data_token
    assert address, "new datatoken has no address"
```

**To Reproduce**

1. Do initial steps in developers.md to install dependencies, run the services, set up contracts

2. In console:
```console
export CONFIG_FILE=config.ini
```

(This is actually done as part of "set up contracts", but putting here just to make it obvious that it was set.)

3. In python console:

```python
import os
from ocean_lib.config import Config

private_key = os.getenv('TEST_PRIVATE_KEY1')
config = Config() # note how no config file is specified. In this case, it uses CONFIG_FILE envvar
ocean = Ocean(config)

print("create wallet: begin")
wallet = Wallet(ocean.web3, private_key=private_key)
print(f"create wallet: done. Its address is {wallet.address}")

print("create datatoken: begin.")
```

The last line will fail. Traceback is below.

Note: it _does_ pass if the filename is explicitly stated. 

**Expected behavior**
The last line not to fail. While all pytest tests still passing.

Related: 
- #268 "Aquarius + tox crashes, looks like ocean.py artifacts issue"
- #185 "create_data_token() fail when addresses not specified, looks like artifacts issue" -- failed with the same error, with a different envvar setup (it didn't use config.ini, _this_ issue does)
- [multi-repo#83](https://github.com/oceanprotocol/multi-repo-issue/issues/83) "ocean-contracts is hard to use in python, including in ocean.py"

**Config.ini contents**
For reference, here is part of the config.ini file. 
```text
[eth-network]
network = ganache
artifacts.path = ~/.ocean/ocean-contracts/artifacts
address.file = ~/.ocean/ocean-contracts/artifacts/address.json
```

`artifacts.path` and `address.file` are the defaults for barge.

**Traceback**

```console
>>> datatoken = ocean.create_data_token("Dataset name", "dtsymbol", from_wallet=wallet) 
WARNING:root:No logs were found for tx 0x47c1f5e6096eb2d9a5cce8f7cd086b736b570ff5a3c576c4055579ec90df693d.
  File "/home/trentmc/code/ocean.py/ocean_lib/ocean/ocean.py", line 157, in create_data_token
    assert address, "new datatoken has no address"
AssertionError: new datatoken has no address
```

trentmc:
Fixes #205.

Changes proposed in this PR:
- Add support for case when default config file is to be used

### DIFF
--- a/ocean_lib/config.py
+++ b/ocean_lib/config.py
@@ -9,6 +9,7 @@ import os
 import site
 from pathlib import Path
 
+from ocean_lib.ocean.env_constants import ENV_CONFIG_FILE
 from ocean_lib.web3_internal.constants import GAS_LIMIT_DEFAULT
 
 DEFAULT_NETWORK_HOST = "localhost"
@@ -124,7 +125,7 @@ class Config(configparser.ConfigParser):
         self._web3_provider = None
         self._section_name = "eth-network"
         self._logger = logging.getLogger("config")
-
+        
         if filename:
             self._logger.debug(f"Config: loading config file {filename}")
             with open(filename) as fp:
@@ -136,6 +137,12 @@ class Config(configparser.ConfigParser):
         elif options_dict:
             self._logger.debug(f"Config: loading from dict {options_dict}")
             self.read_dict(options_dict)
+        else:
+            filename = os.getenv(ENV_CONFIG_FILE)
+            self._logger.debug(f"Config: loading config file {filename}")
+            with open(filename) as fp:
+                text = fp.read()
+                self.read_string(text)
 
         self._load_environ()
 
--- a/ocean_lib/config.py
+++ b/ocean_lib/config.py
@@ -139,6 +139,8 @@ class Config(configparser.ConfigParser):
             self.read_dict(options_dict)
         else:
             filename = os.getenv(ENV_CONFIG_FILE)
+            if filename is None:
+                raise ValueError(f'Config file envvar "{ENV_CONFIG_FILE}" is empty')
             self._logger.debug(f"Config: loading config file {filename}")
             with open(filename) as fp:
                 text = fp.read()
--- a/ocean_lib/config.py
+++ b/ocean_lib/config.py
@@ -125,7 +125,7 @@ class Config(configparser.ConfigParser):
         self._web3_provider = None
         self._section_name = "eth-network"
         self._logger = logging.getLogger("config")
-        
+
         if filename:
             self._logger.debug(f"Config: loading config file {filename}")
             with open(filename) as fp:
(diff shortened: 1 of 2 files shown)


ocean_lib/config.py
```diff
@@ -136,6 +137,12 @@ def __init__(self, filename=None, options_dict=None, **kwargs):
         elif options_dict:
             self._logger.debug(f"Config: loading from dict {options_dict}")
             self.read_dict(options_dict)
+        else:
+            filename = os.getenv(ENV_CONFIG_FILE)
```
calina-c, line 141: can you please add a test for this? I believe there are some configuration tests which can be adapted quickly.

calina-c: what happens if the ENV_CONFIG_FILE is empty? Will it try to open an invalid filename or None? Can we check that the file exists before loading it?

Also, is it a file local to us, or could the end user interfere and potentially send a huge file/overload the system?

calina-c (changes requested):

ocean_lib/config.py
trentmc: > please add a test for this.

OK. I added many more, much better now. c781cfe..e698292

> what happens if ENV_CONFIG_FILE is empty

Good point. I added a test for this, and give a ValueError when it happens.

ocean_lib/test/test_config_provider.py
```diff
@@ -54,6 +148,16 @@ def test_config_from_text():
     assert config.aquarius_url == "https://another-aqua.url"
 
 
+def test_config__from_text__malfformed_content():
```
calina-c, line 151: just a tiny typo here, I assume it should be one f for "malformed".
also I am not partial to the double underscoring, since double underscore has meaning other than separator: https://dbader.org/blog/meaning-of-underscores-in-python and also in some dialects like e.g. Django they are automatically interpreted by internal tools e.g. attr1__attr2 means attr1.attr2 in the database. I would suggest simple underscores, since either way the Setup and Expect are documented. But it's up to you, I'm not *very* adamant about this.

trentmc: thanks, fixed both

calina-c (approved):

PR #274, opened 2021-04-09
Exception Handlers

### LINKED ISSUE #230: Non-specific errors hinder maintenance
Right now, ocean.py throws a lot of generic `AssertionError` or even plain `Exception`. This is cumbersome to track, since a _plain_ Exception can show up with things unrelated to what someone is trying to test.

**Solution**
Reuse more specific errors from libraries we are already requiring.

**Alternatives and additional context**
- @trentmc suggested we do NOT over-specify and write custom exception classes for everything, and I agree. However, we need to better distinguish between exceptions and we can find a middle ground.
- An example of exceptions we could use from web3, these seem more granular: https://github.com/ethereum/web3.py/blob/master/web3/exceptions.py.

0x3bfc:
Fixes #230 

Changes proposed in this PR:

- [x] Add Exception module
- [x] Replace the current assertion errors with the new exceptions
- [x] Refine the exceptions and make them more generic


calina-c:
@0x3bfc according to the specs described in the ticket, we were not supposed to write custom classes, but reuse existing ones from packages, or Python ones. At least that is what I understood from @trentmc. I'll add some possible suggestions on each exception individually.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -214,15 +215,15 @@ def create(
             dt = DataToken(data_token_address)
             minter = dt.contract_concise.minter()
             if not minter:
-                raise AssertionError(
+                raise InvalidDatatokenContract(
```
calina-c, line 218: e.g. web3.InvalidAddress exception

ocean_lib/ocean/ocean_assets.py
```diff
@@ -214,15 +215,15 @@ def create(
...
                     f"datatoken address {data_token_address} does not seem to be a valid DataToken contract."
                 )
             elif minter.lower() != publisher_wallet.address.lower():
-                raise AssertionError(
+                raise InvalidDatatokenMinter(
```
calina-c, line 222: e.g.  web3 ValidationError or some Python generic error?

ocean_lib/ocean/ocean_assets.py
```diff
@@ -314,7 +315,7 @@ def create(
                 publisher_wallet,
             )
             if not ddo_registry.verify_tx(tx_id):
-                raise AssertionError(
+                raise FailedToCreateDDO(
```
calina-c, line 318: something like web3 transaction reverted? etc.

calina-c (changes requested):
added some suggestions, not set in stone but possibilities.

calina-c:
I also had a new idea. Up to discussion @0x3bfc and @trentmc how do you feel about introducing *some* custom exceptions, indicative of another repo/component? E.g. something like a ProviderError, AquariusError which would help with debugging when tracing issues to other components?

ocean_lib/ocean/ocean_assets.py
trentmc: > web3.InvalidAddress exception

+1, ok with me

ocean_lib/ocean/ocean_assets.py
trentmc: I just checked Python generic errors at https://www.tutorialsteacher.com/python/error-types-in-python

None of them are appropriate.

Web3.ValidationError is sufficient.

trentmc:
> I also had a new idea. Up to discussion @0x3bfc and @trentmc how do you feel about introducing some custom exceptions, indicative of another repo/component? E.g. something like a ProviderError, AquariusError which would help with debugging when tracing issues to other components?

I'm ok with this too.

Related: I just (re) discovered the [eth-event](https://github.com/iamdefinitelyahuman/eth-event) repo, a small python library for handling events. Built & maintained over the last 2+  years by the brownie creator. We have another github issue looking to improve events, this would be helpful. And for this issue, the repo introduces a [few error types](https://github.com/iamdefinitelyahuman/eth-event/blob/master/eth_event/main.py) that we could use as well.

ocean_lib/ocean/ocean_assets.py
0x3bfc: I didn't find something like that in the web3.exceptions but I am going to define it in the exception module.

calina-c:
@0x3bfc can you please fix and adjust the tests first? I will review then.

calina-c (approved):

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright 2021 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+
+class ContractNotFound(Exception):
```
trentmc, line 7: For this exception class, and all the other exception classes in this file:
-The docstring should all be on one line, including triple-quotes
-The class definition needs a "pass" (I'm surprised that unit tests passed without having this, hmmm..)

Example of correct way:
----
class ContractNotFound(Exception):
    """Contract address is not found in the factory events"""
    pass
----
More examples:
https://www.programiz.com/python-programming/user-defined-exception

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+
+class TransactionReverted(Exception):
+    """
+    Generic Transaction Revert Error
+    """
+
+
+class InvalidURL(Exception):
```
trentmc, line 50: Don't need this as our own custom exception. Replace it with requests.exceptions.InvalidURL

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+
+class BPoolError(Exception):
+    """
+    Generic Balancer Pool Error
+    """
+
+
+class ExchangeError(Exception):
```
trentmc, line 38: "Exchange" is too generic.

Suggest changing to:
-ExchangeError -> FreError
-Generic Ocean Exchange Error -> Generic FRE (Fixed Rate Exchange) error

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Generic Ocean Exchange Error
+    """
+
+
+class TransactionReverted(Exception):
```
trentmc, line 44: Suggest renaming: 
-TransactionReverted -> VerifyTxFailed
-Generic transaction revert error -> Transaction failed verification; tx may have reverted

Why: "transaction reverted" does not sound like an error. And in the places where it's used, it's always about a "verify tx" call that comes back with negative results.

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -89,7 +90,7 @@ def create(self, data_token: str, exchange_rate: float, wallet: Wallet) -> str:
         # get event log from receipt
         logs = exchange.contract.events.ExchangeCreated().processReceipt(tx_receipt)
         if not logs:
-            raise AssertionError(
+            raise ExchangeError(
```
trentmc, line 93: Replace ExchangeError --> VerifyTxFailed

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -122,7 +123,7 @@ def setRate(
         # get event log from receipt
         logs = exchange.contract.events.ExchangeRateChanged().processReceipt(tx_receipt)
         if not logs:
-            raise AssertionError(
+            raise ExchangeError(
```
trentmc, line 126: Replace ExchangeError --> VerifyTxFailed

ocean_lib/ocean/ocean_pool.py
```diff
@@ -92,14 +93,14 @@ def create(
             pool_address, data_token_amount, from_wallet, wait=True
         )
         if dt.get_tx_receipt(tx_id).status != 1:
-            raise AssertionError(
+            raise BPoolError(
```
trentmc, line 96: Replace BPoolError --> VerifyTxFailed

ocean_lib/ocean/ocean_pool.py
```diff
@@ -92,14 +93,14 @@ def create(
...
                 f"Approve datatokens failed, pool was created at {pool_address}"
             )
 
         ot = DataToken(self.ocean_address)
         tx_id = ot.approve_tokens(pool_address, OCEAN_amount, from_wallet, wait=True)
         if ot.get_tx_receipt(tx_id).status != 1:
-            raise AssertionError(
+            raise BPoolError(
```
trentmc, line 103: Replace BPoolError --> VerifyTxFailed

ocean_lib/ocean/ocean_pool.py
```diff
@@ -114,7 +115,7 @@ def create(
             from_wallet,
         )
         if pool.get_tx_receipt(tx_id).status != 1:
-            raise AssertionError(
```
trentmc, line 117: Replace BPoolError --> VerifyTxFailed

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+class AssetsError(Exception):
+    """
+    Generic Assets Error
+    """
+
+
+class BPoolError(Exception):
+    """
+    Generic Balancer Pool Error
+    """
```
trentmc, line 35: We can delete this, since all the places it's used below can be replaced by VerifyTxFailed

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -64,7 +65,7 @@ def buy_at_fixed_rate(
         # Figure out the amount of ocean tokens to approve before triggering the exchange function to do the swap
         ocean_amount_base = exchange.get_base_token_quote(exchange_id, amount_base)
         if ocean_amount_base > max_OCEAN_amount_base:
-            raise AssertionError(
+            raise ExchangeError(
```
trentmc, line 68: Replace ExchangeError with web3.exceptions.ValidationError.

ocean_lib/exceptions.py
0x3bfc: This `pass` make the codacy report fails.
https://pycodequ.al/docs/pylint-messages/w0107-unnecessary-pass.html

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+#
+
+
+class ContractNotFound(Exception):
+    """
+    Contract address is not found
+    in the factory events
+    """
+
+
```
trentmc, line 13: It's best that errors aren't named based on a **location** (module) for the error since that information is already provided in the stack trace; but rather the error is more the **type** of issue. 

DDOError, BPoolError, and ExchangeError only give a location, whereas e.g. ValidationError is the type.

Therefore many of my suggestions below are about renaming to specific types of errors. By the time we're done, there shouldn't be any location-based error classes (DDOError, BPoolError, and ExchangeError).

ocean_lib/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+class AssetsError(Exception):
+    """
+    Generic Assets Error
+    """
+
+
+class BPoolError(Exception):
```
trentmc, line 32: Delete/rename as discussed above

ocean_lib/exceptions.py
trentmc, line 38: Delete/rename as discussed above

ocean_lib/ocean/ocean_assets.py
```diff
@@ -303,7 +311,7 @@ def create(
                 del file["url"]
             metadata_copy["encryptedFiles"] = files_encrypted
         else:
-            raise AssertionError("Encrypting the files failed.")
+            raise DDOError("Encrypting the files failed.")
```
trentmc, line 314: Should rename away from **location** to **type** of error. 

The closest thing I found in other python repos was "UnicodeEncodeError". But that's not quite right.

Therefore here I suggest a custom class of "EncryptionError".

ocean_lib/ocean/ocean_assets.py
```diff
@@ -324,7 +332,7 @@ def create(
                 publisher_wallet,
             )
             if not ddo_registry.verify_tx(tx_id):
-                raise AssertionError(
+                raise TransactionReverted(
```
trentmc, line 335: Rename to VerifyTxFailed

ocean_lib/ocean/ocean_assets.py
```diff
@@ -348,7 +356,7 @@ def update(self, asset: Asset, publisher_wallet: Wallet) -> str:
...
```
trentmc, line 359: Rename to VerifyTxFailed

trentmc (changes requested):
This is great progress. +1

My main suggestion is to move away from errors that say the location of the module (that's already in the stack trace), to the type of the error. I give precise suggestions.

Also in a couple places we can reuse errors from requests and web3.py library.

calina-c:
@trentmc @0x3bfc can we hold off on this until I finish my audit of common-utils-py? It seems wee have some missed opportunities in exception handling there.

trentmc:
> can we hold off on this until I finish my audit of common-utils-py? It seems wee have some missed opportunities in exception handling there.

Or we can simply close _this_ issue off (once concerns are addressed), and in your (Calina's) work on common-utils-py you add some specific exceptions there.

I'm fine with either approach. To reduce complexity, I lean to the latter - close _this_ issue off sooner.

calina-c:
Closing this because of multiple conflicts and TODOs. Partly superseded by #293. The rest will be done later (issue is not closed).

PR #276, opened 2021-04-12
SQL Alchemy conversion

### LINKED ISSUE #201: Deprecate SQL statements

SQL statements are obsolete as of Ocean V3. So deprecate them.

(Rather than: Right now, we are using raw SQL for statements in ocean_lib/data_store/auth_tokens.py. This is flagged as a potential security risk, via SQL injection. And suggested solution:  We should move to using Flask SQL Alchemy instead.)

Deprecate

Note: blocked by "remove fake OCEAN"

calina-c:
Closes #201.

Changes proposed in this PR:

- add SQL Alchemy as a backend for SQL statements
- prevent SQL injections with auth token SQL statements
- drop one usage of an ocean_utils class (StorageBase)

Pending: Since file DB creation is not tested, I don't know whether the path given for the DB file in config.ini is absolute or relative. And, if relative... to what? I need to test this separately, or get an answer from someone who does know :)


calina-c:
Superseded by #287.

PR #277, opened 2021-04-13
Removes restart function from OceanCompute.

### LINKED ISSUE #232: Restart function in OceanCompute is either obsolete or broken. So deprecate or adjust.
A couple of weeks ago, we had updates on the `start` function, allowing for multiple inputs. However, we overlooked the fact that the `restart` function also needs to be updated. Right now, the function itself is BROKEN (it calls on a function which has different parameters, so there is always an Exception thrown). That is bad. Updating it will most likely BREAK the interface and we need to release a new ocean.py version. That is also bad. But we need to fix it one way or the other.

**Possible solutions**
- force the user to pass the input list again when restarting the job -> breaks the interface, since the input_list parameter needs to be added
- somehow recover the original input list from the original job when restarting -> is this even possible using the status endpoint, or something similar?
- remove the `restart` function entirely. Per @ssallam there is no `restart` function in provider, so why do we bother wrapping this so much? In essence, it is just a stop() and start().

**Discussion:**
@alexcos20 how did you handle this in ocean.js? @trentmc, @ssallam which of the possibilities would you vote for?

Btw: there is a TODO in test_compute_flow.py referencing this. Whoever works on this needs to remove it.

calina-c:
Closes #232.

Changes proposed in this PR:

- remove the obsolete restart function in OceanCompute

### DIFF
--- a/ocean_lib/ocean/ocean_compute.py
+++ b/ocean_lib/ocean/ocean_compute.py
@@ -388,26 +388,6 @@ class OceanCompute:
             )
         )
 
-    def restart(self, did, job_id, wallet):
-        """
-        Attempt to restart the compute job by stopping it first, then starting a new job.
-
-        :param did: str id of the asset offering the compute service of this job
-        :param job_id: str id of the compute job
-        :param wallet: Wallet instance
-        :return: str -- id of the new compute job
-        """
-        _, service_endpoint = self._get_service_endpoint(did)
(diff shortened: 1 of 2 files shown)


trentmc (approved):

trentmc:
Reviewed, approved. I defer to you @calina-c to squash & merge

PR #280, opened 2021-04-13
Don't ignore IDE settings folders

DMats:
Justification: We already ignore .idea/ (IntelliJ IDEA-based IDE settings folder)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,6 @@ ocean_lib.egg-info/
 /scripts/
 
 .coverage
-.idea/
-.vscode/
 .tox/
 consume-downloads/
 coverage.xml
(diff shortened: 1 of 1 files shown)


calina-c:
That's true, but I'm ambivalent about this. I think these should reside in each developer's local gitignore. Not sure if we intentionally ignore IDEA files, or it was part of some gitignore boilerplate. I have nothing against the change, though. Depends on how strict we want to be.

DMats:
@calina-c and I discussed it privately.

Conclusion: 

1. `ocean.py/.gitignore` file should not ignore any IDE settings folders. 
2. Developers can locally ignore their particular IDE's settings folder if they choose. For instructions to do this see https://stackoverflow.com/a/1753078

Action items:

1. Remove `.idea/` from `.gitignore` instead of adding `.vscode/`.

PR #281, opened 2021-04-13
Issue231 reduce type conversions

### LINKED ISSUE #231: Base 18 conversions
There are currently inconsistencies in the handling of large integers. Smart contracts use big integer number representation and multiply a number with 10^18 in order to properly represent it. Ocean.py is inconsistent in dealing with this representation, sometimes showing end-users the "long" number and internally using the short version etc.

**Solution**
Everything related to internals should be handled in the "long" format.
Everything related to GUIs should be handled in the "short" format (aka normal numbers 😅 )

**Discussion and additional context**
- We need to identify and fix where each format is appropriate throughout the application.
- An example: in the models/data_token.py file, requesting `get_info` will return a dictionary. One of its keys is `totalSupply`, with a value of x. This makes me think there are x tokens. However, running the `totalSupply` function directly will return x * 10^18. This makes me think I have many more tokens than x. And if I don't check the value to *actually see* that something is wrong, I might perform operations on them, and the result will be wrong.

DMats:
Towards #231  .

# Changes

- Add currency utilities - `to_wei`, `from_wei`, `pretty_ether`, `ether_fmt`
- Remove `to_base_18`, `from_base_18`, and `to_base`. Replace uses with `to_wei` and `from_wei`
- Add convenience functions - `wei_and_pretty_ether`, `pretty_ether_from_wei`
- Refactor Datatoken
  - Remove Datatoken methods that took ether parameter instead of wei - `approv_tokens`, `mint_tokens`, `transfer_tokens`, `calculate_fee`
  - Remove unused Datatoken methods - `calculate_max_fee`, `get_max_fee_percentage`
  - Refactor Datatoken.DEFAULT_CAP to be expressed in wei instead of ether
  - Add temporary hack DEFAULT_CAP_IN_ETHER constant to be removed after all APIs are in wei
  - Refactor Datatoken fee constants to be expressed as wei per token instead of floating-point percentages
- Remove commented code from test_fixed_rate_exchange.py
- Add constants - `MAX_UINT256`, `MAX_INT256`, `MIN_INT256`

# Target `wei-interface` branch
This PR targets the `wei-interface` branch instead of the `main` branch. The reasoning is because this PR is in an intermediate state:
1. The to/from wei conversion utilities have been updated to use Decimal instead of float.
2. BUT many of the APIs still contain float parameters so the internal logic still converts from float to Decimal and then from ether to wei.

# Hacks
All the tests are passing, but making this possible required several hacks that should be removed once all APIs only support amounts denoted in wei (encoded as ints):
1. The constant `Datatoken.DEFAULT_CAP_IN_ETHER`
2. assertions in `test_public_pool` and `test_spot_price` check for int values that result from float calculations
3. `ocean_exchange.buy_at_fixed_rate` approves max_OCEAN_amount instead of estimated Ocean amount.


DMats:
@calina-c I've requested your review despite this PR still being in draft state.  I'm hoping you could take a glance at the utilities I've created - `to_wei`, `from_wei`, `pretty_ether`, `ether_fmt`, `wei_and_pretty_ether`, `pretty_ether_from_wei` - and let me know what you think.  

One thing I want to do, but haven't had time yet, is move them to `ocean_lib.ocean.currency` instead of `ocean_lib.web3_internal.currency`.  This would be a hint to lib users that it's intended for public use.

Another thing I continue to wonder about is whether my to_wei and from_wei need to support decimal places other than 18.

As you can see from the PR description - the changes thus far have laid the groundwork but actually haven't eliminated very many of the conversions between wei and ether. I'll be aggressively tackling this next when I get back to this task.

The checks are currently failing due to floating-point precision errors. These tests will need drastic changes once the APIs are changed to be wei-only so I'm not too worried about them right now.

calina-c:
@DMats I'll assign some time to this review for Monday, since it is a large chunk and not urgent (still in progress). Sorry for the delay, but I need to focus on some other emergencies today.

calina-c (commented):
I have taken a look. Seems ok to me until now, waiting for removal of draft label to actually review :)

DMats:
This PR targets the `wei-interface` branch instead of the `main` branch. The reasoning is because this PR is in an intermediate state:
1. The to/from wei conversion utilities have been updated to use Decimal instead of float.
2. BUT many of the APIs still contain float parameters so the internal logic still converts from float to Decimal and then from ether to wei.

All the tests are passing, but making this possible required several hacks that should be removed once all APIs only support amounts denoted in wei (encoded as ints):
1. The constant `Datatoken.DEFAULT_CAP_IN_ETHER`
2. assertions in `test_public_pool` and `test_spot_price` check for int values that result from float calculations
3. `ocean_exchange.buy_at_fixed_rate` approves max_OCEAN_amount instead of estimated Ocean amount.

calina-c (approved):
What can I say about this one? It looks like a titanic amount of work. I have largely reviewed it, but I can't possibly go to the depths @DMats went to. LGTM. Amazing work, David! This looks excellent, well-done, solid and surprisingly fast.

🥳

PR #285, opened 2021-04-14
References to aquarius_url & metadata_store_url are inconsistent.

### LINKED ISSUE #229: References to aquarius_url & metadata_store_url are inconsistent; replace with metadata_cache_uri everywhere
There are multiple references inside ocean.py to metadata_store_uri and aquarius_url, and they are loaded from the same configuration option, called aquarius_url. However, the user can supersede these with metadata_cache_uri. Currently we have 3 names for the same thing, or things that are very similar, and it is confusing to users and code maintainers.

**Solution**
Settle on one of these and adjust variable naming and config loading to use one name, or document why and when different names can be used.

The one to use is: `metadata_cache_uri`. 
- Details: `aquarius_url` was the first name, then it was changed to `metadata_store_url` to move away from "aquarius" in general. 
- Some places were kept simply for backwards compatibility. But the essence of this ticket is correct: things could be a lot cleaner without sacrificing backwards compatibility.

**Additional context**
This has been tackled in a previous issue (#126), but only in terms of config. The naming is still random in the codebase itself.

mariacarmina:
Fixes #229  .

Changes proposed in this PR:

- replace with metadata_cache_uri everywhere;
- add fallback to the change above;
- test the modification.


ocean_lib/test/test_config.py
```diff
@@ -0,0 +1,14 @@
+#
+# Copyright 2021 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from ocean_lib.config import Config
+
+
+def test_metadata_cache_url_property():
```
DMats, line 9: Please add a docstring

ocean_lib/config.py
```diff
@@ -217,6 +217,15 @@ def gas_limit(self):
     def metadata_store_url(self):
         return self.get("resources", NAME_AQUARIUS_URL)
 
+    @property
+    def metadata_cache_url(self):
```
DMats, line 221: Please add a docstring

DMats (changes requested):
From https://github.com/oceanprotocol/ocean.py/issues/229#issue-833091642
>The one to use is: `metadata_cache_uri`.

Please change all usages of `metadata_cache_url` to `metadata_cache_uri`.

Please also change:

1. [READMEs/marketplace-flow.md::133](https://github.com/oceanprotocol/ocean.py/blame/issue229-replace-with-metadataCacheUri/READMEs/marketplace-flow.md#L133)
`ocean.config.metadata_store_url` to `ocean.config.metadata_cache_uri`
2. [READMEs/compute-flow.md::86](https://github.com/oceanprotocol/ocean.py/blame/issue229-replace-with-metadataCacheUri/READMEs/compute-flow.md#L86)
`ocean.config.metadata_store_url` to `ocean.config.metadata_cache_uri`

DMats:
@mariacarmina I left a comment on the issue that could result in additional changes. https://github.com/oceanprotocol/ocean.py/issues/229#issuecomment-819858470

ocean_lib/config.py
```diff
@@ -217,6 +217,15 @@ def gas_limit(self):
...
+        cache_key = self.get("resources", NAME_AQUARIUS_URL)
```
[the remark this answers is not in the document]
mariacarmina: @calina-c wanted me to do a fallback and I needed some guidance in this case. In the first instance, I have also put `return self.get("resources", NAME_AQUARIUS_URL)`, but I wanted a feedback according to the fallback method.

mariacarmina (commented):
Will add a fallback for the `AQUARIUS_URL` as @trentmc suggested in #229.

mariacarmina:
> Fixes #229 .
> 
> Changes proposed in this PR:
> 
>     * replace with metadata_cache_uri everywhere;
> 
>     * add fallback to the change above;
> 
>     * test the modification.

Note: It might break the API.

READMEs/parameters.md
```diff
@@ -38,8 +38,6 @@ d = {
 ocean = Ocean(d)
 ```
 
-For legacy support, you can also use `metadataStoreUri` instead of `metadataCacheUri`.
```
calina-c, line 41: this is still true, I think. We name the variable metadata_cache_uri everywhere, but the user can configure it using any of these keys.

ocean_lib/ocean/ocean.py
```diff
@@ -78,12 +78,7 @@ def __init__(self, config=None, data_provider=None):
                 config = Config(os.getenv(ENV_CONFIG_FILE))
                 ConfigProvider.set_config(config)
         if isinstance(config, dict):
-            # fallback to metadataStoreUri
-            cache_key = (
```
calina-c, line 82: Please keep the config fallback.

calina-c (changes requested):

READMEs/parameters.md
mariacarmina: I have talked with @DMats who wasn't sure about the fallback that's why I removed it, but I can add it right now.

DMats: Apologies, I must have mispoke. Calina's right.

DMats:
@mariacarmina , your work was on the right track but there was a bug in the fallback logic. Plus, I saw there were many small instances where `aquarius_url` was still used that needed to be stripped out.  I decided it would be easier if I just did the changes myself instead of trying to describe them in the PR.

While working, I unearthed a latent bug in `Config` where the section names were not handled properly. For example, config values from env vars were assigned to the `eth-network` section of the config even if they belonged in the `resources` section or `util` section. I had to fix this bug for the fallback logic to work properly for config options (eg. `resources.metadata_cache.uri`)

After my changes here is how the fallback logic works:

1. If both `resources.metadata_cache.uri` and `resources.aquarius.url` config options are defined (via config file, config dict, or config string) a `ValueError` is raised. 
2. If only `resources.aquarius.url` is defined, its value is assigned to `resources.metadata_cache.uri` and a deprecation warning is logged. 
3. If the `METADATA_CACHE_URI` env var is defined, it will override the `resources.metadata_cache.uri` config option.
4. If both the `METADATA_CACHE_URI` and `AQUARIUS_URL` env vars are defined, a `ValueError` is raised.
5. If only `AQUARIUS_URL` is defined, its value is assigned to the `resources.metadata_cache.uri` and a deprecation warning is logged.

I've added unit tests that verify this behavior. I updated all readmes. The newly added Aquarius wrapper from `common_utils` is the only place where the usage of `aquarius_url` is still prevalent, but I think this makes sense.

DMats (approved):

READMEs/parameters.md
```diff
@@ -59,8 +58,8 @@ First, in your working directory, create `config.ini` file and fill as follows:
     network = https://rinkeby.infura.io/v3/<your infura project id>
 
     [resources]
-    aquarius.url = https://provider.rinkeby.oceanprotocol.com
-    provider.url = https://aquarius.rinkeby.oceanprotocol.com
+    metadata_cache.uri = https://aquarius.rinkeby.oceanprotocol.com
```
calina-c, line 61: should this be an underscore? or dots? maybe not a combination of the two... `metadata_cache_uri`?

calina-c: if we want to make it like provider_url... should we use "L" instead of "I"?

ocean_lib/config.py
```diff
@@ -145,21 +177,71 @@ def __init__(self, filename=None, options_dict=None, **kwargs):
...
+                    f"Use only {SECTION_RESOURCES}.{NAME_METADATA_CACHE_URI} because "
+                    f"{SECTION_RESOURCES}.{NAME_AQUARIUS_URL} is deprecated."
+                )
+            )
+
     def _load_environ(self):
-        for option_name, environ_item in environ_names.items():
-            value = os.environ.get(environ_item[0])
+        for option_name, environ_item in environ_names_and_sections.items():
+            if option_name == NAME_METADATA_CACHE_URI:
```
calina-c, line 213: this has already been tackled in the specific function. We should treat all aspects in "_handle_and_validate...", or keep something very minimal here. Otherwise this chunk of code dedicated to cache_uris distracts devs from the rest of the functionality.

ocean_lib/test/test_config.py
```diff
@@ -0,0 +1,63 @@
+#
```
calina-c, line 1: I think there is already a test_config.py. Also, ocean_lib/test/ is for integration tests, not for unit tests. Can we use the other test_config.py instead?

ocean_lib/test/test_config_provider.py
```diff
@@ -22,18 +22,18 @@ def test_set_config():
     assert ConfigProvider.get_config() == "foo config"
 
 
-def test_metadataCacheUri_version():
-    """Tests that the Aquarius URL can use the custom metadataCacheUri key."""
-    config_dict = {"metadataCacheUri": "http://ItWorked.com", "network": "ganache"}
+def test_metadataStoreUri_version():
+    """Tests that the Aquarius URL can use the custom metadataStoreUri key."""
```
calina-c, line 26: please adjust these docstrings to reflect the new purpose of these tests.

calina-c (changes requested):

READMEs/parameters.md
DMats: I chose `metadata_cache.uri` because there are previous examples of options with underscores, yet all of the options that refer to URL/URIs separate the last part with a dot. I'm definitely open to change if there's a strong preference for something different.

My top choice for alternatives would be `metadata_cache_uri` like @calina-c suggested because it does away with the dot notation which seemed odd to me.

I used URI because the issue #229 description seemed to prefer `uri` over `url`. I think breaking the consistency between this new option and the legacy options is OK. In light of this, I think we should stick to uri for this new option. Please let me know if there's a strong opposing opinion.

ocean_lib/config.py
DMats: This method (`_load_environ`) is for handling the environment variables.  The `_handle_and_validate_metadata_cache_uri` method is for the special treatment of the `resources.metadata_cache.uri` config option (passed via config file, dict, or string). I believe both methods are necessary.

ocean_lib/test/test_config.py
DMats: Are you referring to `test_config_provider.py`? I don't see the "other `test_config.py`".

I think that `ocean.py/tests/integration` is the appropriate place to put integration tests and that `ocean_lib/test/test_config.py` is correctly following the convention for unit tests (to place them in a `test` folder that is a sibling of the file under test).

ocean_lib/test/test_config_provider.py
DMats: Done.

READMEs/parameters.md
calina-c: I'm open to both `metadata_cache_uri` and `metadata_cache.uri` but the second one looks like a typo, even if it isn't :D It just pops. @trentmc deferring to you for guidance here.

ocean_lib/config.py
```diff
@@ -145,21 +177,71 @@ def __init__(self, filename=None, options_dict=None, **kwargs):
...
+                metadata_cache_uri = os.environ.get(environ_item[0])
+                aquarius_url = os.environ.get(
+                    deprecated_environ_names[NAME_AQUARIUS_URL][0]
+                )
+
+                if metadata_cache_uri and aquarius_url:
+                    raise ValueError(
```
calina-c, line 220: at least this one is duplicated. Let's try to offload some functionality to handle_and_validate...

ocean_lib/config.py
calina-c: at least this one is duplicated. Let's try to offload some functionality to handle_and_validate...

ocean_lib/test/test_config.py
calina-c: ok, works for me, thanks!

ocean_lib/config.py
DMats: The errors/warnings are different so I think both are needed.  One refers to config options like `resources.metadata_cache.uri` the other refers to env vars like `METADATA_CACHE_URI`.

calina-c: all right. Let's keep it this way :)

READMEs/parameters.md
DMats: @calina-c @trentmc and I agreed that `metadata_cache_uri` with all underscores was preferable.

calina-c (approved):

DMats (approved):

PR #288, opened 2021-04-16
Non-blocking nature of calls hurts DX/maintenance. So, use blocking

mariacarmina:
Fixes # 227.

Changes proposed in this PR:

- add blocking to verify if the tx was successful or not;
- add unit test for that;
- test on multiple networks to see if it works.


mariacarmina (commented):
The last commit represents the current prototype of how I thought to handle the issue.

calina-c:
@mariacarmina This PR contains the metadata cache url changes as well. Please only add what is necessary, to help with the code review process.

PR #290, opened 2021-04-16
Add unit tests and improve code coverage v2.0

### LINKED ISSUE #284: Add unit tests and improve code coverage (part 2)
Follow-on of Issue #214 and PR #267. 

A couple of high-priority bug fixes became wrapped up in #267 so it was cut short and merged before adding all the desired unit tests. This issue #284 is for adding the unit tests that were missing from #267.

TODO:
Add unit tests for the following files (list copied from issue #214 and then augmented):

- [x] ocean_lib/assets/test/test_asset.py
- [x] ocean_lib/assets/test/test_asset_resolver.py
- [x] ocean_lib/assets/test/test_utils.py
    - [x] test that attempts to remove an algo that doesn't exist in the trusted algos list
    - [x] test that attempts to add an algo that already exists in the trusted algos list
    - [x] improve `test_create_asset_with_owner_address` (see below)
- [x] ocean_lib/data_provider/test/test_exceptions.py
- [x] ocean_lib/data_store/test/test_auth_tokens.py
- [x] ocean_lib/models/test/test_metadata.py
- [x] ocean_lib/models/test/test_order.py
- [x] ocean_lib/ocean/test/test_ocean_compute.py
- [x] ocean_lib/ocean/test/test_ocean_assets.py
    - [x] improve `test_create_asset_with_owner_address` (see below)
- [x] ocean_lib/ocean/test/test_util.py

# Improve `test_create_asset_with_owner_address`
Problem:
This test only checks whether the asset is created when both `owner_address` and `data_token_address` parameters are set. I think more assertions are warranted. 

Details:
After digging into `OceanAssets.create()` I realized that the `owner_address` parameter is silently ignored if `data_token_address` parameter is set.

Possible Solution:

```python
# Create two assets: one with `owner_address` set, the other without.
asset_1 = ocn.assets.create(
        asset.metadata,
        alice,
        [auth_service],
        owner_address=alice.address,
        data_token_address=token.address,
    )
assert asset_1
asset_2 = ocn.assets.create
        asset.metadata,
        alice,
        [auth_service],
        data_token_address=token.address,
    )
assert asset_2
# Verify that the resulting assets are equivalent.
assert asset_1 == asset_2
```

# Improve `test_create_asset_with_owner_address` 
Problem:

The current test calls `OceanAssets.create()` with the `owner_address` parameter set to the same address as the `publisher_wallet`. 

Possible Solution:

1. Duplicate this test.
2. Change the first test so that the owner_address parameter is different from the publisher address.
3. Keep the second test the same.

Reasoning:
This tests both the intended and unintended usage of the method.

mariacarmina:
Towards #284  .

Changes proposed in this PR:

- add the tests that were not implemented in the first PR of unit tests;
- improve code coverage too;
- add docstrings as well.


ocean_lib/assets/test/test_utils.py
```diff
@@ -100,3 +129,38 @@ def test_remove_trusted_algorithm(publisher_ocean_instance):
 
     assert new_publisher_trusted_algorithms is not None
     assert len(new_publisher_trusted_algorithms) < len(publisher_trusted_algorithms)
+
+
+def test_remove_unexisting_trusted_algorithm(publisher_ocean_instance):
```
calina-c, line 134: can you please combine the tests? It seems we have the same publishing setup on these tests. And the setup with publishing a ddo is costly. If we publish DDOs for each test case, that is more likely an integration test and not a unit test.

Possible suggestions: You can use a sample DDO (with compute service) instead. OR you can sequentially run the trusted_algorithm tests on the same published dataset.

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -254,13 +254,36 @@ def test_create_asset_with_owner_address(publisher_ocean_instance):
...
+    assert asset_1
+    asset_2 = ocn.assets.create(
+        asset.metadata,
+        alice,
+        [auth_service],
+        data_token_address=token.address,
+    )
+    assert asset_2
+    assert (
+        asset_1.__dict__["_proof"]["creator"] == asset_2.__dict__["_proof"]["creator"]
```
calina-c, line 284: can you use the object instead of the __dict__ representation? Using __dict__ is more like a shortcut for testing, when you don't know the structure of the object closely. But it's better to check with object properties.

calina-c (changes requested):

ocean_lib/assets/test/test_utils.py
mariacarmina: I think I will go with the first option (to use DDO sample) since I started to create separate functions.

ocean_lib/assets/test/test_utils.py
```diff
@@ -41,13 +45,41 @@ def test_add_trusted_algorithm(publisher_ocean_instance):
...
+    assert new_publisher_trusted_algorithms is not None
+    assert len(new_publisher_trusted_algorithms) < len(publisher_trusted_algorithms)
+
+    # remove a trusted algorithm that does not belong to publisher_trusted_algorithms list
+    new_publisher_trusted_algorithms = remove_publisher_trusted_algorithm(
+        ddo.did, algorithm_ddo_v3.did, publisher_ocean_instance.config.aquarius_url
+    )
+
+    assert new_publisher_trusted_algorithms is not None
+    for _, trusted_algorithm in enumerate(publisher_trusted_algorithms):
```
calina-c, line 79: if you are not using the indices, use a simple for trusted_algorithm in publisher_trusted_algorithms. I think I did this modification already, but you force-pushed over my correction.

calina-c (approved):

calina-c (approved):

PR #291, opened 2021-04-17
Issue228 data token erc20 interface

### LINKED ISSUE #228: data_token.py ERC20 functionality doesn't reflect ERC20 interface
**Problems**

Datatokens implement the ERC20 interface, and tack on a couple more pieces of functionality. 

This is exactly the case in the smart contract level: [DataTokenTemplate inherits ERC20](https://github.com/oceanprotocol/contracts/blob/master/contracts/templates/DataTokenTemplate.sol). Good.

But [data_token.py](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/models/data_token.py) does _not_ reflect it cleanly:
- Its ERC20 interface should reflect those of the smart contract. But it doesn't. When we wrote the ocean.py alpha in mid 2020, it had just a subset, more for reasons of speed of implementation. But then new methods were added in an ad-hoc fashion; those new methods break the principle. 
- For example, get_info() returns name, symbol and more as a single function :/ . This was first introduced as [this PR](https://github.com/oceanprotocol/ocean.py/pull/74) with [this commit](https://github.com/oceanprotocol/ocean.py/commit/26f241eb12da750767c0d4088e44521a7a943cdb)). This is non-standard. It's also a poor OO interface: each getter should focus on a single piece of info, just as the smart contract level.

**Proposed Solution**

Ensure that the interface for data_token.py's ERC20 functionality is the same as the underlying smart contract. Like how bpool.py does it. 

**Tips**

- We can likely have it inherit from BToken.py which is an ERC20 python interface. 
- Or perhaps make it directly use [web3.py ERC20 functionality](https://web3py.readthedocs.io/en/stable/examples.html#working-with-an-erc20-token-contract). To be explored:)

**Notes**
- Doing _this_ issue will help towards #231 "base type conversions" because it will deprecate interfaces that break base18.

DMats:
Closes #228 .

# Changes proposed in this PR

- Add functions that exist in [DataTokenTemplate.sol](https://github.com/oceanprotocol/contracts/blob/master/contracts/templates/DataTokenTemplate.sol) and [Open Zeppelin's ERC20.sol (version 2.5.0)](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v2.5.0/contracts/token/ERC20/ERC20.sol) but are missing in [data_token.py](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/models/data_token.py)
- Add and update unit tests to cover newly added methods
- Create the `DataToken.calculate_token_holders()` method and use it in `ocean_pool.py`, where appropriate.
- Eliminate the `DataToken.get_info()` method because it breaks the OO principle of each getter returning a single piece of info.


DMats:
I request approval that this PR can be merged without fixing the two code climate issues: 

>Function `initialize` has 7 arguments (exceeds 4 allowed).

`initialize` is merely a wrapper so it makes sense that it's signature match the underlying smart contract.

>`DataToken` has 28 functions (exceeds 20 allowed).

Reducing the number of DataToken methods will be accomplished in https://github.com/oceanprotocol/ocean.py/issues/231

trentmc:
This looks very good from the perspective of ocean.py.

This may cause breaking changes to Provider or Aquarius. Have you looked into that? If it does, then please make sure your release process accounts for updating them. 

And related to that -- the Ocean Market team has asked to hold off on any new releases in backend until C2D ships (which is soon).

This may be a good opportunity to create some integration tests across ocean.py, Aquarius, and Provider all at once.

trentmc (changes requested):
See my comments above. That is: The changes _here_ may be fine but we need to account for interaction with Provider etc.

ocean_lib/models/data_token.py
```diff
@@ -40,6 +41,128 @@ class DataToken(ContractBase):
...
+        fee_collector_address: str,
+        from_wallet: Wallet,
+    ) -> str:
+        return self.send_transaction(
+            "initialize",
+            (name, symbol, minter_address, cap, blob, fee_collector_address),
+            from_wallet,
+        )
+
+    def mint(self, account_address: str, value_base: int, from_wallet: Wallet) -> str:
```
calina-c, line 62: @DMats I like what you did with the interface here. But since I added the tests and documented this part, I was wondering whether we could make all these functions more DRY.

It seems like all of them are self.send_transaction("`name_of_function`", `args_except_wallet`, `wallet`). Is there a way we could compound all of these somehow?

DMats:
I have not analyzed what effect this PR will have on Provider or Aquarius. I'll look into it and perhaps create some integration tests in the process.

I won't merge to main on any of the backend components until I hear from the frontend team that C2D is shipped.

ocean_lib/models/data_token.py
DMats: @calina-c Thanks. 

I struggle to imagine how we could make these functions **more** DRY. Every `DataToken` method that calls `ContractBase.send_transaction` does so with different parameters. This same pattern is used throughout ocean.py, not just in `DataToken`.

Adding a compound data class (shown below) is possible, but would only reduce repetition when sending the same transaction twice. Since this is not a common use case, I don't see the need to explicitly support it. Users can use a [packed argument list](https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists) if they really want to send the same transaction twice.

```python
@dataclass
class Transaction
    """Class to hold the parameters for sending a transaction"""
    fn_name: str,
    fn_args: tuple,
    from_wallet: Wallet,
    raw_transaction: dict = None,
```

In conclusion, I don't think further changes are warranted. If I've misunderstood your question, please let me know.

DMats:
@trentmc The only potentially API breaking change is the removal of the `get_info` method. I verified that `get_info` is not used in Aquarius and Provider using "find all". 

I ended up not removing the methods that denominate token amounts in Ether instead of Wei because I think that change fits better in #231 

I'm not sure how to go about creating integration tests so I think it might be good to push this off to a different PR.

ocean_lib/models/data_token.py
calina-c: OK, let's leave it for some other time. It's outside of the scope either way. Great job!

calina-c (approved):

trentmc (approved):
Thanks for the comments David. Sounds good to me. I'm ok with you merging.

PR #294, opened 2021-04-19
Non-blocking nature of calls hurts DX/maintenance. Use blocking.

### LINKED ISSUE #227: Non-blocking nature of calls hurts DX/maintenance. So, use blocking
**Problem**
In debugging issue #185 and #205, I would have found errors sooner if it was clear that the tx did not go through. For those: there was a call to create a datatoken, but the tx did not work.

In other cases, the tx may work but there could be non-negligible network latency until the tx goes through.

The general challenge: the developer needs to muck around to know if the tx is done, whether successful or not.

**Proposed Solution**
We can make it better: make the calls _blocking_. That is, wait until the tx has completed. If it doesn't complete or fails, return an error or None. Otherwise simply keep going. Do it in `contract_base.py::send_transaction()`, using `is_tx_successful()`.

This will make the DX better because it abstracts away all network-y stuff for ocean.py users. 

It also helps ocean.py maintainability.

**TODOs**

- [x] first cut implement
- [x] ask: is it addressing when a tx fails? And are there unit tests on this? if not, fix
- [x] ask: is it addressing when gas is super-low and tx hangs indefinetely? And are there unit tests on this? If not, fix. ocean.js addresses by having a timeout of 10 min.
- [x] get all pytest / tox tests working locally
- [x] update READMEs
- [x] test on rinkeby or ropsten
- [ ] test on eth mainnet
- [x] test on polygon 
- [x] test with "import ocean_lib". Including on rinkeby or ropsten, eth mainnet, and polygon. Why: may cause weird things here too
- [ ] commit PR to main
- [ ] create a new release

mariacarmina:
Fixes #227 .

Changes proposed in this PR:

- add blocking to verify if the tx was successful or not;
- add unit test for that;
- test on multiple networks to see if it works.

### DIFF
--- a/ocean_lib/web3_internal/contract_base.py
+++ b/ocean_lib/web3_internal/contract_base.py
@@ -224,6 +224,11 @@ class ContractBase(object):
         if transact:
             _transact.update(transact)
 
+        is_successful = self.is_tx_successful(
+            contract_function.transact(_transact).hex()
+        )
+        if not is_successful:
+            return "Tx was not successful."
         return contract_function.transact(_transact).hex()
 
     def get_event_argument_names(self, event_name: str):
--- a/ocean_lib/web3_internal/test/test_contract_base.py
+++ b/ocean_lib/web3_internal/test/test_contract_base.py
@@ -26,6 +26,34 @@ class MyFactory(ContractBase):
         )
 
 
+def test_send_transaction_with_a_wrong_password(dtfactory_address, alice_wallet):
+    factory = MyFactory(dtfactory_address)
+    old_hash = factory.createToken(
+        "foo_blob", "DT1", "DT1", to_base_18(1000.0), alice_wallet
+    )
+    _transact = {
+        "from": alice_wallet.address,
+        "passphrase": "my_secret_passphrase",
+        "account_key": alice_wallet.key,
+    }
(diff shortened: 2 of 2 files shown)


trentmc:
Below are the PRs before committing to main, and therefore before taking the PR out of draft.

It seems that there's quite a lot more work to go. E.g. it doesn't seem to handle "tx hangs indefinitely". etc etc. 

I'm happy to give you feedback as you go of course! Of the stuff so far, here's feedback:
- there _is_ a first implementation for "is it addressing when a tx fails". However returning an error message as a string seems a bit weird, because the error message isn't a string. Maybe raise an error? Or other. I suggest: take a look at the ocean.js logic and see what it does. Also it has logic for "tx hangs indefinitely". If both ocean.js and ocean.py behave similarly, we're in good shape.

----

- [ ] first cut implement
- [ ] ask: is it addressing when a tx fails? And are there unit tests on this? if not, fix
- [ ] ask: is it addressing when gas is super-low and tx hangs indefinetely? And are there unit tests on this? If not, fix. ocean.js addresses by having a timeout of 10 min.
- [ ] get all pytest / tox tests working locally
- [ ] update READMEs
- [ ] test on rinkeby or ropsten
- [ ] test on eth mainnet
- [ ] test on polygon

PR #300, opened 2021-04-22
Deprecate deploy_fake_OCEAN & on-the-fly mint and distribute in test setup

### LINKED ISSUE #278: Deprecate creating fake OCEAN from within ocean.py

In ocean.py for marketplace README quickstart, we generate fake OCEAN as explicit step. Most recently for #200.

Alex mentioned late last week that now ocean contracts repo generates fake OCEAN, and then it shows up in barge. If this is truly the case then we don't need the explicit step in OCEAN.

Deprecate creating fake OCEAN from within ocean.py

Make sure it works in the marketplace README.

**TODOs**

- [x] Revive Calina's PR
- [ ] test the PR, fix as needed
- [ ] commit PR to main
- [ ] new ocean_lib release to pypi
- [ ] test "import ocean_lib" in marketplace README, does it work?

~~Notes: fixing this will unblock #201 - SQL alchemy. PR #276~~ (That issue is deprecated, don't need)

calina-c:
Fixes #278.


mariacarmina:
As far as I saw the symbol for `BTOKEN` is kind of irrelevant, because @trentmc said that the purpose of the `BTOKEN` is to generate  ERC20 tokens.

trentmc:
> the purpose of the BTOKEN is to generate ERC20 tokens.

To be precise, the purpose of the BToken class is to _represent_ ERC20 tokens. The "B" means Balancer, as BToken is one of the Balancer smart contracts. Balancer code uses it for managing ERC20 in its codebase, e.g. each pool has its own ERC20 tokens called BPTs.

Then the Python BToken module is simply a wrapper of BToken class. This module is convenient for creating an ERC20 token (using a "deploy" of BToken contract), or modeling an ERC20 token that already exists on the network (e.g. OCEAN).

ocean_lib/ocean/mint.py
```diff
@@ -71,32 +66,14 @@ def deploy_fake_OCEAN():
...
-            deployer_addr,
-            OCEAN_cap_base,
-            "",
-            deployer_addr,
-        )
-    )
-    addresses["Ocean"] = OCEAN_token.address
-    print("****Deploy fake OCEAN: done****\n")
-
+    OCEAN_token = DataToken(address=network_addresses["development"]["Ocean"])
```
trentmc, line 72: All the other changes are about changing from "deploy & mint" to just "mint". But _this_ line is actually a _deploy_. I'm confused, what is the intent?

And I might have an answer. The idea of this issue overall is that we can _deprecate_ all of deploy & mint fake OCEAN here, because it's apparently already done in Barge. Therefore we should need any deploy(), or mint(), or really any of this code. Instead in the quickstart README we can simply add one line of code to mint() from within the demo code. _All_ of this can get deprecated.

trentmc (changes requested):
The idea of this issue overall is that we can deprecate all of deploy & mint fake OCEAN here, because it's apparently already done in Barge. Therefore we should need any deploy(), or mint(), or really any of this code. Instead in the quickstart README we can simply add one line of code to mint() from within the demo code. All of this can get deprecated.

mariacarmina:
> > the purpose of the BTOKEN is to generate ERC20 tokens.
> 
> To be precise, the purpose of the BToken class is to _represent_ ERC20 tokens. The "B" means Balancer, as BToken is one of the Balancer smart contracts. Balancer code uses it for managing ERC20 in its codebase, e.g. each pool has its own ERC20 tokens called BPTs.
> 
> Then the Python BToken module is simply a wrapper of BToken class. This module is convenient for creating an ERC20 token (using a "deploy" of BToken contract), or modeling an ERC20 token that already exists on the network (e.g. OCEAN).

So actually do we need the specific symbol for that if this class wants to model a certain ERC20?

ocean_lib/ocean/mint.py
calina-c: @trentmc here we are simply retrieving the token deployed by Barge. The deployment part was deprecated.

mariacarmina: Agreed with @calina-c. We can use instead the address for deployment from address.json as far as I understood.

trentmc: > @trentmc here we are simply retrieving the token deployed by Barge. The deployment part was deprecated.

Ah right.

But why do you need this script at all then? Why not just directly reference it in the unit tests etc?

calina-c: I think we could mint it on the spot, if necessary. We can discuss this together.

ocean_lib/ocean/mint.py
```diff
@@ -21,40 +21,36 @@
```
DMats, line 21: Please remove `setup_enforce_typing_shim()` and its requisite import statement. They are not required because methods in this file are no longer executed before the static `ConfigProvider` object has been initialized. As a result, `isort:skip_file` and all instances of `# noqa: E402` should also be removed.

DMats (changes requested):

ocean_lib/ocean/mint.py
calina-c: Done :)

trentmc (dismissed):
I'm satisfied with this now. Thanks. I defer to @calina-c to squash and merge.

Issue #283, opened 2021-04-14
Assertions without error messages


0x3bfc:
The error messages need to be defined below: 

https://github.com/oceanprotocol/ocean.py/blob/028fcacae9554ea85bde3e72295762bde98f99c8/ocean_lib/assets/utils.py#L63

https://github.com/oceanprotocol/ocean.py/blob/028fcacae9554ea85bde3e72295762bde98f99c8/ocean_lib/assets/utils.py#L73

https://github.com/oceanprotocol/ocean.py/blob/028fcacae9554ea85bde3e72295762bde98f99c8/ocean_lib/assets/utils.py#L22

https://github.com/oceanprotocol/ocean.py/blob/028fcacae9554ea85bde3e72295762bde98f99c8/ocean_lib/assets/utils.py#L90

Also consider any missing errors in the other modules.

```bash
grep -irn 'assert ' ocean_lib
```

PR #310, opened 2021-04-29
Issue 304 consolidate web3 utils

### LINKED ISSUE #304: Move contents of Web3Helper to utils.py, change static methods to module functions, remove duplicates, use Web3Provider
Doing this would offer a couple of benefits:

1. Imports can be more explicit. Only the functions used are imported, instead of the Web3Helper class.
2. Call sites can be shorter - Don't need the `Web3Helper.` prefix.
3. This would also put all of our utilities built on top of Web3.py in one place.

In my mind, this issue builds towards #231 because abbreviating the call sites will reduce the formatting churn that #231 will cause. (In other words, it avoids adding the `Web3Helper.` prefix everywhere.) I split it out into a separate issue because #231 is already going to be very large and anything I can do to reduce its size and complexity will make reviewing it much easier.

DMats:
Fixes #304 .

Changes proposed in this PR:

- Move Web3Helper methods that require a Wallet to `transactions.py`
- Move remaining Web3Helper methods to `utils.py`
- Make all moved methods into module functions.


calina-c (dismissed):

conftest.py
```diff
@@ -62,10 +63,8 @@ def setup_all(request):
     print(
         f"sender: {wallet.key}, {wallet.address}, {wallet.password}, {wallet.keysStr()}"
     )
-    print(
-        f"sender balance: {Web3Helper.from_wei(Web3Helper.get_ether_balance(wallet.address))}"
-    )
-    assert Web3Helper.from_wei(Web3Helper.get_ether_balance(wallet.address)) > 10
+    print(f"sender balance: {from_wei(get_ether_balance(wallet.address))}")
+    assert from_wei(get_ether_balance(wallet.address)) > 10
```
calina-c, line 67: I forgot about this when I approved. Please add messages to all asserts, per #283. It is in progress on another PR to add them retrospectively, so let's add them in all the new PRs as well.

calina-c: Actually, I needed the web3 helper changes in #309, so I added the asserts. Hope you don't mind! 😄

PR #314, opened 2021-05-04
Adds DDO credentials.

### LINKED ISSUE #296: Credentials support in asset
See [EPIC](https://app.zenhub.com/workspaces/ocean-main-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/multi-repo-issue/98) for details

Ocean.py should:
- allow publisher to set credentials
- check credentials on consumption
- allow publisher to update credentials

### LINKED ISSUE #298: Add ocean.assets.isConsumable
This function should check if the asset is consumable by :
- check isDisabled flag (it exists & true -> asset is not consumable)
- check file conectivity
- check credentials

Return should be a interface, containing a status and a status_message: Eg
   0, All good
   1, Asset is disabled
   2, Connectivity check failure
   3, Credential missing from allow list
   4, Credential found on deny list

calina-c:
Closes #296. Closes #298. Clean-up #277

### DIFF
--- a/ocean_lib/common/agreements/consumable.py
+++ b/ocean_lib/common/agreements/consumable.py
@@ -1,0 +1,20 @@
+#
+# Copyright 2021 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+
+class ConsumableCodes:
+    OK = 0
+    ASSET_DISABLED = 1
+    CONNECTIVITY_FAIL = 2
+    CREDENTIAL_NOT_IN_ALLOW_LIST = 3
+    CREDENTIAL_IN_DENY_LIST = 4
+
+
+class UnsupportedCredential(Exception):
+    pass
+
+
+class MalformedCredential(Exception):
+    pass
--- a/ocean_lib/common/agreements/consumable.py
+++ b/ocean_lib/common/agreements/consumable.py
@@ -6,7 +6,7 @@
 
 class ConsumableCodes:
     OK = 0
-    ASSET_DISABLED = 1
+    ASSET_NOT_LISTED = 1
     CONNECTIVITY_FAIL = 2
     CREDENTIAL_NOT_IN_ALLOW_LIST = 3
     CREDENTIAL_IN_DENY_LIST = 4
--- a/ocean_lib/common/agreements/consumable.py
+++ b/ocean_lib/common/agreements/consumable.py
@@ -12,9 +12,5 @@ class ConsumableCodes:
     CREDENTIAL_IN_DENY_LIST = 4
 
 
-class UnsupportedCredential(Exception):
-    pass
-
-
 class MalformedCredential(Exception):
     pass
--- a/ocean_lib/common/agreements/consumable.py
+++ b/ocean_lib/common/agreements/consumable.py
@@ -14,3 +14,8 @@ class ConsumableCodes:
 
 class MalformedCredential(Exception):
     pass
+
+
+class AssetNotConsumable(Exception):
+    def __init__(self, consumable_code):
+        self.consumable_code = consumable_code
--- a/ocean_lib/common/agreements/consumable.py
+++ b/ocean_lib/common/agreements/consumable.py
@@ -6,7 +6,7 @@
 
 class ConsumableCodes:
     OK = 0
-    ASSET_NOT_LISTED = 1
+    ASSET_DISABLED = 1
     CONNECTIVITY_FAIL = 2
     CREDENTIAL_NOT_IN_ALLOW_LIST = 3
     CREDENTIAL_IN_DENY_LIST = 4
--- a/ocean_lib/common/ddo/credentials.py
+++ b/ocean_lib/common/ddo/credentials.py
@@ -11,13 +11,10 @@ class AddressCredential:
 
     def get_addresses_of_class(self, access_class="allow"):
         """Get a filtered list of addresses from credentials (use with allow/deny)."""
-        entries = self.asset._credentials.get(access_class, [])
-
-        for entry in entries:
-            if entry["type"] == "address":
-                return [val.lower() for val in entry["values"]]
-
-        return []
+        address_entry = self.get_address_entry_of_class(access_class)
+        return (
+            [addr.lower() for addr in address_entry["values"]] if address_entry else []
+        )
 
     def requires_credential(self):
         """Checks whether the asset requires an address credential."""
@@ -54,16 +51,15 @@ class AddressCredential:
             ]
             return
 
-        address_entries = self.get_address_entry_of_class(access_class)
+        address_entry = self.get_address_entry_of_class(access_class)
 
-        if not address_entries:
+        if not address_entry:
             self.asset._credentials[access_class].append(
                 {"type": "address", "values": [address]}
             )
             return
 
-        address_entry = address_entries[0]
-        lc_addresses = self.get_lc_addresses_from_entry(address_entry)
+        lc_addresses = self.get_addresses_of_class(access_class)
 
         if address not in lc_addresses:
             lc_addresses.append(address)
@@ -77,13 +73,12 @@ class AddressCredential:
         if not self.asset._credentials or access_class not in self.asset._credentials:
             return
 
-        address_entries = self.get_address_entry_of_class(access_class)
+        address_entry = self.get_address_entry_of_class(access_class)
 
-        if not address_entries:
+        if not address_entry:
             return
 
-        address_entry = address_entries[0]
-        lc_addresses = self.get_lc_addresses_from_entry(address_entry)
+        lc_addresses = self.get_addresses_of_class(access_class)
 
         if address not in lc_addresses:
             return
@@ -92,9 +87,10 @@ class AddressCredential:
         address_entry["values"] = lc_addresses
 
     def get_address_entry_of_class(self, access_class="allow"):
-        """Get a filtered list of addresses of a given access_class from credentials (use with allow/deny)."""
-        entries = self.asset._credentials.get(access_class)
-        return [entry for entry in entries if entry["type"] == "address"]
+        """Get address credentials entry of the specified access class. access_class = "allow" or "deny"."""
+        entries = self.asset._credentials.get(access_class, [])
+        address_entries = [entry for entry in entries if entry["type"] == "address"]
+        return address_entries[0] if address_entries else None
 
     def get_lc_addresses_from_entry(self, address_entry):
         """Get an address entry of a given access class from credentials (use with allow/deny)."""
--- a/ocean_lib/common/ddo/credentials.py
+++ b/ocean_lib/common/ddo/credentials.py
@@ -20,12 +20,14 @@ class AddressCredential:
         return []
 
     def requires_credential(self):
+        """Checks whether the asset requires an address credential."""
         allowed_addresses = self.get_addresses_of_class("allow")
         denied_addresses = self.get_addresses_of_class("deny")
 
         return allowed_addresses or denied_addresses
 
     def validate_access(self, credential=None):
+        """Checks a credential dictionary against the address allow/deny lists."""
         address = simplify_credential_to_address(credential)
 
         allowed_addresses = self.get_addresses_of_class("allow")
@@ -42,48 +44,46 @@ class AddressCredential:
 
         return ConsumableCodes.OK
 
-    def add_address_to_list_class(self, address, list_class="allow"):
+    def add_address_to_access_class(self, address, access_class="allow"):
+        """Adds an address to an address list (either allow or deny)."""
         address = address.lower()
 
-        if not self.asset._credentials or list_class not in self.asset._credentials:
-            self.asset._credentials[list_class] = [
+        if not self.asset._credentials or access_class not in self.asset._credentials:
+            self.asset._credentials[access_class] = [
                 {"type": "address", "values": [address]}
             ]
             return
 
-        address_entry = get_list_entry_with_type_address(
-            self.asset._credentials[list_class]
-        )
+        address_entries = self.get_address_entry_of_class(access_class)
 
-        if not address_entry:
-            self.asset._credentials[list_class].append(
+        if not address_entries:
+            self.asset._credentials[access_class].append(
                 {"type": "address", "values": [address]}
             )
             return
 
-        address_entry = address_entry[0]
-        lc_addresses = [addr.lower() for addr in address_entry["values"]]
+        address_entry = address_entries[0]
+        lc_addresses = self.get_lc_addresses_from_entry(address_entry)
 
         if address not in lc_addresses:
             lc_addresses.append(address)
 
         address_entry["values"] = lc_addresses
 
-    def remove_address_from_list_class(self, address, list_class="allow"):
+    def remove_address_from_access_class(self, address, access_class="allow"):
+        """Removes an address from an address list (either allow or deny)i."""
         address = address.lower()
 
-        if not self.asset._credentials or list_class not in self.asset._credentials:
+        if not self.asset._credentials or access_class not in self.asset._credentials:
             return
 
-        address_entry = get_list_entry_with_type_address(
-            self.asset._credentials[list_class]
-        )
+        address_entries = self.get_address_entry_of_class(access_class)
 
-        if not address_entry:
+        if not address_entries:
             return
 
-        address_entry = address_entry[0]
-        lc_addresses = [addr.lower() for addr in address_entry["values"]]
+        address_entry = address_entries[0]
+        lc_addresses = self.get_lc_addresses_from_entry(address_entry)
 
         if address not in lc_addresses:
             return
@@ -91,8 +91,18 @@ class AddressCredential:
         lc_addresses.remove(address)
         address_entry["values"] = lc_addresses
 
+    def get_address_entry_of_class(self, access_class="allow"):
+        """Get a filtered list of addresses of a given access_class from credentials (use with allow/deny)."""
+        entries = self.asset._credentials.get(access_class)
+        return [entry for entry in entries if entry["type"] == "address"]
+
+    def get_lc_addresses_from_entry(self, address_entry):
+        """Get an address entry of a given access class from credentials (use with allow/deny)."""
+        return [addr.lower() for addr in address_entry["values"]]
+
 
 def simplify_credential_to_address(credential):
+    """Extracts address value from credential dictionary."""
     if not credential:
         return None
 
@@ -100,7 +110,3 @@ def simplify_credential_to_address(credential):
         raise MalformedCredential("Received empty address.")
 
     return credential["value"]
-
-
-def get_list_entry_with_type_address(types_list):
-    return [entry for entry in types_list if entry["type"] == "address"]
--- a/ocean_lib/common/ddo/credentials.py
+++ b/ocean_lib/common/ddo/credentials.py
@@ -2,11 +2,7 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
-from ocean_lib.common.agreements.consumable import ConsumableCodes
-from ocean_lib.common.utils.utilities import (
-    get_list_entry_with_type_address,
-    simplify_credential_to_address,
-)
+from ocean_lib.common.agreements.consumable import ConsumableCodes, MalformedCredential
 
 
 class AddressCredential:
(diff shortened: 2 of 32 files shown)


calina-c:
Leaving as Draft and removing reviewers for now, to be combined with #298.

ocean_lib/common/agreements/consumable.py
```diff
@@ -0,0 +1,20 @@
...
+
+class ConsumableCodes:
+    OK = 0
+    ASSET_DISABLED = 1
+    CONNECTIVITY_FAIL = 2
+    CREDENTIAL_NOT_IN_ALLOW_LIST = 3
+    CREDENTIAL_IN_DENY_LIST = 4
+
+
+class UnsupportedCredential(Exception):
```
calina-c, line 15: remove. If it is not an address, do not throw the unsupported credential error

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +376,49 @@ def create_authentication_from_json(values):
...
+
+        if allowed_addresses and address.lower() not in allowed_addresses:
+            return ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST
+
+        if not allowed_addresses and address.lower() in denied_addresses:
+            return ConsumableCodes.CREDENTIAL_IN_DENY_LIST
+
+        return ConsumableCodes.OK
+
+    def is_consumable(self, credential=None, provider_uri=None):
```
calina-c, line 411: flag for "with_connectivity" 🆗

ocean_lib/models/test/test_ddo.py
```diff
@@ -37,6 +39,93 @@ def get_ddo_sample(datatoken_address):
...
+def test_ddo_credentials_addresses_both():
+    """Tests DDO credentials when both deny and allow lists exist on the asset."""
+    sample_ddo_path = get_resource_path("ddo", "ddo_sa_sample_with_credentials.json")
+    assert sample_ddo_path.exists(), "{} does not exist!".format(sample_ddo_path)
+
+    ddo = DDO(json_filename=sample_ddo_path)
+    assert ddo.get_addresses_of_type("allow") == ["0x123", "0x456a"]
+    assert ddo.get_addresses_of_type("deny") == ["0x2222", "0x333"]
+    assert (
+        ddo.get_address_allowed_code({"type": "address", "value": "0x111"})
```
DMats, line 51: Based on my recommendation to remove the `get_address_allowed_code` method, I recommend that you use `is_consumable` here like so:

```python
assert (
    ddo.is_consumable({"type": "address", "value": "0x111"}, with_connectivity_check=False) == ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST
)
```

Extrapolate this recommendation to all other uses of `get_address_allowed_code`

ocean_lib/common/agreements/consumable.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ConsumableCodes:
+    OK = 0
+    ASSET_DISABLED = 1
+    CONNECTIVITY_FAIL = 2
+    CREDENTIAL_NOT_IN_ALLOW_LIST = 3
+    CREDENTIAL_IN_DENY_LIST = 4
+
+
+class MalformedCredential(Exception):
```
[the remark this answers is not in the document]
calina-c: That would only leave the ConsumableCodes declared here. And the two are tightly linked (2 of the codes are `CREDENTIAL`-related). I would keep them in the same file.

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +387,64 @@ def create_authentication_from_json(values):
 
         authentication = {"type": authentication_type, "publicKey": key_id}
         return authentication
+
+    def get_addresses_of_type(self, section_type="allow"):
```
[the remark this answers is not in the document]
calina-c: I wouldn't go into generalising quite yet. That is why I only left some comments indicating that there can be extensions in the future. Why?
1. this PR already does a lot of things
2. we don't know which transformations apply to other credentials. E.g. address must be lowercased for any comparison, but maybe other credentials are case sensitive or need some symbol stripping etc. Which is why we can't generalise _now_.

Ref. your point (1), I agree the naming is bad now, but I'm not in love with `allow_or_deny` either, because it contains an `or`, and it also restricts. Will all future credentials only have `allow` and `deny`, or could there be other modifiers? I'll add some clarity here, but I need to give it some more thought.

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +387,64 @@ def create_authentication_from_json(values):
...
+            if entry["type"] == "address":
+                return [val.lower() for val in entry["values"]]
+
+        return []
+
+    def enable(self):
+        if not self._status:
+            self._status = {}
+
+        self._status.pop("isOrderDisabled")
```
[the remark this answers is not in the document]
calina-c: I went for implicit rather than explicit. If that key is not present, I think the structure of the DDO is clearer, otherwise a lot of flags will be set even if not needed. What do you think?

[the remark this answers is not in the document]
calina-c: I went for implicit rather than explicit. If that key is not present, I think the structure of the DDO is clearer, otherwise a lot of flags will be set even if not needed. What do you think?

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +387,64 @@ def create_authentication_from_json(values):
...
+
+    def disable(self):
+        if not self._status:
+            self._status = {}
+
+        self._status["isOrderDisabled"] = True
+
+    def get_address_allowed_code(self, credential=None):
```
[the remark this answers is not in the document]
calina-c: I think it makes it much more easier to read, by separating the concerns. One function handles high-level access granting, while this one is a helper. No, it should not be called outside is_consumable, but we don't have protected/private in Python :(

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +387,64 @@ def create_authentication_from_json(values):
...
+
+        if allowed_addresses and address.lower() not in allowed_addresses:
+            return ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST
+
+        if not allowed_addresses and address.lower() in denied_addresses:
+            return ConsumableCodes.CREDENTIAL_IN_DENY_LIST
+
+        return ConsumableCodes.OK
+
+    def requires_address_credential(self):
```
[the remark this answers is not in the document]
calina-c: See my comment above regarding generalising.

ocean_lib/common/ddo/ddo.py
```diff
@@ -354,3 +387,64 @@ def create_authentication_from_json(values):
...
+        allowed_addresses = self.get_addresses_of_type("allow")
+        denied_addresses = self.get_addresses_of_type("deny")
+
+        return allowed_addresses or denied_addresses
+
+    def is_consumable(
+        self, credential=None, with_connectivity_check=True, provider_uri=None
```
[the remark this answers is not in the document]
calina-c: I don't want to do this. I started like this, but CodeClimate gave me really bad ratings on readability and I agree. See my comment above on separation of concerns.

ocean_lib/common/utils/utilities.py
```diff
@@ -73,3 +75,13 @@ def checksum(seed):
 def get_timestamp():
     """Return the current system timestamp."""
     return f"{datetime.utcnow().replace(microsecond=0).isoformat()}Z"
+
+
+def simplify_credential_to_address(credential):
```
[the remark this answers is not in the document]
calina-c: This function does one thing: takes the "value" out of the credential object :)

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -312,56 +312,6 @@ def stop_compute_job(did, job_id, service_endpoint, consumer_address, signature)
             "put", did, job_id, service_endpoint, consumer_address, signature
         )
 
-    @staticmethod
-    def restart_compute_job(
```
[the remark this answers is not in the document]
calina-c: The restart functionality was previously removed from higher-level components, and it was left stale here (not called anywhere).

ocean_lib/models/test/test_ddo.py
calina-c: Addressed that in separation of concerns. And, as you can see, tests are also easier to do when functions handle separate concerns. I am interested in testing ONLY address access here, not connectivity or disabling. Using your proposed version of `is_consumable`, I would have to bundle up all these tests together.

calina-c:
@DMats I refactored the whole shabang for a middle ground between your review and my... "reluctance" 😁 
I only left some convenience methods related to addresses in ddo.py, and moved most of the functionality to an AddressCredential class. When we decide to extend with other credentials, we can implement similar classes, and follow a factory-like pattern to select the Credential type.

Let me know, and let's touch base again when it's daytime in your timezone.

ocean_lib/data_provider/data_service_provider.py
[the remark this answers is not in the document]
calina-c: Ref: https://github.com/oceanprotocol/ocean.py/commit/1dab70d164ca36a6cff284e8be82ae04344ad13f

ocean_lib/common/ddo/ddo.py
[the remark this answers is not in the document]
DMats: > we don't know which transformations apply to other credentials. E.g. address must be lowercased for any comparison, but maybe other credentials are case sensitive or need some symbol stripping etc. Which is why we can't generalise now.

I totally missed this in my initial review.  You're right. It's too early to generalize.

> Will all future credentials only have allow and deny, or could there be other modifiers? 

Standard ACL rules used in networking are either "permit" or "deny". Other modifiers are treated as separate parameters. That said, I like your chosen variable name "access_class" which resolves my concern about the variable name.

ocean_lib/common/ddo/ddo.py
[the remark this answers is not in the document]
DMats: Fair enough.

ocean_lib/common/utils/utilities.py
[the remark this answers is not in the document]
DMats: Based on your assertion that this method takes the "value" out of the credential object, I'd argue that it shouldn't exist because it should be as simple as `credential["value"]`

But if you insist on keeping this method, then I think it should be moved to `common/agreements/consumable.py` so that all the inputs and outputs of credential-related logic can be stored together. I don't think it belongs in `utilities.py` because the rest of the file focuses on id creation, converting web3 instances to blobs of bytes or hex, checksums, and timestamps.

Based on our discussion earlier today, `consumable.py` is separate from the credential-related logic so that users don't need to import the whole `AddressCredential` or `DDO` classes. I'm fine with this.

ocean_lib/common/ddo/ddo.py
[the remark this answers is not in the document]
DMats: Fair enough. In that case, I recommend prepending an underscore like so:  `_get_allowed_code` to hint that this method is meant for internal use. I based this name on the latest changes which have already removed the "address" part of the method name

ocean_lib/common/agreements/consumable.py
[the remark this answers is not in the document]
DMats: Fair enough.

ocean_lib/data_provider/data_service_provider.py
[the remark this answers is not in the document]
DMats: I added a note to the description about how this PR Cleans up #277

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if allowed_addresses and address.lower() not in allowed_addresses:
+            return ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST
+
+        if not allowed_addresses and address.lower() in denied_addresses:
+            return ConsumableCodes.CREDENTIAL_IN_DENY_LIST
+
+        return ConsumableCodes.OK
+
+    def add_address_to_list_class(self, address, list_class="allow"):
```
DMats, line 45: Consider standardizing all uses of "list_class" to "access_class" because it's more descriptive.

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+        """Get a filtered list of addresses from credentials (use with allow/deny)."""
+        entries = self.asset._credentials.get(access_class, [])
+
+        for entry in entries:
+            if entry["type"] == "address":
+                return [val.lower() for val in entry["values"]]
+
+        return []
+
+    def requires_credential(self):
```
DMats, line 22: Add docstrings.

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+    if not credential:
+        return None
+
+    if not credential.get("value"):
+        raise MalformedCredential("Received empty address.")
+
+    return credential["value"]
+
+
+def get_list_entry_with_type_address(types_list):
```
DMats, line 105: I think this should be a class method instead of a module function. And it should match the conventions set by `get_addresses_of_class`.  In fact, `get_addresses_of_class` should use this.

```python
    def get_address_entry_of_class(self, access_class="allow"):
        """Get a filtered list of addresses of a given access_class from credentials (use with allow/deny)."""
        entries = self.asset._credentials.get(access_class, [])
        return [entry for entry in entries if entry["type"] == "address"]

    def get_addresses_of_class(self, access_class="allow"):
        """Get an address entry of a given access class from credentials (use with allow/deny)."""
        address_entry = get_address_entry_of_class(access_class)
        return [val.lower() for val in address_entry["values"]]
```

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+from ocean_lib.common.agreements.consumable import ConsumableCodes, MalformedCredential
+
+
+class AddressCredential:
+    def __init__(self, asset):
+        self.asset = asset
+
+    def get_addresses_of_class(self, access_class="allow"):
```
DMats, line 12: Refactor based on the comment for `get_list_entry_with_type_address`

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+            self.asset._credentials[list_class]
+        )
+
+        if not address_entry:
+            self.asset._credentials[list_class].append(
+                {"type": "address", "values": [address]}
+            )
+            return
+
+        address_entry = address_entry[0]
```
DMats, line 64: Why is this line necessary?

Same question for the identical line in `remove_address_from_list_class`

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,106 @@
...
+        if address not in lc_addresses:
+            return
+
+        lc_addresses.remove(address)
+        address_entry["values"] = lc_addresses
+
+
+def simplify_credential_to_address(credential):
+    if not credential:
+        return None
```
DMats, line 97: Shouldn't a MalformedCredential error be raised at this point as well?

DMats (changes requested):

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,96 @@
...
+#
+from ocean_lib.common.agreements.consumable import ConsumableCodes
+from ocean_lib.common.utils.utilities import (
+    get_list_entry_with_type_address,
+    simplify_credential_to_address,
+)
+
+
+class AddressCredential:
+    def __init__(self, asset):
```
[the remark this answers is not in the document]
calina-c: I think it can be Asset too. I don't want to risk it for now 😟

ocean_lib/common/ddo/credentials.py
calina-c: I didn't pluralise corectly. I need to test that there is an address type before I modify the one addres type. It will be clearer after my next commit.

ocean_lib/common/ddo/credentials.py
calina-c: no "Credential" means no dictionary as well, i.e. there is no address credential, nor other type of credential. In the future we might chain multiple credentials, so I check for the "requires address" separately.

DMats: I don't understand your last comment. It sounds like you're talking about the `credentials` (plural) data structure described by https://github.com/oceanprotocol/multi-repo-issue/issues/98.

The `simplify_credential_to_address` function extracts the "value" from a `credential` (singular) which looks like `{"type": "something", "value": "something"}`.  If the caller passes anything that doesn't match that schema, this function should raise a `MalformedCredential` error.

ocean_lib/common/utils/utilities.py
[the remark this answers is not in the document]
DMats: I'll mark this resolved because I see `simplify_credential_to_address` was moved to `ocean_lib/common/ddo/credentials.py`

ocean_lib/common/ddo/credentials.py
DMats: My apologies, I just realized my previous comment mixed up the docstrings.

ocean_lib/common/ddo/credentials.py
DMats: Ohh I see now.  `get_address_entry_of_class` currently returns a list from which you extract the 0th entry.

I think `get_address_entry_of_class` should extract the 0th entry before returning because there should only ever be 1 address entry in a given access class list.

```python
    def get_address_entry_of_class(self, access_class="allow"):
        """Get the address entry of a given credentials access class list. access_class = "allow" or "deny"."""
        entries = self.asset._credentials.get(access_class, [])
        address_entries = [entry for entry in entries if entry["type"] == "address"]
        return address_entries[0] if address_entries else None
```

ocean_lib/common/ddo/credentials.py
DMats: I'll just resolve this comment because I think it's become stale compared to the latest commits.

ocean_lib/common/ddo/credentials.py
DMats: Resolving because this comment has become stale. I'll mention it in my next review.

ocean_lib/common/ddo/ddo.py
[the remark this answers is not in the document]
DMats: I'm resolving this comment because this method is no longer internal because it's now located in `credentials.py` but it's called from `ddo.py`. Thus the prepended underscore no longer makes sense.

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,112 @@
...
+        address_entry = address_entries[0]
+        lc_addresses = self.get_lc_addresses_from_entry(address_entry)
+
+        if address not in lc_addresses:
+            return
+
+        lc_addresses.remove(address)
+        address_entry["values"] = lc_addresses
+
+    def get_address_entry_of_class(self, access_class="allow"):
```
DMats, line 94: I think `get_address_entry_of_class` should extract the 0th entry before returning because there should only ever be 1 address entry in a given access class list.

Also I just realized that I mixed up the docstrings between `get_address_entry_of_class` and `get_lc_addresses_from_entry`. So I decided to just re-write it to be a little more clear.

```python
    def get_address_entry_of_class(self, access_class="allow"):
        """Get address credentials entry of the specified access class. access_class = "allow" or "deny"."""
        entries = self.asset._credentials.get(access_class, [])
        address_entries = [entry for entry in entries if entry["type"] == "address"]
        return address_entries[0] if address_entries else None
```

ocean_lib/common/ddo/credentials.py
```diff
@@ -0,0 +1,112 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+from ocean_lib.common.agreements.consumable import ConsumableCodes, MalformedCredential
+
+
+class AddressCredential:
+    def __init__(self, asset):
+        self.asset = asset
+
+    def get_addresses_of_class(self, access_class="allow"):
```
DMats, line 12: `get_addresses_of_class` and `get_lc_addressess_from_entry` return the same thing - the list of allowed or denied addresses (lower case). I think that only `get_addresses_of_class` should exist and it should look like this:

```python
    def get_addresses_of_class(self, access_class="allow"):
        """Get a filtered list of addresses from credentials (use with allow/deny)."""
        address_entry = self.get_address_entry_of_class(access_class)
        return [addr.lower() for addr in address_entry["values"]] if address_entry else []
```

DMats (changes requested):

ocean_lib/common/ddo/credentials.py
calina-c: The credential can be empty and the address can be None, it still needs checking. E.g. explicitly or implicitly passing empty. We still have to check if they are allowed. The case is covered by the same functions for allow and deny. 

Maybe I can't "see" these anymore, since I've spun them around too much. I'll look into it this morning. I'll submit answers for the other two comments.

alexcos20 (approved):
lgtm

calina-c:
@DMats merging in order to finish provider work. If anything is pending or reappears, I'll address it in a new PR.

PR #315, opened 2021-05-06
Add unit tests for ocean.py code base v3.0

### LINKED ISSUE #284: Add unit tests and improve code coverage (part 2)

mariacarmina:
Fixes #284 .

Changes proposed in this PR:

- added unit tests for ocean.py code base;
- refactored a tweak regarding the logic of if-else blocks in `ocean_assets`;
- added assertion error messages wherever possible.


mariacarmina:
I noticed that in `metadata.py` file, there are `create` and `update` functions which are already tested in `test_ddo.py` from the same module at line 58 and 78. I don't know if it's necessary basically to repeat the same chunks of code, because if I will implement in `test_metadata.py`, the workflow will be the same and redundant. Any guidance here, please? @calina-c @DMats

mariacarmina:
Any thoughts why the `test_register_asset` test fails even though I didn't modify it at all? @calina-c
And when I tested locally, it worked.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -230,6 +230,10 @@ def create(
             if owner_address:
                 data_token.proposeMinter(owner_address, from_wallet=publisher_wallet)
         else:
+            if not dtfactory.verify_data_token(data_token_address):
```
mariacarmina, line 234: When I have implemented unit tests for the exceptions from `exceptions.py`, I have noticed here that we didn't verify in the first place if the contract actually exists. And I have switched some conditions here, because if we want to find the `minter`, firstly we need to check if the contract exists. What are your thoughts here @trentmc? 
P.S. : Don't want to break the interface again :(

ocean_lib/assets/test/test_asset_resolver.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_bad_resolved_asset(publisher_ocean_instance, metadata):
+    publisher = get_publisher_wallet()
+    metadata_copy = metadata.copy()
+
+    asset = publisher_ocean_instance.assets.create(metadata_copy, publisher)
+    wait_for_ddo(publisher_ocean_instance, asset.did)
+    assert asset is not None, "The asset is not cached."
+
+    try:
```
calina-c, line 40: I would suggest using assert pytest.raises() instead of a try catch. Although I do want to keep the message checking somehow, because that's precise and good.

ocean_lib/models/test/test_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+    config = ConfigProvider.get_config()
+    ddo_address = get_contracts_addresses("ganache", config)[
+        MetadataContract.CONTRACT_NAME
+    ]
+    ddo_registry = MetadataContract(ddo_address)
+
+    # Tested the event properties.
+    assert (
+        ddo_registry.event_MetadataCreated.__dict__["abi"]["name"]
```
calina-c, line 27: can we look into using the object instead of the dict, if possible?

ocean_lib/models/test/test_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+    # Tested get_event_log for create event.
+    original_ddo = create_asset(ocn, alice)
+    assert original_ddo, "create asset failed."
+
+    asset_id = original_ddo.asset_id
+    creation_log = ddo_registry.get_event_log(
+        ddo_registry.EVENT_METADATA_CREATED, block, asset_id, 30
+    )
+    assert creation_log, "no ddo created event."
+    assert creation_log.__dict__["args"]["createdBy"] == alice.address
```
calina-c, line 44: same here. I think it's as easy as creation_log.args.createdBy (but please check)

ocean_lib/models/test/test_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+    assert creation_log.__dict__["event"] == ddo_registry.EVENT_METADATA_CREATED
+
+    # Tested get_event_log for update event.
+    ddo = wait_for_ddo(ocn, original_ddo.did)
+    _ = ocn.assets.update(ddo, alice)
+    updating_log = ddo_registry.get_event_log(
+        ddo_registry.EVENT_METADATA_UPDATED, block, asset_id, 30
+    )
+    assert updating_log, "no ddo updated event."
+    assert updating_log.__dict__["args"]["updatedBy"] == alice.address
```
calina-c, line 54: same here

ocean_lib/models/test/test_order.py
```diff
@@ -0,0 +1,81 @@
...
+        asset.did, alice_wallet.address, sa.index
+    )
+    assert order_requirements
+
+    args = [
+        order_requirements.amount,
+        order_requirements.data_token_address,
+        asset.did,
+        service.index,
+        "0xF9f2DB837b3db03Be72252fAeD2f6E0b73E428b9",
```
calina-c, line 40: What is this, exactly? If it's one of the test keys, please use that instead of hardcoding it.

ocean_lib/models/test/test_order.py
```diff
@@ -0,0 +1,81 @@
...
+        _order_tx_id,
+        alice_ocean.config.downloads_path,
+    )
+    assert len(os.listdir(asset_folder)) >= 1
+    for order_log in dt.get_start_order_logs(alice_ocean.web3):
+        order_log_dict = dict(order_log.args.items())
+        order_log_dict["amount"] = from_base_18(int(order_log.args.amount))
+        order_log_dict["marketFee"] = from_base_18(int(order_log.args.marketFee))
+
+        order = Order(
```
calina-c, line 58: you can create the arguments on the fly, and use ** to pass the dict as arguments. That would be a cleaner solution here. Also, I would name this order_args to be more descriptive.

ocean_lib/ocean/ocean_assets.py
calina-c: I think this is good, but I defer to @trentmc.

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -0,0 +1,537 @@
...
+from tests.resources.ddo_helpers import (
+    get_registered_algorithm_ddo,
+    wait_for_ddo,
+    get_registered_ddo_with_compute_service,
+)
+from tests.resources.helper_functions import get_publisher_wallet
+
+
+def test_build_cluster_attributes():
+    data_provider = DataServiceProvider
```
calina-c, line 22: please use the class directly, do not assign the class to a variable, because it looks like an instance. Same for below :)

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -0,0 +1,537 @@
...
+        and container_dict["entrypoint"] == "entrypoint.exe"
+    ), "Entrypoint does not belong to the attributes."
+
+
+def test_build_server_attributes():
+    data_provider = DataServiceProvider
+    config = Ocean.config
+    compute = OceanCompute(config=config, data_provider=data_provider)
+    server_dict = compute.build_server_attributes(
+        "123", "foo_server_type", 4, 4, "20", "20", 30
```
calina-c, line 64: what are these attributes? Can you please make them more deescriptive? E.g. if you have the serverID, input "test_server_id_123" instead of "123"

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -0,0 +1,537 @@
...
+
+def test_build_service_provider_attributes():
+    data_provider = DataServiceProvider
+    config = Ocean.config
+    compute = OceanCompute(config=config, data_provider=data_provider)
+
+    cluster_dict = compute.build_cluster_attributes(
+        "Kubernetes", "http://10.0.0.17/my_cluster"
+    )
+    assert cluster_dict, "Cluster dictionary is None."
```
calina-c, line 99: can you please collect these asserts together? I would create a dictionary containing the expected values and compare the entire dictionary instead of comparing each item separately.

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -0,0 +1,537 @@
...
+    assert algorithm_ddo is not None
+
+    privacy_dict = compute.build_service_privacy_attributes(
+        trusted_algorithms=[algorithm_ddo.did],
+        allow_raw_algorithm=True,
+        allow_all_published_algorithms=True,
+        allow_network_access=True,
+    )
+
+    assert privacy_dict, "Privacy dictionary is None."
```
calina-c, line 199: same suggestion here.

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -0,0 +1,537 @@
...
+    assert isinstance(
```
calina-c, line 285: and here.

ocean_lib/ocean/test/test_util.py
```diff
@@ -121,12 +134,75 @@ def test_get_contracts_addresses():
...
+    assert address == addresses["Ocean"]
+
+
+def test_init_components():
+    config = Config(os.getenv(ENV_CONFIG_FILE))
+    ConfigProvider.set_config(config)
+    init_components()
+    assert ConfigProvider.get_config() == config
+    assert (
+        get_web3_connection_provider(config.network_url).__dict__["endpoint_uri"]
```
calina-c, line 205: let's try to remove __dict__ from actual assertions. We use dict as a simple debugging tool, but we should avoid it for assertions.

ocean_lib/ocean/util.py
```diff
@@ -169,5 +169,7 @@ def init_components(config=None):
         config = Config(os.getenv(ENV_CONFIG_FILE))
 
     ConfigProvider.set_config(config)
+    import pdb
```
calina-c, line 172: please remove debugging statements when committing.

calina-c (changes requested):
Good job! Only minor changes required.

ocean_lib/models/test/test_order.py
mariacarmina: It's not one of the test private keys, I retrieve it from the `test_market_flow`

ocean_lib/assets/test/test_asset_resolver.py
```diff
@@ -0,0 +1,45 @@
...
+    asset = publisher_ocean_instance.assets.create(metadata_copy, publisher)
+    wait_for_ddo(publisher_ocean_instance, asset.did)
+    assert asset is not None, "The asset is not cached."
+
+    resolved_asset = resolve_asset(
+        asset.did,
+        publisher_ocean_instance.config.metadata_cache_uri,
+        asset.data_token_address,
+    )
+    assert isinstance(asset, Asset), "The asset does not have Asset instance."
```
calina-c, line 26: I would move this line before asset resolving, or delete it altogether.

ocean_lib/assets/test/test_asset_resolver.py
```diff
@@ -0,0 +1,45 @@
...
+    assert isinstance(resolved_asset, Asset), "The asset does not have Asset instance."
```
calina-c, line 27: I woul dsuggest the following message: "the resolved asset is not an instance of Asset."

ocean_lib/models/test/test_order.py
calina-c: can you please investigate it? Why is it this address specifically? If it's a random address serving for test purposes, let's define it and reuse it between test_market_flow and test_order.

ocean_lib/models/test/test_order.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    assert len(os.listdir(asset_folder)) >= 1, "The asset folder is empty."
+    for order_log in dt.get_start_order_logs(alice_ocean.web3):
+        order_log_dict = dict(order_log.args.items())
+        order_log_dict["amount"] = from_base_18(int(order_log.args.amount))
+        order_log_dict["marketFee"] = from_base_18(int(order_log.args.marketFee))
+
+        order_args = [
+            order_log.address,
+            order_log_dict["amount"],
```
calina-c, line 60: This test is still not clear to me. It appears to be comparing the order logs with the order logs. I think it is enough to assert just that the order can be created. It seems like this is a very complicated test and I might be missing the point. We can discuss it during a call, too.

calina-c (changes requested):

ocean_lib/models/test/test_order.py
mariacarmina: Never mind, just tested with `alice_wallet.address` and it works fine. Mystery resolved.

calina-c (approved):

ocean_lib/ocean/ocean_assets.py
trentmc: For a given unit test that uses a contract, it will fail if the contract doesn't exist. This is fine. You don't need to make it more explicit.

FYI dozens of unit tests assume the contract exists. If the contracts don't exist, dozens of unit tests will fail. This is fine.

PR #322, opened 2021-05-20
Remove "tests.resources" from mint_fake_ocean & add wallets depending test keys for marketplace quickstart 

### LINKED ISSUE #324: Marketplace quickstart crashes when minting fake OCEAN
When I try to reproduce the market quickstart, I got stuck on minting part. After the installation of the ocean-lib, I have noticed that minting process crashes because of "tests.resources" module which said that is not found and that means that the wallets such as `ganache wallet` or `publisher_ocean_instance` could not be used.

mariacarmina:
Fixes #324 .

Changes proposed in this PR:

- tested & updated marketplace-flow README;
- fixed some modifications in `mint_fake_ocean` regarding wallets.

### DIFF
--- a/READMEs/marketplace-flow.md
+++ b/READMEs/marketplace-flow.md
@@ -101,6 +101,7 @@ In the work console:
 #set private keys of two accounts
 export TEST_PRIVATE_KEY1=0xbbfbee4961061d506ffbb11dfea64eba16355cbf1d9c29613126ba7fec0aed5d
 export TEST_PRIVATE_KEY2=0x804365e293b9fab9bd11bddd39082396d56d30779efbb3ffb0a6089027902c4a
+export FACTORY_DEPLOYER_PRIVATE_KEY=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58
 
 #start python
 python
--- a/READMEs/marketplace-flow.md
+++ b/READMEs/marketplace-flow.md
@@ -101,6 +101,8 @@ In the work console:
 #set private keys of two accounts
 export TEST_PRIVATE_KEY1=0xbbfbee4961061d506ffbb11dfea64eba16355cbf1d9c29613126ba7fec0aed5d
 export TEST_PRIVATE_KEY2=0x804365e293b9fab9bd11bddd39082396d56d30779efbb3ffb0a6089027902c4a
+
+#needed to mint fake OCEAN
 export FACTORY_DEPLOYER_PRIVATE_KEY=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58
 
 #start python
--- a/ocean_lib/ocean/mint_fake_ocean.py
+++ b/ocean_lib/ocean/mint_fake_ocean.py
@@ -15,11 +15,6 @@ from ocean_lib.web3_internal.utils import from_wei, get_ether_balance
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_provider import Web3Provider
 
-# from tests.resources.helper_functions import (
-#     get_ganache_wallet,
-#     get_publisher_ocean_instance,
-# )
-
 
 def mint_fake_OCEAN():
     """
--- a/ocean_lib/ocean/mint_fake_ocean.py
+++ b/ocean_lib/ocean/mint_fake_ocean.py
@@ -14,10 +14,11 @@ from ocean_lib.web3_internal.transactions import send_ether
 from ocean_lib.web3_internal.utils import from_wei, get_ether_balance
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_provider import Web3Provider
-from tests.resources.helper_functions import (
-    get_ganache_wallet,
-    get_publisher_ocean_instance,
-)
+
+# from tests.resources.helper_functions import (
+#     get_ganache_wallet,
+#     get_publisher_ocean_instance,
+# )
 
 
 def mint_fake_OCEAN():
@@ -33,14 +34,19 @@ def mint_fake_OCEAN():
 
     addresses_file = config.address_file
 
-    ocean = get_publisher_ocean_instance()
+    ocean = Wallet(
+        Web3Provider.get_web3(), private_key=os.environ.get("TEST_PRIVATE_KEY1")
+    )
     web3 = ocean.web3
 
     with open(addresses_file) as f:
         network_addresses = json.load(f)
 
     network = "development"
-    deployer_wallet = get_ganache_wallet()
+    deployer_wallet = Wallet(
+        Web3Provider.get_web3(),
+        private_key=os.environ.get("FACTORY_DEPLOYER_PRIVATE_KEY"),
+    )
 
     OCEAN_token = DataToken(address=network_addresses[network]["Ocean"])
 


mariacarmina:
Note: To actually work, we need a pypi release for that.

calina-c (dismissed):
Approving, but it should also be seen by @trentmc.
Reasons for this change: the marketplace readme relied on files inside "tests" folder, which is outside the scope of the pypi release. It is corrected now.

READMEs/marketplace-flow.md
```diff
@@ -101,6 +101,7 @@ In the work console:
 #set private keys of two accounts
 export TEST_PRIVATE_KEY1=0xbbfbee4961061d506ffbb11dfea64eba16355cbf1d9c29613126ba7fec0aed5d
 export TEST_PRIVATE_KEY2=0x804365e293b9fab9bd11bddd39082396d56d30779efbb3ffb0a6089027902c4a
+export FACTORY_DEPLOYER_PRIVATE_KEY=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58
```
trentmc, line 104: Why is this added? Is it deploying a new factory? It shouldn't be - it should be using the contracts already deployed, with addresses specified in addresses.json.

mariacarmina:
The reason why diff-coverage test is failling is because I cannot add an unit test for this type of change.

trentmc (changes requested):
This PR doesn't seem to have anything to do with the issue of https://github.com/oceanprotocol/provider/issues/54 "Provider: ocean.py marketplace quickstart hangs when downloading >1 files".

Rather, it seems to be fixing some apparent problem in marketplace quickstart. 

That's fine. But then first you need to report the problem as a new issue in ocean.py. And point to the issue in this PR (change the title, etc). Then use this PR to point the the bug that it's actually fixing. And then describe how the changes fix the bug.

Finally, I don't see evidence that the issue of https://github.com/oceanprotocol/provider/issues/54 "Provider: ocean.py marketplace quickstart hangs when downloading >1 files" was actually reproduced, let alone in tests going forward. Please reproduce it by writing a unit/system test that has >1 file service for the same ERC20 token.  (Like it says in the issue description: "write a unit or system test on >1 file.")

READMEs/marketplace-flow.md
mariacarmina: I saw that the ganache wallet uses this key, so that's why I have exported it before. Should I replace it?

trentmc: I can't comment more on this PR until I see a github issue reporting what this PR is trying to fix.

trentmc: Thanks for creating the github issue, and linking to it here. +1

It doesn't seem logical that someone using a quickstart should have to point to the private key of some deployer. The only private keys a quickstart should really have are those of the user. So, something's up. You'll need to get to the bottom of this.

trentmc: I thought more about it. This is really here in order to mint fake ocean. Alright, fair enough.

But you need to make it more clear why that private key is being exported. For example, this part of the README could be like:
```console
#set private keys of two accounts
export TEST_PRIVATE_KEY1=0xbbfbee4961061d506ffbb11dfea64eba16355cbf1d9c29613126ba7fec0aed5d
export TEST_PRIVATE_KEY2=0x804365e293b9fab9bd11bddd39082396d56d30779efbb3ffb0a6089027902c4a

#needed to mint fake OCEAN
export FACTORY_DEPLOYER_PRIVATE_KEY=0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58

#start python
python
```

trentmc (changes requested):
Small tweak requested, mentioned in a comment. Then we're good to go.

READMEs/marketplace-flow.md
mariacarmina: Ok, will modify right now.

trentmc (approved):

trentmc:
Approved from my side now too.

calina-c:
> This PR doesn't seem to have anything to do with the issue of [oceanprotocol/provider#54](https://github.com/oceanprotocol/provider/issues/54) "Provider: ocean.py marketplace quickstart hangs when downloading >1 files".
> 
> Rather, it seems to be fixing some apparent problem in marketplace quickstart.
> 
> That's fine. But then first you need to report the problem as a new issue in ocean.py. And point to the issue in this PR (change the title, etc). Then use this PR to point the the bug that it's actually fixing. And then describe how the changes fix the bug.
> 
> Finally, I don't see evidence that the issue of [oceanprotocol/provider#54](https://github.com/oceanprotocol/provider/issues/54) "Provider: ocean.py marketplace quickstart hangs when downloading >1 files" was actually reproduced, let alone in tests going forward. Please reproduce it by writing a unit/system test that has >1 file service for the same ERC20 token. (Like it says in the issue description: "write a unit or system test on >1 file.")

I agree to adding the extra ticket. But I also don't know what else Maria can test. We can't have unit tests on the marketplace readme since it implies the market components. And we would basically do marketplace_flow again. I also understood the scope of this ticket was to test the multiple files case using the README and check if it reproduces. @mariacarmina You did this, right?

We might be looking at a fluke in provider that doesn't reproduce at all, or a speed issue, or an issue related to asset availability. (i.e. the hosting for the asset or assets was compromised temporarily). The question is how much effort should we put into this by this point.

mariacarmina:
> > This PR doesn't seem to have anything to do with the issue of [oceanprotocol/provider#54](https://github.com/oceanprotocol/provider/issues/54) "Provider: ocean.py marketplace quickstart hangs when downloading >1 files".
> > Rather, it seems to be fixing some apparent problem in marketplace quickstart.
> > That's fine. But then first you need to report the problem as a new issue in ocean.py. And point to the issue in this PR (change the title, etc). Then use this PR to point the the bug that it's actually fixing. And then describe how the changes fix the bug.
> > Finally, I don't see evidence that the issue of [oceanprotocol/provider#54](https://github.com/oceanprotocol/provider/issues/54) "Provider: ocean.py marketplace quickstart hangs when downloading >1 files" was actually reproduced, let alone in tests going forward. Please reproduce it by writing a unit/system test that has >1 file service for the same ERC20 token. (Like it says in the issue description: "write a unit or system test on >1 file.")
> 
> I agree to adding the extra ticket. But I also don't know what else Maria can test. We can't have unit tests on the marketplace readme since it implies the market components. And we would basically do marketplace_flow again. I also understood the scope of this ticket was to test the multiple files case using the README and check if it reproduces. @mariacarmina You did this, right?
> 
> We might be looking at a fluke in provider that doesn't reproduce at all, or a speed issue, or an issue related to asset availability. (i.e. the hosting for the asset or assets was compromised temporarily). The question is how much effort should we put into this by this point.

Yep, I did.

trentmc:
> But I also don't know what else Maria can test. 

It's: a unit/system test that has >1 file service for the same ERC20 token. 

(This is not in our current marketplace README, nor does it need to be there. We simply need to test this)

calina-c:
> > But I also don't know what else Maria can test.
> 
> It's: a unit/system test that has >1 file service for the same ERC20 token.
> 
> (This is not in our current marketplace README, nor does it need to be there. We simply need to test this)
> 
> BTW that work is outside the scope of this PR, it's for [oceanprotocol/provider#54](https://github.com/oceanprotocol/provider/issues/54) . We should move our discussion either to that ticket or to a PR/branch working on addressing that ticket.

Got it. So we'll add a new test in the marketplace flow.
Btw ref CodeClimate checks, I think it's broken today. It didn't analyse some provider branches either.

PR #323, opened 2021-05-20
Specific exceptions for ocean.py

### LINKED ISSUE #230: Non-specific errors hinder maintenance

mariacarmina:
Fixes #230  .

Changes proposed in this PR:

- added some specific exceptions where it was possible.


ocean_lib/web3_internal/contract_base.py
```diff
@@ -120,7 +121,7 @@ def get_tx_receipt(tx_hash: str, timeout=20):
                 f"ConnectionClosed error waiting for transaction receipt failed: {e}."
             )
             raise
-        except Exception as e:
```
calina-c, line 123: please keep the previous arrangement here and in contract.py. The point is that any exception thrown by the web3 checks, will be caught and raised as VeriftTxFailed.

mariacarmina: Ok, will modify that right now. So all these web3 checks should be caught as VerifyTxFailed.

ocean_lib/common/ddo/ddo.py
```diff
@@ -20,6 +20,7 @@
 from .constants import DID_DDO_CONTEXT_URL, PROOF_TYPE
 from .public_key_rsa import PUBLIC_KEY_TYPE_RSA, PublicKeyRSA
 from .service import Service
+from ...exceptions import AquariusError
```
calina-c, line 23: please do not use relative imports. The reason why relative imports exist here in common, is that we imported common-utils, and we didn't do every SWE thing in the book for time purposes.

Minimally: import the exception using it's full path
For extra effort: adjust older imports to be absolute instead of relative :)

ocean_lib/common/ddo/ddo.py
```diff
@@ -122,7 +123,7 @@ def encrypted_files(self):
 
     def assign_did(self, did):
         if self._did:
-            raise AssertionError('"did" is already set on this DDO instance.')
+            raise AquariusError('"did" is already set on this DDO instance.')
```
calina-c, line 126: I don't see this as an Aquarius Error. It does not make any call to Aquarius. AssertionError is ok.

ocean_lib/common/ddo/test/test_common_ddo.py
```diff
@@ -11,13 +11,16 @@
...
 )
+from ocean_lib.common.aquarius.aquarius import Aquarius
 from ocean_lib.common.ddo.ddo import DDO
 from ocean_lib.common.ddo.public_key_base import PublicKeyBase
 from ocean_lib.common.ddo.public_key_rsa import (
     PUBLIC_KEY_TYPE_ETHEREUM_ECDSA,
     PUBLIC_KEY_TYPE_RSA,
 )
 from ocean_lib.common.did import DID
+from ocean_lib.config import Config
```
calina-c, line 22: for me, here, there appear to be some imports. Are they used anywhere? Can you please make sure you do not import things which are not used?

ocean_lib/ocean/mint_fake_ocean.py
```diff
@@ -14,10 +14,11 @@
 from ocean_lib.web3_internal.utils import from_wei, get_ether_balance
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_provider import Web3Provider
-from tests.resources.helper_functions import (
-    get_ganache_wallet,
-    get_publisher_ocean_instance,
-)
+
```
calina-c, line 17: this is part of another PR. Please keep things separate for the purpose of simple code review.

ocean_lib/web3_internal/contract_base.py
```diff
@@ -12,6 +12,7 @@
 from eth_typing import BlockIdentifier
 from hexbytes import HexBytes
 from ocean_lib.enforce_typing_shim import enforce_types_shim
+from ocean_lib.exceptions import VerifyTxFailed
```
calina-c, line 15: here as well, there's an unused import.

ocean_lib/web3_internal/web3_overrides/contract.py
```diff
@@ -127,7 +128,7 @@ def wait_for_tx(tx_hash, web3, timeout=30):
         try:
             web3.eth.waitForTransactionReceipt(tx_hash, timeout=timeout)
             break
-        except Exception:
+        except VerifyTxFailed:
```
calina-c, line 131: please revert this, as discussed previously.

ocean_lib/common/ddo/ddo.py
mariacarmina: Ok, I will keep with `AssertionError`. I just thought that the did is resolved via Aquarius. I misunderstood.

ocean_lib/web3_internal/contract_base.py
calina-c: @mariacarmina this is still unused, I think. Maybe you missed this comment :)

mariacarmina: A ok, sorry, missed that one.

ocean_lib/common/aquarius/aquarius.py
```diff
@@ -135,7 +135,7 @@ def text_search(self, text, sort=None, offset=100, page=1):
         if response.status_code == 200:
             return self._parse_search_response(response.content)
         else:
-            raise Exception(f"Unable to search for DDO: {response.content}")
+            raise ValueError(f"Unable to search for DDO: {response.content}")
```
calina-c, line 138: the code diff is failing because the tests are still checking for a generic exception. Take care to adapt the tests too! Instead of catching an Exception, you need to catch a Value Error now, where this is tested.

calina-c (changes requested):

ocean_lib/models/data_token.py
```diff
@@ -350,7 +351,7 @@ def verify_order_tx(self, web3, tx_id, did, service_id, amount_base, sender):
             )
 
         if tx_receipt.status == 0:
-            raise AssertionError("order transaction failed.")
+            raise VerifyTxFailed("order transaction failed.")
```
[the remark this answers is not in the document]
calina-c: pinging here :)

ocean_lib/common/aquarius/aquarius.py
mariacarmina: Should I keep it like that, or revert to `Exception`?

ocean_lib/models/data_token.py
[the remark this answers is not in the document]
mariacarmina: I will leave as it has been before.

ocean_lib/models/data_token.py
```diff
@@ -13,6 +13,7 @@
 from ocean_lib.common.http_requests.requests_session import get_requests_session
 from ocean_lib.data_provider.data_service_provider import DataServiceProvider
 from ocean_lib.enforce_typing_shim import enforce_types_shim
+from ocean_lib.exceptions import VerifyTxFailed
```
calina-c, line 16: you have added an import here, but it is not needed. Please remove.

ocean_lib/ocean/util.py
```diff
@@ -55,8 +55,6 @@ def get_infura_url(infura_id, network):
     if conn_type == "websocket":
         return f"wss://{network}.infura.io/ws/v3/{infura_id}"
 
-    raise AssertionError(f"Unknown connection type {conn_type}")
```
calina-c, line 58: why was this removed?

mariacarmina: I saw that by default the connection type was set to `http`, when the type is neither `http`, neither `websocket`. But I will revert it.

Issue #225, opened 2021-03-16
Move plecos, do basic cleanups


trentmc:
[plecos](https://github.com/oceanprotocol/plecos) has its own pypi version, which makes maintenance a pain.

TODOs for this issue:
- [x] investigate move plecos into ocean.py/plecos or aquarius/plecos
- [x] copy/repurpose tests from plecos
- [x] rename "plecos" -> something else, anything else
- [x] do the low hanging fruit sw eng on plecos
- [x] (optional) Improve unit tests to get >90% coverage -> Călina: moved it to aquarius, where both code-diff and total coverage checks from CodeClimate are passing. I marked it as done, although with a different "meaning".
- [x] adapt/remove/rework usages from other repositories

Related: similar issue for common-util: #226

trentmc:
@ssallam can you give your thoughts on scenario (1) vs (2) in the description, please?

trentmc:
Close because now with #226

calina-c:
@trentmc Added this into aquarius now, since we needed plecos updates and Alex asked to move it too. I addressed most of the concerns, except for renaming and work in other repos.

Ref. renaming: is it necessary? What would you suggest?
Ref. work in other repos: to be done after we merge stable-aqua.

trentmc:
> Ref. renaming: is it necessary? What would you suggest?
Not strictly necessary, but having a random name is unnecessary complexity and can hinder DX. So I strongly suggest renaming it. E.g. to "validate-metadata" directory, or ideally a shorter name. (Open to suggestions! :)

trentmc:
BTW It's great to see this ticket happening! :)

calina-c:
Done on provider.
In review/discussion on ocean.py.
Blocked by stable-aqua merge in aquarius.

AnaLoznianu:
@alexcos20 Please take a look over this one once we deploy stable aqua

calina-c:
Moved in `aquarius/ddo_checker`. Marking this as closed.

PR #325, opened 2021-05-21
Removed plecos.

calina-c:
Related to #225.

### DIFF
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -49,7 +49,6 @@ from ocean_lib.web3_internal.utils import (
 )
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_provider import Web3Provider
-from plecos import plecos
 
 logger = logging.getLogger("ocean")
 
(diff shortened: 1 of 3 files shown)


calina-c:
@trentmc ref https://github.com/oceanprotocol/ocean.py/issues/225
Alex suggested that for the sake of a unified architecture, we use validation calls instead of plecos for aquarius. Previously, the customised aquarius may have used a different version of plecos and the local and remote validation call would have had different results. I support this change entirely, but unfortunately I also had to break the interface a bit for specificity.

i.e. previously the validation function in ocean.py assets would only return True/False. Now it returns a tuple, firstly True/False and then the list of errors. This way, the user has more visibility. I believe it is worth it to break the interface in this case, but let me know if you have any other suggestion.

alexcos20 (approved):

ocean_lib/ocean/ocean_assets.py
```diff
@@ -171,9 +170,10 @@ def create(
             "dataset",
             "algorithm",
         ), f"Invalid/unsupported asset type {asset_type}"
-        if not plecos.is_valid_dict_local(metadata_copy):
-            errors = plecos.list_errors_dict_local(metadata_copy)
-            msg = f"Metadata has validation errors: {errors}"
+
+        validation_result, validation_errors = self.validate(metadata)
```
trentmc, line 174: Before, the call went from ocean.py -> plecos.

Now it goes from ocean.py -> `aquarius.validate_metadata()`. But aquarius itself uses ocean.py. So ocean.py and aquarius are circularly dependent on each other. Python doesn't complain because one direction is hidden under the REST API. It is a circular dependency nonetheless.

We shouldn't have circular dependencies.

Why not just move the functionality that's inside `aquarius.validate_metadata()` into ocean.py?

trentmc (changes requested):
Remove the circular dependencies.

It should be aquarius depending on ocean.py. Without any ocean.py uses of aquarius.

trentmc (approved):

PR #336, opened 2021-05-27
[WIP] Remove ConfigProvider, many unit tests are broken

### LINKED ISSUE #329: Remove ConfigProvider

DMats:
Fixes #329

Changes proposed in this PR:

- 
- 
-


calina-c (commented):
I love the simplification. Why haven't we done this earlier?

Is there any discussion on how this will affect other components? I am working on removing ocean.py from Aquarius, but is it a blocker? Does the Provider (repo) use the ConfigProvider (class)?

DMats:
>I love the simplification. Why haven't we done this earlier?

I love the simplification too!  I'm guessing this change wasn't done earlier because no one was looking at it or it just wasn't high on the priority list.

>Is there any discussion on how this will affect other components? 

Aquarius uses `ConfigProvider`, and `get_contracts_addresses` which relies on `ConfigProvider`. So yes, removing `ConfigProvider` will affect Aquarius, but the effect is small.

> I am working on removing ocean.py from Aquarius, but is it a blocker?

I believe your work to remove ocean.py from Aquarius is going to also remove `ConfigProvider` so it has a low risk of directly conflicting with what I'm doing. So no, I wouldn't call it a blocker.

> Does the Provider (repo) use the ConfigProvider (class)?

No, Provider doesn't use the `ConfigProvider`, nor does it use any functions that will be affected by removing `ConfigProvider`.

ocean_lib/common/ddo/ddo.py
```diff
@@ -445,8 +445,12 @@ def is_consumable(
         if self.is_disabled:
             return ConsumableCodes.ASSET_DISABLED
 
-        if with_connectivity_check and not DataServiceProvider.check_asset_file_info(
-            self, provider_uri
+        if (
+            with_connectivity_check
+            and provider_uri
```
DMats, line 450: Please note logic change: Connectivity check is only performed when provider_uri is set.

ocean_lib/ocean/mint_fake_ocean.py
```diff
@@ -16,14 +14,12 @@
 from ocean_lib.web3_internal.web3_provider import Web3Provider
 
 
-def mint_fake_OCEAN():
+def mint_fake_OCEAN(config):
```
calina-c, line 17: any changes to mint_fake_OCEAN must be reflected in the READMEs, specifically the marketplace flow README. Please change that readme in this same PR, and test it thoroughly.

calina-c (changes requested):
I left a couple of comments to address. Otherwise looks good to me.

ocean_lib/common/ddo/ddo.py
DMats: The caller should pass the default provider_uri then. It shouldn't be this method's responsibility to get the default provider_uri.  Reverting the calling convention is not necessary to achieve the behavior you describe and doing so would yield an API that is harder to test and less obvious what all the inputs are.

ocean_lib/ocean/mint_fake_ocean.py
DMats: Good catch!  Thanks.  I'll do this and comment here when I'm done.

DMats: I updated marketplace-flow.md and tested.  `mint_fake_OCEAN` executed successfully. Unfortunately, I ran into an error later in the workflow. 

However, I believe the error is not caused by my changes.  I verified this by repeating the workflow using the `main` branch, and the same error occurred.

Thus, I do not think there are any problems with the changes included in this PR. I created #345 to track the error I observed.

DMats:
This PR includes non-backward-compatible API changes. Since `ocean.py` is still <1.0.0, I believe it is okay to merge this to `main` (as indicated by the PR target branch) and release it with an incremented minor version number.  

If anyone disagrees, please let me know.  @alexcos20 @trentmc @calina-c @AnaLoznianu

trentmc:
This is fine by me.

PR #342, opened 2021-06-03
Removed deprecated functions.

mariacarmina:
Fixes #337 .

Changes proposed in this PR:

- 
- 
-

PR #344, opened 2021-06-07
Replace camelCase with snake_case

### LINKED ISSUE #338: Replace camelCase occurrences with snake_case in ocean.py
Replace camelCase occurrences with snake_case in ocean.py except where the camelCase is a 1:1 match with an underlying smartcontract method. Required in #333.

mariacarmina:
Fixes #338  .

Changes proposed in this PR:

- replace camelCase occurrences which do not reflect Solidity methods.

### DIFF
--- a/READMEs/wallets.md
+++ b/READMEs/wallets.md
@@ -31,7 +31,7 @@ Here's how. In Python:
 
 ```python
 import web3
-private_key = web3.Web3().eth.account.create().key
+private_key = web3.Web3().eth.account.create().privateKey
 ```
 
 ## 2. Where to store private keys
--- a/READMEs/wallets.md
+++ b/READMEs/wallets.md
@@ -30,6 +30,7 @@ ocean-lib includes the [web3.py library](https://web3py.readthedocs.io/en/stable
 Here's how. In Python:
 
 ```python
+
 import web3
 private_key = web3.Web3().eth.account.create().privateKey
 ```
--- a/READMEs/wallets.md
+++ b/READMEs/wallets.md
@@ -31,7 +31,7 @@ Here's how. In Python:
 
 ```python
 import web3
-private_key = web3.Web3().eth.account.create().privateKey
+private_key = web3.Web3().eth.account.create().key
 ```
 
 ## 2. Where to store private keys
--- a/READMEs/wallets.md
+++ b/READMEs/wallets.md
@@ -30,7 +30,6 @@ ocean-lib includes the [web3.py library](https://web3py.readthedocs.io/en/stable
 Here's how. In Python:
 
 ```python
-
 import web3
 private_key = web3.Web3().eth.account.create().privateKey
 ```
--- a/ocean_lib/ocean/ocean_exchange.py
+++ b/ocean_lib/ocean/ocean_exchange.py
@@ -104,7 +104,7 @@ class OceanExchange:
 
         return exchange_id
 
-    def setRate(
+    def set_rate(
         self,
         new_rate: float,
         wallet: Wallet,
--- a/ocean_lib/ocean/ocean_exchange.py
+++ b/ocean_lib/ocean/ocean_exchange.py
@@ -104,7 +104,7 @@ class OceanExchange:
 
         return exchange_id
 
-    def set_rate(
+    def setRate(
         self,
         new_rate: float,
         wallet: Wallet,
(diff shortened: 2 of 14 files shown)


mariacarmina:
NOTE: To find the camelCase, I used `grep -rn <folder_name> -e "\s[a-z]\+\([A-Z]\+[a-z]\+\)\+"`, because my plugin for PyCharm was not very useful and didn't work for me :(

READMEs/wallets.md
```diff
@@ -31,7 +31,7 @@ Here's how. In Python:
 
 ```python
 import web3
-private_key = web3.Web3().eth.account.create().privateKey
+private_key = web3.Web3().eth.account.create().key
```
trentmc, line 34: This is not a change from camelCase -> snake case, it is a change in the usage of a 3rd party tool's API. Don't do it here. If you _do_ want to do it somewhere, put it in a different issue + PR and explain why you're making the change.

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -104,7 +104,7 @@ def create(self, data_token: str, exchange_rate: float, wallet: Wallet) -> str:
 
         return exchange_id
 
-    def setRate(
+    def set_rate(
         self,
```
trentmc, line 108: The instructions are to preserve labels of methods to be 1:1 with Solidity code.

This is a 1:1 for setRate() in fixedRateExchange.sol https://github.com/oceanprotocol/contracts/blob/main/contracts/fixedRate/FixedRateExchange.sol

ocean_lib/ocean/test/test_ocean_exchange.py
```diff
@@ -76,7 +76,7 @@ def test_ocean_exchange(publisher_ocean_instance):
         is True
     ), "buy datatokens failed"
 
-    assert ox.setRate(1.0, bob_wallet, exchange_id=x_id)
+    assert ox.set_rate(1.0, bob_wallet, exchange_id=x_id)
```
trentmc, line 79: See above

trentmc (changes requested):

READMEs/wallets.md
calina-c: @mariacarmina can you please attach the relevant documentation? Has the new version of web3 equated these properties? (I doubt it, maybe it's private_key... But if the web3 version did deprecate privateKey and replaced it with "key", I think we can handle that separately, in #342)

calina-c: @trentmc while I agree there is some overlap between this ticket and the deprecation one, I would rather keep them separate. However, in some cases because the new web3 version deprecated camelCase and replaced it with snake_case, we are fixing two things at once. I have no issue with this overlap, if not blatant. Switching between branches is exhausting, especially for small changes such as casing.

trentmc: Agreed. And: If it's two things at once, then the PR name and github issue should be for the more important issue. Pointing to an updated API is the more important thing. We shouldn't do it as a side effect of something cosmetic like snake case.

calina-c: Addressed here: https://github.com/oceanprotocol/ocean.py/pull/342/files#r647164717
It can be removed from this PR.

mariacarmina: I have also did the same modification in the #342 and I will attach here a relevant snapshot from the code, because that was my documentation at that moment.
Screenshot from 2021-06-08 10-08-54

ocean_lib/ocean/ocean_exchange.py
mariacarmina: Sorry, will modify right now as how it was.

READMEs/wallets.md
mariacarmina: The path for that is `venv/lib/python3.8/site-packages/eth_account/signers/local.py`

mariacarmina: Here is the demo :)
Screenshot from 2021-06-08 10-39-43

trentmc: If the change is already being done in another issue / PR then why is it still in this PR?

trentmc (changes requested):

READMEs/wallets.md
mariacarmina: I discovered the deprecated function in this PR and after I modified in #342 , that was the reason, now I think it's all good. I have modified in the right place.

trentmc (approved):

trentmc (approved):

Issue #345, opened 2021-06-08
Error during ocean.py marketplace-flow


DMats:
An error occurs in the marketplace-flow when executing this line

```pycon
>>> asset = ocean.assets.resolve(did)
```

Full console log:
```console
david@david-Inspiron-5502:~/projects/ocean.py$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
david@david-Inspiron-5502:~/projects/ocean.py$ git pull
Already up to date.
david@david-Inspiron-5502:~/projects/ocean.py$ cd ..
david@david-Inspiron-5502:~/projects$ mkdir test3
david@david-Inspiron-5502:~/projects$ cd test3/
david@david-Inspiron-5502:~/projects/test3$ python -m venv venv
david@david-Inspiron-5502:~/projects/test3$ source venv/bin/activate
(venv) david@david-Inspiron-5502:~/projects/test3$ pip install ../ocean.py/
Processing /home/david/projects/ocean.py
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Collecting enforce-typing==1.0.0.post1
  Using cached enforce_typing-1.0.0.post1-py2.py3-none-any.whl (15 kB)
Collecting requests>=2.21.0
  Using cached requests-2.25.1-py2.py3-none-any.whl (61 kB)
Collecting pyopenssl
  Using cached pyOpenSSL-20.0.1-py2.py3-none-any.whl (54 kB)
Collecting cryptography==3.3.2
  Using cached cryptography-3.3.2-cp36-abi3-manylinux2010_x86_64.whl (2.6 MB)
Collecting pytz
  Using cached pytz-2021.1-py2.py3-none-any.whl (510 kB)
Collecting coloredlogs
  Using cached coloredlogs-15.0-py2.py3-none-any.whl (45 kB)
Collecting web3==5.19.0
  Using cached web3-5.19.0-py3-none-any.whl (470 kB)
Collecting pycryptodomex
  Using cached pycryptodomex-3.10.1-cp35-abi3-manylinux2010_x86_64.whl (1.9 MB)
Collecting tqdm
  Using cached tqdm-4.61.0-py2.py3-none-any.whl (75 kB)
Collecting PyYAML==5.4.1
  Using cached PyYAML-5.4.1-cp38-cp38-manylinux1_x86_64.whl (662 kB)
Collecting scipy
[...]

PR #350, opened 2021-06-17
Install wheel first to eliminate benign errors while installing ocean-lib

DMats:
Towards https://github.com/oceanprotocol/ocean.py/issues/345#issuecomment-863375587

Changes proposed in this PR:

- Install wheel before ocean-lib in all the READMEs
- Remove wheel from install_requirements because (1) it's not strictly required and (2) installing it like this doesn't fix the errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,9 @@ This is in beta state and you can expect running into problems. If you run into 
 ## 🏗 Installation
 
 ```console
+#Install the ocean.py library. Install wheel first to avoid errors.
 pip install wheel ocean-lib
 ```
-
-NOTE: `wheel` is not strictly necessary, but installing it first eliminates
-benign error messages that occur while installing `ocean-lib`.
 
 ## 🏄 Quickstart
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -47,7 +47,12 @@ This is in beta state and you can expect running into problems. If you run into
 
 ## 🏗 Installation
 
-`pip install ocean-lib`
+```console
+pip install wheel ocean-lib
+```
+
+NOTE: `wheel` is not strictly necessary, but installing it first eliminates
+benign error messages that occur while installing `ocean-lib`.
```
trentmc, line 55: This takes up too much real estate early on. I recommend skipping the "NOTE:" and instead have it like:
```console
#including wheel avoids errors 
pip install wheel ocean-lib
```

The same for the changes below.

trentmc (changes requested):

README.md
DMats: Done.

calina-c (approved):

trentmc (approved):

PR #352, opened 2021-06-23
Injected chainId for EIP-155

### LINKED ISSUE #246: EIP-155
With Geth v1.10.0, EIP-155 is enforced (see https://blog.ethereum.org/2021/03/03/geth-v1-10-0/, chapter "ChainID enforcement")

All transactions need to follow https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md or they will be rejected in the future (Infura runs v1.10.0, but with --rpc.allow-unprotected-txs, so we are safe for the moment ).

All sending transactions should include chainId as well.

mariacarmina:
Fixes #246 .

Changes proposed in this PR:

- added `chainId` key to the transactions dicts.


mariacarmina:
NOTE: Right now, I need to test also the marketplace-flow README with the local geth node. But it needs time to sync with the current block from ropsten, that's why I will leave as a draft PR.

ocean_lib/web3_internal/transactions.py
```diff
@@ -30,14 +31,10 @@ def send_ether(from_wallet: Wallet, to_address: str, ether_amount: int):
         "from": from_wallet.address,
         "to": to_address,
         "value": w3.toWei(ether_amount, "ether"),
+        "chainId": w3.eth.chain_id,
     }
-    _ = w3.eth.estimate_gas(tx)
```
calina-c, line 34: I am concerned with the removal of the tx assignment, can I ask you why you did it? It seems not connected with chainId at all.

ocean_lib/web3_internal/web3_overrides/contract.py
```diff
@@ -49,6 +50,9 @@ def transact(self, transaction):
                 raise ValueError(
                     "Please ensure that this contract instance has an address."
                 )
+        if "chainId" not in transact_transaction:
+            chain_id = cf.web3.eth.chain_id
+            transact_transaction["chainId"] = chain_id
```
calina-c, line 55: I think you can comprise 54 and 55 into one line, and remove the local assignment.

DMats:
We might want to consider adding a unit test or two for this. Also may want to improve doc strings where applicable.

trentmc:
> We might want to consider adding a unit test or two for this. Also may want to improve doc strings where applicable.

Yes and yes. "It only exists if it has unit tests"

ocean_lib/web3_internal/transactions.py
mariacarmina: It is not connected with `chainId`, you are right, but I thought that it was completely unnecessary, because you try to estimate the gas with `w3.eth.estimate_gas()` of the tx and after that you replace the gas value with 5000 which I don't exactly know where it comes from. The main point, as I discussed with @DMats, was to keep consistency with `estimate_gas` function.

mariacarmina:
Do you think is a good approach for the unit tests @DMats? I tried to retrive the `chainId` but it's not visible after the transaction is signed.

DMats:
@mariacarmina Unit tests look good to me.

ocean_lib/web3_internal/transactions.py
mariacarmina: @trentmc What do you think? Should we keep the initial version with not using the `estimate_gas` value in the tx dict? Or does it seem fine like this?

mariacarmina:
> @mariacarmina Unit tests look good to me.

I will add more. Thanks

ocean_lib/web3_internal/transactions.py
trentmc: If you don't understand why something is there, don't remove it. Especially if it's not in the scope of the issue.

trentmc (changes requested):

ocean_lib/web3_internal/transactions.py
mariacarmina: Ok, I will keep the initial version, I thought it was unnecessary. No problem.

mariacarmina: Update: I try to revert the tx dict with the gas value set to 500000, but it actually affects the chain id value and by default my unit test for `send_ether` function where it was declared in the first place. After setting the gas value up to 500000, I applied to formula for the `chainId` and it turned out to be negative and I didn't see any negative chain ids in other documentation. In conclusion, `web3.eth.estimate_gas` does affect the scope of the task, so it should be modified as above to actually get a valid chain id.

trentmc (approved):

ocean_lib/web3_internal/test/test_utils.py
```diff
@@ -33,6 +34,33 @@ def test_prepare_fixed_hash():
     ), "The address is not the expected one."
 
 
+def test_chain_id_send_ether(alice_wallet, bob_address, web3_instance):
+    """Tests if the chainId has the right value for send ether transactions."""
+    receipt = send_ether(alice_wallet, bob_address, 1)
+    if not receipt:
```
DMats, line 40: Replace with

```python
assert receipt, "Send ether was unsuccessful."
```

ocean_lib/web3_internal/test/test_utils.py
```diff
@@ -33,6 +34,33 @@ def test_prepare_fixed_hash():
...
+    result = True if web3_instance.eth.chain_id in chain_ids else False
+    assert result, "The chain ID is not the right one."
+
+
+def test_chain_id_cancel_or_replace_transaction(
+    alice_wallet, bob_address, web3_instance
+):
+    """Tests if the chainId has the right value for cancelled tx."""
+    receipt_cancelled = cancel_or_replace_transaction(alice_wallet, None)
+    if not receipt_cancelled:
```
DMats, line 54: Replace with

```python
assert receipt_cancelled, "Cancel or replace transaction failed."
```

ocean_lib/web3_internal/transactions.py
DMats: Sending ether from one address to another costs 21000 gas, so arguably we should just set the gas limit to 21000.

That said, I think using the `estimate_gas` function works just as well so I don't think further changes are required. 

According to the docs it "Executes the given transaction locally without creating a new transaction on the blockchain. Returns amount of gas consumed by execution which can be used as a gas estimate."

@mariacarmina I suspect that you got a "negative chain_id" because you accidentally ran `web3.eth.estimate_gas` on a transaction that didn't have `chain_id` in it so it fell back to the old calculation of `v` resulting in `v = 27` or `v = 28`.

ocean_lib/web3_internal/test/test_utils.py
```diff
@@ -33,6 +34,33 @@ def test_prepare_fixed_hash():
     ), "The address is not the expected one."
 
 
+def test_chain_id_send_ether(alice_wallet, bob_address, web3_instance):
+    """Tests if the chainId has the right value for send ether transactions."""
+    receipt = send_ether(alice_wallet, bob_address, 1)
+    if not receipt:
+        raise AssertionError("Send ether was unsuccessful.")
+    tx = web3_instance.eth.get_transaction(receipt["transactionHash"])
```
DMats, line 42: `web3_instance` might have a different chain_id than `alice_wallet.web3`. 

I think you should use `alice_wallet.web3.eth.get_transaction` and `alice_wallet.web3.eth.chain_id`.

Do this for both `test_chain_id_send_ether` and `test_chain_id_cancel_or_replace_transaction`

DMats (changes requested):

ocean_lib/web3_internal/test/test_utils.py
mariacarmina: Ok, will try that right now.

ocean_lib/web3_internal/transactions.py
mariacarmina: I added the `chainId` field to the last tx so we can keep the initial form as @trentmc said before. What do you think @DMats?

DMats: Honestly, I think it should look like one of the following.

```python
tx = {
    "from": from_wallet.address,
    "to": to_address,
    "value": w3.toWei(ether_amount, "ether"),
    "chainId": w3.eth.chain_id,
    "gas": 21000,
}
```

or like this:

```python
tx = {
    "from": from_wallet.address,
    "to": to_address,
    "value": w3.toWei(ether_amount, "ether"),
    "chainId": w3.eth.chain_id,
}
tx["gas"] = web3.eth.estimate_gas(tx)
```

mariacarmina: I have approached the second option, but @trentmc should I keep the second option that @DMats mentioned?

mariacarmina: I can use the second option that you have purposed @DMats, because @trentmc agreed with that in the private conversation as far as I see.

trentmc: I am fine with this.

trentmc (approved):

mariacarmina:
UPDATE: Tested with @DMats the marketplace-flow README, both roles, Alice & Bob, and it works. I have also tested with the web interface of the Ocean Market and I successfully published a data asset [()](http://localhost:8000/asset/did:op:414a98b068254F0D5E93dC2A6d4f832dc6527c19) so, the custom network that points to a geth node that has the EIP155 enforcement works. I think it can be merged after I resolve the conflicts, hopefully. What do you think @trentmc , @DMats , @calina-c ?

mariacarmina:
For this task, the `config.ini` from marketplace-flow README is shorter, because you don't need the `provider.address`, neither the `address.file` and obviously the network needs to be different (I put the URL that points to the geth node).

ocean_lib/web3_internal/contract_base.py
```diff
@@ -227,6 +228,7 @@ def send_transaction(
             "from": from_wallet.address,
             "passphrase": from_wallet.password,
             "account_key": from_wallet.key,
+            "chainId": contract_function._contract_function.web3.eth.chain_id
```
DMats, line 231: I think it's better to pull the Web3 instance from the `ContractBase.w3` instance variable instead of the `CustomContractFunction` you've created here.

Please change to 

```python
"chainId": self.w3.eth.chain_id,
```

ocean_lib/web3_internal/test/test_utils.py
```diff
@@ -33,6 +34,29 @@ def test_prepare_fixed_hash():
     ), "The address is not the expected one."
 
 
+def test_chain_id_send_ether(alice_wallet, bob_address, web3_instance):
```
DMats, line 37: Please move the transactions.py-related unit tests to a new file named `test_transactions.py`

This includes:

* `test_chain_id_send_ether`
* `test_chain_id_cancel_or_replace_transaction`
* `test_send_ether`
* `test_cancel_or_replace_transaction`

This inconsistency existed before your current task, but this PR seems like a sensible place to fix it.

DMats (changes requested):

DMats (approved):

PR #356, opened 2021-06-25
Issue 327 remove artifacts path

### LINKED ISSUE #327: Remove ARTIFACTS_PATH, contract ABI JSON loading, and ContractHandler
Since the ocean-contracts PyPi package now supports loading contract ABI JSON files using `jsonsempai`, there is a lot of code simplification that can be done in ocean.py. See for details: https://github.com/oceanprotocol/multi-repo-issue/issues/83#issuecomment-846274680

- [x] Eliminate JSON loading and validation for ocean ABIs
- [x] Eliminate ARTIFACTS_PATH envvar and artifacts.path config option 
- [x] Eliminate ContractsHandler

calina-c:
Fixes #327.

Changes proposed in this PR:

- 
- 
-


calina-c:
@DMats here's what I did until now:
- I removed ARTIFACTS_PATH from the config and ContractHandler, now it is loaded automatically
- I replaced the read_abi_from_file funciton in ContractHandler with a smart loading of the abi from the artifacts lib path

These issues should also unblock the provider removal of ContractHandler, since artifacts_path does not need to be set anymore, it is smartly loaded.

However, I'm not sure how to continue. Why do we want to remove the entire ContractHandler?

DMats:
The `ContractHandler` is still a singleton because it contains a `_contracts` class variable that acts as a cache. This means that two `Ocean` instances initialized on different chains will affect each other through their shared `ContractHandler` singleton.  

At a minimum, the `_contracts` class variable needs to be removed.

The question becomes: "Should we drop the caching behavior?  Or should we keep the caching behavior by changing `_contracts` into an instance variable?"

I think you should remove the caching behavior.  This would eliminate the following methods:
1. _set()
2. set()
3. has
4. _verifyContractConsistency()

I don't want to keep the caching behavior because then we'd have to pass the `ContractHandler` instance around the codebase as a parameter. This would cause a lot of API changes. Also, caching is a performance optimization. But, AFIK we haven't shown contract loading to be a performance bottleneck. Until we do, caching should be considered a premature optimization.

I had originally recommended removing the ContractHandler because once the caching behavior is gone, the remaining methods (get_contracts_addresses, _get, get, get_concise_contract, _load) could all be combined into a two relatively short methods - something like this.

```python
import Web3

def get_contract_definition(contract_name):
    try:
        return importlib.import_module("artifacts." + contract_name).__dict__
    except ModuleNotFoundError:
        raise TypeError("Contract name does not exist in artifacts.")

def load_contract(web3, contract_name, address=None):
    contract_definition = get_contract_definition(contract_name)
    if not address:
        address = Web3.toChecksumAddress(contract_definition["address"])
    assert address is not None, "address shouldn't be None at this point"

    contract = web3.eth.contract(
    )
    return contract
```

ocean_lib/config.py
```diff
@@ -205,27 +195,13 @@ def _load_environ(self):
...
-        _path_string = self.get(SECTION_ETH_NETWORK, NAME_ARTIFACTS_PATH)
-        path = Path(_path_string).expanduser().resolve() if _path_string else None
-
-        if path and path.exists():
-            return path
-
-        return Path(artifacts.__file__).parent.expanduser().resolve()
-
     @property
     def address_file(self):
```
DMats, line 199: I think `config.address_file` should fall back to the `address.json` included with `ocean-contracts` if not specified via envvar or config option, like so:

```python
import artifacts
...snip...
    @property
    def address_file(self):
        file_path = self.get(SECTION_ETH_NETWORK, NAME_ADDRESS_FILE)
        if file_path:
            file_path = Path(file_path).expanduser().resolve()
        else:
            file_path = Path(artifacts.__file__).parent.expanduser().resolve()
        return str(file_path)
```

This logic should work because `DEFAULT_ADDRESS_FILE = ""` (empty string) which evaluates to `False`.

ocean_lib/web3_internal/utils.py
```diff
@@ -2,9 +2,14 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
-import logging
+import artifacts
 from collections import namedtuple
 from decimal import Decimal
+import importlib
+from jsonsempai import magic  # noqa: F401
```
DMats, line 9: Please remove unused imports

```
import importlib
from jsonsempai import magic  # noqa: F401
from web3 import Web3
```

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -0,0 +1,67 @@
+#
```
DMats, line 1: Please add docstrings to this file.

setup.py
```diff
@@ -31,7 +31,8 @@
     "web3==5.19.0",
     "cryptography==3.3.2",
     "scipy",
-    "enforce-typing==1.0.0.post1"
+    "enforce-typing==1.0.0.post1",
+    "json-sempai==0.4.0",
```
DMats, line 35: Interestingly, if you implement my commments for removing unused imports from `contract_utils.py` and `utils.py`, then nothing in `ocean.py` will be using `json-sempai`.  Should we remove it from the dependency list?

DMats (changes requested):

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def get_contract_definition(contract_name):
+    try:
+        return importlib.import_module("artifacts." + contract_name).__dict__
+    except ModuleNotFoundError:
+        raise TypeError("Contract name does not exist in artifacts.")
+
+
+def load_contract(web3, contract_name, address):
```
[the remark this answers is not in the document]
calina-c: I think it should stay this way, since the others are "regular" utils, and "load" is more descriptive and high-level. I would keep it as such. The "get_contract_concise" is the outlier, actually, but that is set to be deprecated.

setup.py
calina-c: actually it will stay, since the magic I'm doing with importlib uses the json-sempai magic in turn :)

DMats (approved):

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -0,0 +1,57 @@
+#
+# Copyright 2021 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+import importlib
+import json
+import logging
+import os
+
+import artifacts  # noqa
```
DMats, line 10: Strictly speaking, importing `artifacts` is not required.  I tried the unit tests without it and they all passed.

That said, I'm okay with importing it so that the import list presents an accurate picture of this file's dependencies, even those imported using `importlib`.

calina-c: that was basically my reasoning, too :)

PR #362, opened 2021-06-30
Remove contract concise

### LINKED ISSUE #343: Remove contract_concise, replace with ContractCaller
From the previous part #342, the only changes that need to be done are:

- [ ] remove contract concise which is deprecated; Replace it with ContractCaller
Code : `pool = ocean.pool.create( token_address, data_token_amount=100.0, OCEAN_amount=10.0, from_wallet=alice_wallet)`
_Log:_ `BFactory.newBPool(). Done.
/home/maria/Desktop/test3/venv/lib/python3.8/site-packages/web3/_utils/decorators.py:48: DeprecationWarning: __init__ is deprecated in favor of contract.caller.<method name> or contract.caller({transaction_dict}).<method name>`

mariacarmina:
Fixes #343  .

Changes proposed in this PR:

- remove the contract;
- replace with contract caller;
- remove BPool constructor.

### DIFF
--- a/ocean_lib/models/bpool.py
+++ b/ocean_lib/models/bpool.py
@@ -24,7 +24,7 @@ class BPool(BToken):
     def __init__(self, *args, **kwargs):
         """Initialises BPool object."""
         BToken.__init__(self, *args, **kwargs)
-        self._ccontract = self.contract_concise
+        self._ccontract = self.contract.caller
 
     def __str__(self):
         """Formats with attributes as key, value pairs."""
--- a/ocean_lib/models/bpool.py
+++ b/ocean_lib/models/bpool.py
@@ -21,11 +21,6 @@ logger = logging.getLogger(__name__)
 class BPool(BToken):
     CONTRACT_NAME = "BPool"
 
-    def __init__(self, *args, **kwargs):
-        """Initialises BPool object."""
-        BToken.__init__(self, *args, **kwargs)
-        self._ccontract = self.contract.caller
-
     def __str__(self):
         """Formats with attributes as key, value pairs."""
         s = []
@@ -106,7 +101,7 @@ class BPool(BToken):
 
     # ==== View Functions
     def isPublicSwap(self) -> bool:
-        return self._ccontract.isPublicSwap()
+        return self.contract.caller.isPublicSwap()
 
     def isFinalized(self) -> bool:
         """Returns true if state is finalized.
@@ -116,7 +111,7 @@ class BPool(BToken):
         `JOIN`, and `EXIT` are public. `CONTROL` capabilities are disabled.
         (https://docs.balancer.finance/smart-contracts/api#access-control)
         """
-        return self._ccontract.isFinalized()
+        return self.contract.caller.isFinalized()
 
     def isBound(self, token_address: str) -> bool:
         """Returns True if the token is bound.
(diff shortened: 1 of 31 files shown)


mariacarmina:
UPDATE: I modified the unit test `test_dispenser_status`, because after the replacement with `contract.caller` I noticed that the owner get a certain value for `None` which is the `0x0000000000000000000000000000000000000000000000000000000000000000` address.
image

ocean_lib/web3_internal/contract_base.py
```diff
@@ -55,13 +55,13 @@ def __init__(self, address: Optional[str], abi_path=None):
         assert abi_path, f"abi_path is required, got {abi_path}"
 
         self.w3 = Web3Provider.get_web3()
-        self.contract_concise = get_concise_contract(self.w3, self.name, address)
+        self.contract_caller = get_contract_caller(self.w3, self.name, address)
```
calina-c, line 58: do we still need this assignment? If the caller can be accessed using `self.contract.caller` I don't see the point in assigning another instance variable :)

calina-c (changes requested):
Left one suggestion/question. Otherwise LGTM.

ocean_lib/web3_internal/contract_base.py
mariacarmina: This function `get_contract_caller` returns a `ContractCaller` instance with the specific abi, web3 instance and address, as contract_concise was before with `get_concise_contract`. And after I use this variable `contract_caller` in `dispenser`, too. Maybe we can replace with `contract_caller` wherever is possible.
Other than that, if I remove the assignment, pytest hangs (I waited some couple of minutes and nothing happened).

mariacarmina: UPDATE: @calina-c, I removed the assignment for contract_caller and that `get_contract_caller`, too.

ocean_lib/models/bpool.py
```diff
@@ -24,7 +24,7 @@ class BPool(BToken):
     def __init__(self, *args, **kwargs):
```
DMats, line 24: I propose we take this PR as the opportunity to remove the `BPool._ccontract` instance variable and the `BPool.__init__` method. @calina-c Do you agree?

Reasoning:
The `_ccontract` instance variable is a duplicate. The base class `ContractBase` already has a `contract` instance variable. Anywhere BPool uses `self._ccontract` it should instead use `self.contract.caller`

mariacarmina:
I noticed this issue when I reproduced the marketplace-flow README, more exactly when I tried to create a pool, I received a deprecation warning due to contract concise. I will attach here how it looks right now
image

DMats (changes requested):

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -9,7 +9,7 @@
 
 import artifacts  # noqa
 from jsonsempai import magic  # noqa: F401
-from web3.contract import ConciseContract
+from web3.contract import ContractCaller
```
calina-c, line 12: I don't think this import is needed :)

ocean_lib/models/bpool.py
calina-c: I agree. And it's not a big change :)

mariacarmina:
NOTE: Tested again with marketplace-flow README, works just fine.

calina-c (approved):

DMats (approved):

DMats:
@mariacarmina When you merge this, please do a "Squash and merge" so only 1 commit appears in the `main` branch history. The default title of this commit should be fine - it will be the title of the PR along with the PR # in parentheses.

The default body is usually not what you want - it puts all the commits from before the squash into a list of bullet points (including intermediate changes that may not be present in the final commit). I usually erase the default body and replace it with a more complete and succinct description of the changes.  Since I like to keep my PR descriptions up to date, I usually just copy and paste the list of changes from the description into the squash commit body.

PR #363, opened 2021-06-30
Issue 330 remove web3provider

### LINKED ISSUE #330: Remove Web3Provider

DMats:
Fixes #330

Changes proposed in this PR:

  * Remove `Web3Provider` singleton. Propagate web3 instance via parameters 
  * Add `Web3` instance variable to `ContractBase` so its methods can easily
    call RPCs. This means that the `ContractBase.__init__` method now
    requires a Web3 instance passed as a parameter. This affects
    subclasses including: `BFactory`, `BToken`, `BPool`, `DataToken`, `DTFactory`,
    `DispenserContract`, `MetadataContract`, and `MyFactory`.
  * Add web3 instance variable to `OceanAssets`, `OceanPool`, and
    `OceanExchange`. This means that the `__init__` method for these classes
    now require a Web3 instance passed as a parameter.
  * Remove web3 parameters wherever possible:
      * Where `Web3` static or class methods were called using web3
        instances unnecessarily.
      * Where `eth_account` static and class methods were called using
        web3 instances unnecessarily - like `web3.eth.account`
  * Update unit test parameter lists - if a test already included an
    `Ocean` instance as a parameter, then use `ocean.web3` wherever a `Web3`
    instance was required, else, add a web3 parameter to the unit test.
    Care was taken to NOT use the `Web3` from `Wallet` instances or from
    `Contract` instances except where strictly necessary.
  * `EventFilter.uninstall()` now uses the `Web3` instance from its `event`
    instance variable
  * Add type hints and type enforcement anywhere `Web3` parameter added.
  * Fixed incorrect return type hints in `OceanPool`
  * Remove assignments like `w3 = web3`
  * Add `helper_functions.get_example_config()`
  * Remove `init_components` and `test_init_components`
  * Add testcase to `test_resolve_asset`, cleanup logic in `resolve_asset`
  * Add `test_get_network_name` and `test_get_network_id`


calina-c:
I agree with the changes, in theory. However, are we sure we want to pass around a web3 object? I would have assumed an ocean object, that includes web3, would have been more appropriate and extensible. I base this on the premise that, if we want to add more things later, we can avoid changing the interface _again_ and bundle what we need in ocean.

However, I suppose this has been discussed :)
LGTM in all other respects.

mariacarmina (commented):
I wanted to resolve the conflicts for this branch, because I saw that they are related to the EIP 155 task that was assigned for me. Other than that, it looks good from my side, good job!
BTW, @DMats, please tell me if you consider correct the changes that I've made. Thank you!

calina-c (dismissed):

mariacarmina (approved):

mariacarmina (approved):
Great work, @DMats !

PR #364, opened 2021-07-02
Remove RSA Public Key Complexities

### LINKED ISSUE #306: Assess public key base and public key RSA contents, remove if necessary
Formerly in common-utils, ocean.py now contains the files `public_key_base` and `public_key_rsa`. The contents of these files are outdated and unused, according to Alex. But preliminary trials to remove them failed, and there are bigger priorities right now.

calina-c:
Closes #306.


mariacarmina (commented):
Do you think that an unit test could be a bad idea to have for this? @DMats, @calina-c any thoughts?

calina-c:
> Do you think that an unit test could be a bad idea to have for this? @DMats, @calina-c any thoughts?

Added.

mariacarmina (approved):
Nice work @calina-c !

PR #372, opened 2021-07-07
[VERY Draft and WIP] v4 Contracts

calina-c:
Added this as a PR just in case someone needs to know about it. This is still highly WIP and only contains the contract functions transcribed, as of now.


mariacarmina:
I will close this because the main logic is in #512.

PR #380, opened 2021-07-13
Remove ocean pool

### LINKED ISSUE #239: OceanPool isn't needed, so deprecate
NOTE: we might not do this, we need to discuss whether we really want to do it first.

**Motivation**

OceanPool was intended as a thin wrapper for BPool, for convenience: so users wouldn't have to think about many steps to add OCEAN token, etc.

But there are issues:
- It has very poor test coverage. We could write tests but it will take time
- OCEAN is a less obvious default going forward: Ocean Market frontend will be moving to using a stablecoin (eg OceanO), and frontend forks may use other tokens too. 
- We've discussed internally and will move away from all of ocean.py thin wrappers, including this one. Having thin wrappers like this adds complexity and hurts maintainability. You can still have good DX by making modules easy to use, in the same way that web3.py and scikit-learn does.

**TODO for this issue**

- [ ] Deprecate OceanPool and any dependencies

** Notes **

- On timing: doing this causes breaking changes to the ocean.py interface, which means a major version change. So it needs to get shipped in eg "V4".
- #240 is similar, but for the ocean.py convenience methods
- Complementary: better readthedocs-style documentation

calina-c:
Fixes #239.


calina-c:
Another idea would be to have OceanPool inherit from BPool, in order to simplify some things: E.g. instead of having ocean.pool.pool (a BPool) and ocean.pool (an OceanPool). And since OceanPools are essentially BPools that are aware of web3 and ocean_address, it would make sense.

@DMats has some questions regarding this. Can you please paste the concerns from Slack, David?
@trentmc @alexcos20 right now I did the work, since this was marked as a Status:Refined, but there seem to be issues with the philosophy. Alex asked for more thin wrappers, while Trent asked for less. What should we do with OceanPool and Ocean itself?

To quote from a classic:
_Where do we go now?
Where do we go?
Where do we gooo?_

DMats:
Thanks for the ping @calina-c. I am cautious whenever I add inheritance to a codebase because I think it's harder to de-couple than delegation when things change. At first blush - I thought OceanPools are BPools so inheritance makes sense. But then I questioned - will OceanPools *always* be BPools?  With the impending upgrade to Balancer V2, I suspect the answer is no.

That said, I think it's more pressing that we clarify the original intent of #239 - so I think we should continue the conversation there.

calina-c:
Closing this without merging, since v4 takes priority and Trent changed his mind related to removing this.

PR #396, opened 2021-07-23
Typing improvements

### LINKED ISSUE #259: Improve type hints
This issue is a follow-on of https://github.com/oceanprotocol/ocean.py/issues/37

Possible improvements:

- [x] In cases where a function docstring includes type hints, but the function signature does not, add the type hints to the signature.
- [x] Apply the `@enforce_types_shim` decorator to every class and function.

calina-c:
Closes #259.


ocean_lib/common/ddo/ddo.py
```diff
@@ -282,73 +288,77 @@ def get_service_by_index(self, index):
                 return service
 
         # try to find by type
+        # FIXME: this shouldn't work and is possibly a typo
```
DMats, line 291: Have you created an issue for this?

ocean_lib/common/ddo/ddo.py
```diff
@@ -242,7 +248,7 @@ def _read_dict(self, dictionary):
 
         self._other_values = values
 
-    def add_proof(self, checksums, publisher_account):
+    def add_proof(self, checksums: dict, publisher_account: object) -> None:
```
DMats, line 251: I'm assuming the `publisher_account` is an instance of `eth_account.account.Account`. If I'm right, change `object` to `Account`

ocean_lib/common/ddo/ddo.py
```diff
@@ -31,12 +31,12 @@ class DDO:
```
DMats, line 31: Please add `@enforce_types` to the DDO class

ocean_lib/common/ddo/credentials.py
```diff
@@ -2,28 +2,30 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
+from typing import Optional
+
 from ocean_lib.common.agreements.consumable import ConsumableCodes, MalformedCredential
 
 
 class AddressCredential:
```
DMats, line 10: Please add `@enforce_types`

ocean_lib/common/aquarius/aquarius.py
```diff
@@ -22,7 +20,7 @@
 class Aquarius:
```
DMats, line 20: Please add `@enforce_types`

ocean_lib/common/agreements/service_factory.py
```diff
@@ -13,7 +13,7 @@ class ServiceDescriptor(object):
     item is a dict of parameters and values required by the service"""
```
DMats, line 13: Please add `@enforce_types` to the `ServiceDescriptor` class

ocean_lib/assets/asset.py
```diff
@@ -2,34 +2,36 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
+from typing import Optional
+
 from ocean_lib.common.agreements.service_types import ServiceTypes
 from ocean_lib.common.ddo.ddo import DDO
 
 
 class Asset(DDO):
```
DMats, line 11: Please add `@enforce_types`

ocean_lib/common/ddo/service.py
```diff
@@ -8,6 +8,7 @@
 """
 import copy
 import logging
+from typing import Any, Dict, Optional, Tuple
 
 logger = logging.getLogger(__name__)
```
DMats, line 14: Add `@enforce_types` to the `Service` class

ocean_lib/common/ddo/service.py
```diff
@@ -144,7 +152,7 @@ def _parse_json(cls, service_dict):
         return service_endpoint, _type, _index, _attributes, sd
 
     @classmethod
-    def from_json(cls, service_dict):
+    def from_json(cls, service_dict: Dict[str, Any]) -> "Service":
```
DMats, line 155: Remove the quotes around `Service`

ocean_lib/common/ddo/service.py
```diff
@@ -128,7 +134,9 @@ def as_dictionary(self):
         return values
 
     @classmethod
-    def _parse_json(cls, service_dict):
+    def _parse_json(
+        cls, service_dict: Dict[str, Any]
+    ) -> Tuple[str, str, int, Dict, Dict]:
```
DMats, line 139: After adding `@enforce_types` does specifying `Dict` like this without specifying the contained types actually work?

ocean_lib/common/ddo/status_helper.py
```diff
@@ -5,7 +5,7 @@
 from ocean_lib.common.agreements.service_types import ServiceTypes
 
 
-def enable_flag(ddo, flag_name):
+def enable_flag(ddo: object, flag_name: str) -> None:
```
DMats, line 8: Change `object` to `DDO`

ocean_lib/common/ddo/status_helper.py
```diff
@@ -20,7 +20,7 @@ def enable_flag(ddo, flag_name):
         metadata_service.attributes["status"][flag_name] = True
 
 
-def disable_flag(ddo, flag_name):
+def disable_flag(ddo: object, flag_name: str) -> None:
```
DMats, line 23: Change `object` to `DDO`

ocean_lib/common/ddo/status_helper.py
```diff
@@ -35,7 +35,7 @@ def disable_flag(ddo, flag_name):
         metadata_service.attributes["status"].pop(flag_name)
 
 
-def is_flag_enabled(ddo, flag_name):
+def is_flag_enabled(ddo: object, flag_name: str) -> bool:
```
DMats, line 38: Change `object` to `DDO`

DMats (changes requested):
# Enforce types everywhere
Please add `@enforce_types` to all classes and module functions.

# Consider using more specific type hints
Anywhere `dict` or `list` or `tuple` are used, consider being more specific by using `List` or `Dict` or `Tuple` along with `Optional` and `Union`. 

Examples:
1.  `List[str, str]`
2. `Dict[str, Optional[Any]]`

I think that being specific is better, especially for parameters, and that in theory, any amount of complexity is representable using `Dict`, etc. That said, I recognize this isn't the highest priority task, so I think it's okay to limit it to something as long as the following singly nested dict : `Dict[str, Union[str, Dict[str, Union[str, int]]]`

Also, `@enforce_types` doesn't seem to enforce return types. I think return type hints are useful for documentation and possibly for static analysis tools, but if they're complex, I wouldn't spend a bunch of time on them.

# object and Any
I think we should avoid using `object` in all cases.  I think we should use `Any` only when limiting `Dict`, etc. complexity or when the logic is truly type agnostic. 

# Question
Question: When passing addresses as parameters, should we use the `ChecksumAddress` type like Web3.py does? I suspect this would be preferable, but I don't know if it would break any of our existing logic.

# Progress
I reviewed the following folders:
`assets`
`common/agreements`
`common/aquarius`
`common/ddo`

ocean_lib/common/ddo/status_helper.py
calina-c: that breaks the import logic :( It was originally like that, but it doesn't work.

ocean_lib/common/agreements/service_agreement.py
```diff
@@ -71,7 +72,7 @@ def from_json(cls, service_dict):
         return cls(_attributes, service_endpoint, _type, _index, service_dict)
 
     @classmethod
-    def from_ddo(cls, service_type, ddo):
+    def from_ddo(cls, service_type: str, ddo: object) -> "ServiceAgreement":
```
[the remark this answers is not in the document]
calina-c: Inside ServiceAgreement, ServiceAgreement is not defined, which is why the hint must be done with the class name between quotes.

ocean_lib/common/aquarius/aquarius.py
```diff
@@ -195,7 +201,7 @@ def _parse_search_response(response):
         )
 
 
-def _parse_response(response, default_return):
+def _parse_response(response: object, default_return: Any) -> Any:
```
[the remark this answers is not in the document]
calina-c: The user can pass anything into default_return, so the returned value can be anything, I don't know how it can be tightened. The Response too, can be a mock, dict or Response object...

ocean_lib/common/ddo/service.py
calina-c: See comment from ServiceAgreement

ocean_lib/common/ddo/service.py
calina-c: yes, I added enforce_types everywhere (possible) as of now.

ocean_lib/common/ddo/ddo.py
calina-c: I did now :) https://github.com/oceanprotocol/ocean.py/issues/407

calina-c:
I would like to mention that ocean_pool is not well-tested so I will not be type hinting it now.

ocean_lib/web3_internal/contract_base.py
```diff
@@ -83,7 +84,7 @@ def function_names(self) -> List[str]:
         return list(self.contract.functions)
 
     @staticmethod
-    def to_checksum_address(address: str):
+    def to_checksum_address(address: str) -> str:
```
DMats, line 87: Return value should be `ChecksumAddress`

ocean_lib/web3_internal/contract_base.py
```diff
@@ -157,15 +160,15 @@ def get_event_signature(self, event_name):
...
-        event_filter,
-        callback=None,
-        timeout_callback=None,
-        args=None,
-        wait=False,
-        from_block="latest",
-        to_block="latest",
-    ):
+        timeout: int,
+        event_filter: None,
```
DMats, line 164: Should be `Optional[dict]`

ocean_lib/web3_internal/contract_base.py
```diff
@@ -157,15 +160,15 @@ def get_event_signature(self, event_name):
...
+        callback: None = None,
```
DMats, line 165: Should be `Callable`

ocean_lib/web3_internal/contract_base.py
```diff
@@ -157,15 +160,15 @@ def get_event_signature(self, event_name):
...
+        timeout_callback: None = None,
```
DMats, line 166: `Callable`

ocean_lib/web3_internal/contract_base.py
```diff
@@ -157,15 +160,15 @@ def get_event_signature(self, event_name):
...
+        args: None = None,
```
DMats, line 167: `list`

ocean_lib/web3_internal/account.py
```diff
@@ -6,24 +6,28 @@
...
         self,
-        address=None,
-        password=None,
-        key_file=None,
-        encrypted_key=None,
-        private_key=None,
-    ):
+        address: Optional[str] = None,
+        password: Optional[str] = None,
+        key_file: Optional[PosixPath] = None,
```
DMats, line 27: Should be `Optional[Path]`

ocean_lib/web3_internal/web3_overrides/signature.py
```diff
@@ -19,7 +20,12 @@ class SignatureFix(Signature):
     v value of 27 or 28 instead of 0 or 1
     """
 
-    def __init__(self, signature_bytes=None, vrs=None, backend=None) -> None:
+    def __init__(
+        self,
+        signature_bytes: Optional[bytes] = None,
+        vrs: Optional[Tuple[int, int, int]] = None,
+        backend: Optional[object] = None,
```
DMats, line 27: Should be `Optional[Union[BaseECCBackend, Type[BaseECCBackend], str, None]]`. Copied from `Signature.__init`.

ocean_lib/web3_internal/web3_overrides/request.py
```diff
@@ -8,17 +8,18 @@
 import lru
 import requests
 from requests.adapters import HTTPAdapter
+from requests.sessions import Session
 from web3._utils.caching import generate_cache_key
 
 
-def _remove_session(key, session):
+def _remove_session(key: str, session: Session) -> None:
```
DMats, line 15: Add `@enforce_types`

ocean_lib/web3_internal/web3_overrides/http_provider.py
```diff
@@ -2,14 +2,16 @@
...
+
 from ocean_lib.web3_internal.web3_overrides.request import make_post_request
 from web3 import HTTPProvider
 
 
 class CustomHTTPProvider(HTTPProvider):
     """Override requests to control the connection pool to make it blocking."""
 
-    def make_request(self, method, params):
+    def make_request(self, method: str, params: Any) -> Dict[str, Any]:
```
DMats, line 14: `method` should be `RPCEndpoint`

ocean_lib/web3_internal/web3_overrides/http_provider.py
```diff
@@ -2,14 +2,16 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
+from typing import Any, Dict
+
 from ocean_lib.web3_internal.web3_overrides.request import make_post_request
 from web3 import HTTPProvider
 
 
 class CustomHTTPProvider(HTTPProvider):
```
DMats, line 11: Add `@enforce_types`

ocean_lib/ocean/ocean_compute.py
```diff
@@ -27,13 +29,13 @@ class OceanCompute:
 
     """Ocean assets class."""
 
-    def __init__(self, config, data_provider):
+    def __init__(self, config: Union[Config, Dict], data_provider: object) -> None:
```
DMats, line 32: `data_provider` should be `DataServiceProvider`

ocean_lib/ocean/ocean_compute.py
```diff
@@ -57,8 +61,14 @@ def build_container_attributes(image, tag, entrypoint):
 
     @staticmethod
     def build_server_attributes(
-        server_id, server_type, cpu, gpu, memory, disk, max_run_time
-    ):
+        server_id: str,
+        server_type: str,
+        cpu: Union[str, int],
+        gpu: Union[str, int],
```
DMats, line 67: I think this should just be an `int`

ocean_lib/ocean/ocean_compute.py
```diff
@@ -57,8 +61,14 @@ def build_container_attributes(image, tag, entrypoint):
 
     @staticmethod
     def build_server_attributes(
-        server_id, server_type, cpu, gpu, memory, disk, max_run_time
-    ):
+        server_id: str,
+        server_type: str,
+        cpu: Union[str, int],
```
DMats, line 66: I think this should just be an `int`

ocean_lib/ocean/ocean_compute.py
```diff
@@ -190,7 +204,12 @@ def _status_from_job_info(job_info):
         }
 
     @staticmethod
-    def check_output_dict(output_def, consumer_address, data_provider, config: Config):
+    def check_output_dict(
+        output_def: None,
```
DMats, line 208: `Dict[str, Any]`

ocean_lib/ocean/ocean_compute.py
```diff
@@ -216,7 +235,7 @@ def check_output_dict(output_def, consumer_address, data_provider, config: Confi
         default_output_def.update(output_def)
         return default_output_def
 
-    def create_compute_service_descriptor(self, attributes):
+    def create_compute_service_descriptor(self, attributes: Dict) -> ServiceDescriptor:
```
DMats, line 238: Is `attributes: Dict` valid?  I think `Dict[str, Any]` or `dict` would be better.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -55,7 +56,13 @@ class OceanAssets:
 
     """Ocean assets class."""
 
-    def __init__(self, config, web3, data_provider, ddo_registry_address):
+    def __init__(
+        self,
+        config: Config,
+        web3: Web3,
+        data_provider: object,
```
DMats, line 63: `DataServiceProvider`

ocean_lib/ocean/ocean.py
```diff
@@ -41,7 +42,9 @@ class Ocean:
 
     """The Ocean class is the entry point into Ocean Protocol."""
 
-    def __init__(self, config, data_provider=None):
+    def __init__(
+        self, config: Union[Dict, Config], data_provider: Optional[object] = None
```
DMats, line 46: `data_provider` should be `Optional[DataServiceProvider]`

ocean_lib/ocean/test/test_ocean_compute.py
```diff
@@ -19,8 +18,7 @@
 
 def test_build_cluster_attributes():
     data_provider = DataServiceProvider()
-    config = Ocean.config
-    compute = OceanCompute(config=config, data_provider=data_provider)
+    compute = OceanCompute(config={}, data_provider=data_provider)
```
DMats, line 21: Instead of changing this to `{}`, pass in the `config` fixture like so. Do this same change for `test_build_container_attributes` and `test_build_server_attributes`

```python
def test_build_cluster_attributes(config):
    data_provider = DataServiceProvider()
    compute = OceanCompute(config=config, data_provider=data_provider)
...
```

ocean_lib/models/data_token.py
```diff
@@ -176,11 +177,11 @@ def get_event_signature(self, event_name):
 
     def get_start_order_logs(
         self,
-        consumer_address=None,
-        from_block=0,
-        to_block="latest",
-        from_all_tokens=False,
-    ):
+        consumer_address: Optional[str] = None,
+        from_block: int = 0,
```
DMats, line 181: Whenever `getLogs` is called, `from_block` and `to_block` should be `Optional[BlockIdentifier]` 

This occurs in several places throughout this file.

ocean_lib/models/bpool.py
```diff
@@ -585,12 +586,12 @@ def calcPoolInGivenSingleOut(
...
         self,
-        event_name,
-        from_block,
-        to_block=None,
-        user_address=None,
-        this_pool_only=True,
-    ):
+        event_name: str,
+        from_block: int,
+        to_block: Optional[int] = None,
```
DMats, line 591: `from_block` and `to_block` should be `Optional[BlockIdentifier]`. This happens multiple places in this file.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -46,19 +50,23 @@ class DataServiceProvider:
     provider_info = None
 
     @staticmethod
-    def get_http_client():
+    def get_http_client() -> object:
```
DMats, line 53: Should return `Optional[Session]`

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -46,19 +50,23 @@ class DataServiceProvider:
...
         """Get the http client."""
         return DataServiceProvider._http_client
 
     @staticmethod
-    def set_http_client(http_client):
+    def set_http_client(http_client: object) -> None:
```
DMats, line 58: `http_client` should be `Session`

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -90,15 +98,22 @@ def encrypt_files_dict(
...
             return response.json()["encryptedDocument"]
 
+        return ""
+
     @staticmethod
-    def sign_message(wallet, msg, nonce=None, provider_uri=None):
+    def sign_message(
+        wallet: Wallet,
+        msg: str,
+        nonce: Optional[str] = None,
```
DMats, line 107: Should be `Optional[Union[str, int]]`

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -629,8 +675,11 @@ def check_single_file_info(file_url, provider_uri):
         for file_info in response:
             return file_info["valid"]
 
+        return False
+
     @staticmethod
-    def check_asset_file_info(asset, provider_uri):
+    def check_asset_file_info(asset: object, provider_uri: str) -> bool:
```
DMats, line 681: `asset` should be type `Asset`

ocean_lib/common/utils/utilities.py
```diff
@@ -41,7 +39,7 @@ def convert_to_bytes(data):
     return Web3.toBytes(text=data)
 
 
-def convert_to_string(data):
+def convert_to_string(data: bytes) -> str:
```
DMats, line 42: Return value should be `HexStr`

ocean_lib/common/ddo/status_helper.py
DMats: A circular import is indicative of problems with the design.  I think the easiest solution is to simply move these to `ddo.py` and make them `DDO` methods - thus replacing the `ddo` parameter with `self`.

DMats (changes requested):

ocean_lib/web3_internal/contract_base.py
calina-c: https://web3py.readthedocs.io/en/stable/web3.main.html?highlight=tochecksumaddress#web3.Web3.toChecksumAddress it seems to be a str

ocean_lib/web3_internal/event_filter.py
```diff
@@ -4,22 +4,24 @@
...
     def __init__(
         self,
-        event: ContractEvent,
-        argument_filters=None,
-        from_block=None,
-        to_block=None,
-        address=None,
-        topics=None,
-    ):
+        event: Any,
```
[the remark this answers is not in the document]
calina-c: TypeError: Expected type '<class 'web3.contract.ContractEvent'>' for attribute 'event' but received type '<class 'web3._utils.datatypes.PropertyCheckingFactory'>')

ocean_lib/common/ddo/status_helper.py
calina-c: I would suggest adding these in a separate PR. The ddo file is bloated (partly because of all the `@property` stuff, but not only). Do you agree with adding `Any` for now and creating an issue for moving status_helper into ddo.py?

ocean_lib/web3_internal/web3_overrides/http_provider.py
calina-c: I am having trouble enforcing this. I get issues with the inherited init, and the build errors out.

ocean_lib/ocean/ocean_assets.py
calina-c: This one is compromised by the DataProviderMock. Inside the tests we use DataProviderMock objects, instead of DataServiceProvider. Should we add both as an option? I don't particularly like the usage of a Mock type hint inside the main code.

ocean_lib/common/ddo/status_helper.py
DMats: Yes that sounds fine.

ocean_lib/web3_internal/event_listener.py
```diff
@@ -52,12 +56,12 @@ def make_event_filter(self):
 
     def listen_once(
         self,
-        callback,
-        timeout=None,
-        timeout_callback=None,
-        start_time=None,
-        blocking=False,
-    ):
+        callback: Optional[Callable] = None,
```
DMats, line 59: Did you intentionally make the `callback` optional?  The underlying `watch_one_event` Thread seems to require a non-None `Callable`.

ocean_lib/web3_internal/event_listener.py
```diff
@@ -107,8 +111,13 @@ def _callback(event, *args):
 
     @staticmethod
     def watch_one_event(
-        event_filter, callback, timeout_callback, timeout, args, start_time=None
-    ):
+        event_filter: EventFilter,
+        callback: Callable,
+        timeout_callback: Callable,
```
DMats, line 116: I think the `timeout_callback` should be `Optional[Callable]` based on the logic

```python
                    if timeout_callback is not None:
                        timeout_callback(*args)
```

ocean_lib/web3_internal/contract_base.py
DMats: https://github.com/ethereum/web3.py/blob/master/web3/main.py#L221

ocean_lib/ocean/ocean_assets.py
DMats: I think that `DataProviderMock` should be a subclass of `DataProviderService` instead of `object`.  This would allows the main code to enforce the `DataProviderService` type.

ocean_lib/data_provider/data_service_provider.py
calina-c: HTTP client, along with DataServiceProvider, is one of the types affected by the mocking. Let's discuss these later today, after we settle the other comments.

DMats (changes requested):

ocean_lib/web3_internal/event_filter.py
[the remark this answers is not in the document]
DMats: This indicates an error in the unit test.

Inside `ocean_lib/web3_internal/test/test_event_filter.py::test_transfer_event_filter`, we need to add `()` after `Transfer` on line 20, like so:

```python
    event_filter = EventFilter(token.events.Transfer(), from_block=block, to_block=block)
```

ocean_lib/web3_internal/event_listener.py
calina-c: yes, but the event can be non-blocking. And this was the input detected from the tests. I assume it's ok like this but I will revisit this.

ocean_lib/common/ddo/status_helper.py
calina-c: I ended up moving them... Let's see if we can resolve this after the tests run :)

ocean_lib/web3_internal/contract_base.py
calina-c: I think this can be resolved now :)

ocean_lib/web3_internal/event_filter.py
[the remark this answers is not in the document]
calina-c: Fixed and checked. I think it can be resolved.

ocean_lib/ocean/ocean_compute.py
calina-c: I think this can be resolved too. I fixed the ones related to DataServiceProvider. It is still weird to me that we expect a fully static class there, but this is something we need to figure out later, when removing static stuff. Definitely not now :)

ocean_lib/ocean/ocean_assets.py
calina-c: Fixed.

ocean_lib/models/bpool.py
calina-c: resolvable :)

ocean_lib/web3_internal/event_listener.py
calina-c: can be resolved.

ocean_lib/web3_internal/event_listener.py
calina-c: I don't have a definitive answer on this. I believe it should be kept Optional, but I am open to arguments.

ocean_lib/data_provider/data_service_provider.py
calina-c: This is a cyclical import issue and we can not apply the same principle as for status_helper, since this is clearly a separate class with separate concerns. What do you recomend? `object` or `Any`?

ocean_lib/data_provider/data_service_provider.py
calina-c: Fixed this one too.

ocean_lib/data_provider/data_service_provider.py
calina-c: Fixed.

ocean_lib/web3_internal/web3_overrides/http_provider.py
calina-c: I managed to enforce types for the function. Enforcing for the entire class breaks because of HTTPProvider usages and instances from web3 components. Btw. same for the RPCEndpoint type hinting, I reverted that to str since this is what web3 seems to ask for.

ocean_lib/web3_internal/event_listener.py
DMats: Taking a second look at this, I agree: `Optional` is a more appropriate design. 

But the underlying `watch_one_event` call requires a non-None `Callable`. I now believe that the correct resolution is to add a `None` check before calling `callable()` and change the `callable` parameter to `Optional[Callable]`

ocean_lib/data_provider/data_service_provider.py
DMats: I recommend we modify the logic to take a DID `str` like so:

```python
    @staticmethod
    def check_asset_file_info(did: str, provider_uri: str) -> bool:
        """Check if an asset's file info is valid. Specify asset by its DID."""
        if not did:
            return False
        _, endpoint = DataServiceProvider.build_fileinfo(provider_uri)
        data = {"did": did}
        response = requests.post(endpoint, json=data)

        if response.status_code != 200:
            return False

        response = response.json()
        for ddo_info in response:
            return ddo_info["valid"]

        return False
```

ocean_lib/web3_internal/web3_overrides/http_provider.py
DMats: > I managed to enforce types for the function. Enforcing for the entire class breaks because of HTTPProvider usages and instances from web3 components.

Fair enough.  Thanks for trying.

> Btw. same for the RPCEndpoint type hinting, I reverted that to str since this is what web3 seems to ask for.

I guess that's fine. I like `RPCEndpoint` better because it's more self-documenting. But you're right that it's just a `str` under the hood. https://github.com/ethereum/web3.py/blob/master/web3/types.py#L57

ocean_lib/common/ddo/service.py
calina-c: all functions except __init__ have `@enforce_types` due to the issue with Service and ServiceAgreement inheritance. It spews out the error about arguments (surely it is somewhere in the resolved issues). For __init__ I can not add it, so I can not add it on class-level, only on individual functions.

ocean_lib/web3_internal/event_listener.py
calina-c: I think that check is done implicitly. There is this instruction `original_callback = callback` and then a check `if original_callback:` It also looks like this function is a bit half-baked, with the `#TODO`, so I wouldn't change a lot about it without knowing more. Maybe it warrants a separate issue? Not sure.

DMats: Ah, you're right. Sorry I missed this. `listen_once` defines the nested function `_callback` which is passed into `watch_one_event`, thus ensuring the `callback` parameter is never `None`. I think a separate issue is not warranted.

DMats (approved):

PR #398, opened 2021-07-23
Checked allowance of the token.

### LINKED ISSUE #387: OceanPool.buy_data_tokens works first time, fails subsequent times
`OceanPool.buy_data_tokens` runs `DataToken.approve_tokens`. This works the first time it is run but fails subsequent times if the approval amount hasn't increased.

`OceanPool.buy_data_tokens` should check the current allowance and only request an increase if necessary.

mariacarmina:
Fixes #387 .

Changes proposed in this PR:

- check allowance of the token;
- test the marketplace-flow README.

### DIFF
--- a/ocean_lib/ocean/ocean_pool.py
+++ b/ocean_lib/ocean/ocean_pool.py
@@ -303,7 +303,10 @@ class OceanPool:
         :return: str transaction id/hash
         """
         ocean_tok = DataToken(self.web3, self.ocean_address)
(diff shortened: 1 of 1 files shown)


mariacarmina:
I tested 3 times on ganache network to see if `ocean.pool.buy_data_tokens` works for multiple times and it's fine.
Screenshot from 2021-07-23 14-19-14
I could also download the file
Screenshot from 2021-07-23 14-21-58

DMats (approved):

PR #406, opened 2021-07-27
Remove local node support, update docstrings

### LINKED ISSUE #347: Local Node support is broken
# Background
The `CustomContractFunction` class exists to provide a single interface for sending transactions to Ethereum nodes.  It's supposed to send the transaction to a local node using the `personal` API if the `passphrase` for the `from` account is included in the `transaction` dict, else, it sends the transaction to a hosted node via the `eth` API.

# Problem
Today the local node support is broken because in Web3.py v5, the `personal` API was split to differentiate between Geth and Parity implementations. Concretly, `web3.personal.sendTransaction` was removed in favor of `web3.geth.personal.send_transaction` or `web3.parity.personal.send_transaction`.

# Possible Solution

Fix CustomContractFunction to support both the Geth and Parity implementation by calling `web3._utils.personal.send_transaction`.  Normally, calling internal methods (prefixed with underscore '_') should be avoided but in this case, both the Geth and Parity `send_transaction` methods delegate to `web3._utils.personal.send_transaction`, so using the internal method allows us to not care whether the caller is running a Geth or Parity node.

* Geth: https://github.com/ethereum/web3.py/blob/10e55faccd5250df79608d1fdccd363fef391bea/web3/geth.py#L45
* Parity: https://github.com/ethereum/web3.py/blob/10e55faccd5250df79608d1fdccd363fef391bea/web3/parity.py#L32

DMats:
Fixes #347

Changes proposed in this PR:

- Remove support for sending transactions via a local node
- Update docstrings


calina-c (approved):
LGTM and I know it's been due a long time. Approved with both hands.

PR #414, opened 2021-07-29
Remove list_assets endpoints.

### LINKED ISSUE #410: Remove usage of `aquarius/assets` and `aquarius/assets/ddo` endpoints
Given that the `/assets` and `/assets/ddo` endpoints will be removed from aquarius in https://github.com/oceanprotocol/aquarius/issues/535, we should pre-emptively remove usage of these endpoints from ocean.py.

calina-c:
Closes #410.


ocean_lib/common/aquarius/aquarius.py
```diff
@@ -85,6 +74,17 @@ def get_asset_ddo(self, did):
 
         return DDO(dictionary=parsed_response)
 
+    def ddo_exists(self, did):
```
mariacarmina, line 77: Since this function returns a `bool`, can we provide the return type in the function signature such as `def ddo_exists(self, did) -> bool`?

ocean_lib/test/test_exceptions.py
```diff
@@ -51,16 +51,10 @@ def test_AquariusError(publisher_ocean_instance, metadata):
     metadata_copy = metadata.copy()
```
mariacarmina, line 51: By the way, this function is already used for testing the Aquarius related exception. Would it be redundant if we add an unit test for `ddo_exists`?

ocean_lib/common/aquarius/aquarius.py
calina-c: yes, when we fix the conflicts with the typing branch :))

ocean_lib/test/test_exceptions.py
DMats: I think `ddo_exists` is sufficiently tested by the existing `test_AquariusError` unit test.

DMats (approved):

PR #415, opened 2021-07-29
Add balance and allowance checks before token operations

### LINKED ISSUE #411: Check balance before doing token operations
**Describe the bug**
Before doing any transaction that involves tokens (pool setup, swap, add liquidity, remove, datatoken order, etc), please check the balances and return a nice error message to the user.

Do not let users trying to figure why their transaction failed by debuging smartcontracts/etherscans/etc..

**To Reproduce**
Create a datatoken , approve both DT & Ocean, create BPool and then setup a BPool.
Result:  https://rinkeby.etherscan.io/tx/0xbf9d15a8a46a7aba319519a1bc4267bddca23425135584fa9a80e059500c38dd
image (2)

Reason:  that address has no Ocean tokens  (and never had any)(https://rinkeby.etherscan.io/address/0xaa350d05f253a939ec7fa5a4a8be522372d49b19#tokentxns)

mariacarmina:
Fixes #411  .

Changes proposed in this PR:

- verified the balance and the allowance for Ocean Pool and Ocean Exchange;
- refactored READMEs;
- added the web3 instance as a parameter to compute-flow README.


mariacarmina:
I added some balance & allowance checks for ocean pool. I have tested again with a wallet that has not Ocean tokens to create the Ocean Pool and right now, I receive a descriptive error.
Screenshot from 2021-07-29 17-16-50

mariacarmina:
I have also added some balance checks in READMEs(marketplace-flow and compute-flow as well) and I tested with an empty wallet and here are the results.
For data token creation:
Screenshot from 2021-07-30 16-45-08
For OCEAN pool creation:
Screenshot from 2021-07-30 16-45-59

ocean_lib/ocean/mint_fake_ocean.py
```diff
@@ -46,9 +46,13 @@ def mint_fake_OCEAN(config):
         w = Wallet(web3, private_key=key)
 
         if OCEAN_token.token_balance(w.address) < 1000:
-            OCEAN_token.transfer(
-                w.address, amt_distribute_base, from_wallet=deployer_wallet
-            )
+            if (
+                OCEAN_token.allowance(deployer_wallet.address, w.address)
```
DMats, line 50: Having a sufficient allowance only matters when calling `transferFrom` because the `from_wallet` is different from the token `sender`. When calling `transfer`, the `from_wallet` is the `sender` so they are allowed to send as many of their own tokens as they want. 

Thus, I think this allowance check is unnecessary.

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -82,10 +82,23 @@ def buy_at_fixed_rate(
...
+            )
+            tx_receipt = ocean_token.get_tx_receipt(
+                self._web3,
+                tx_id,
+            )
+            if not tx_receipt or tx_receipt.status != 1:
+                raise VerifyTxFailed(
+                    f"Approve OCEAN tokens failed, exchange address was {self._exchange_address} and tx id was {tx_id}!"
+                )
+        if ocean_token.balanceOf(wallet.address) < amount_base:
```
DMats, line 100: Two things:

1. This logic doesn't make sense.  I suspect that `amount_base` is meant to be `ocean_amount_base`
2. If I'm correct about (1), then I think this logic should be performed before the approval but after `max_OCEAN_amount` check.

Like so:

```python
        if ocean_amount_base > max_OCEAN_amount_base:
            raise ValidationError(
                f"Buying {amount} datatokens requires {from_base_18(ocean_amount_base)} OCEAN "
                f"tokens which exceeds the max_OCEAN_amount {max_OCEAN_amount}."
            )
        ocean_token = DataToken(self._web3, self.ocean_address)
        if ocean_token.balanceOf(wallet.address) < ocean_amount_base:
            raise InsufficientBalance(f"Insufficient funds for buying {amount_base} DataTokens!")
```

P.S. I added `{amount_base}` to the InsufficientBalance error message.

ocean_lib/ocean/ocean_pool.py
```diff
@@ -391,10 +409,22 @@ def add_liquidity_finalized(
         assert self._is_valid_pool(pool_address), "The pool address is not valid."
         dt_address = self.get_token_address(pool_address)
         dt = BToken(self.web3, dt_address)
-        dt.approve(pool_address, max_data_token_amount_base, from_wallet=from_wallet)
+        if dt.balanceOf(from_wallet.address) > max_data_token_amount_base:
```
DMats, line 412: I think this `>` is supposed to be a `<`

ocean_lib/ocean/ocean_pool.py
```diff
@@ -391,10 +409,22 @@ def add_liquidity_finalized(
...
+                f"Insufficient funds for adding liquidity for {dt.address} datatoken!"
+            )
+        if dt.allowance(from_wallet.address, pool_address) < max_data_token_amount_base:
+            dt.approve(
+                pool_address, max_data_token_amount_base, from_wallet=from_wallet
+            )
 
         OCEAN = BToken(self.web3, self.ocean_address)
-        OCEAN.approve(pool_address, max_OCEAN_amount_base, from_wallet=from_wallet)
+        if OCEAN.balanceOf(from_wallet.address) > max_OCEAN_amount_base:
```
DMats, line 422: Same as previous comment, `>` is supposed to be `<`

DMats (changes requested):

ocean_lib/ocean/ocean_pool.py
mariacarmina: Sure, my bad 👍🏻

ocean_lib/ocean/ocean_exchange.py
```diff
@@ -98,8 +98,10 @@ def buy_at_fixed_rate(
                 raise VerifyTxFailed(
                     f"Approve OCEAN tokens failed, exchange address was {self._exchange_address} and tx id was {tx_id}!"
                 )
-        if ocean_token.balanceOf(wallet.address) < amount_base:
-            raise InsufficientBalance("Insufficient funds for buying DataTokens!")
+        if ocean_token.balanceOf(wallet.address) < ocean_amount_base:
```
DMats, line 101: Can you please move this `if` block above the allowance check but after the validation check? I'd hate for someone to send an allowance transaction, only to realize afterward they don't have sufficient OCEAN tokens.

DMats (changes requested):

DMats (approved):

PR #428, opened 2021-08-05
Test that the Provider URL is valid.

### LINKED ISSUE #419: Data_provider:  add isValidProvider(url)
Ocean_lib should have a function isValidProvider (url)

Given a URL (or serviceEndoint), this function should curl the root URL, and detect if that it's a valid provider  ( check https://github.com/oceanprotocol/ocean.js/blob/27d074f3b827bcfbba3eff532a866226a824e2af/src/provider/Provider.ts#L490)

Should return True or False

calina-c:
Closes #419.


calina-c:
Due to the nature of the static DataServiceProvider class, and the fact that every URL build passes through the get_root_uri function, it is my opinion we do not need a separate isValidProvider function in ocean.py, but rather we can check that the provider is valid when building the URLs, directly. @alexcos20 not sure if this is what you had in mind but I think it makes sense. If we need the exact same function to return a bool, please let me know.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -498,6 +498,17 @@ def get_root_uri(service_endpoint: str) -> str:
         if not result:
             raise InvalidURL(f"InvalidURL {service_endpoint}.")
 
+        try:
```
DMats, line 501: I think you should pull this logic out into a separate method like `is_provider_valid(provider_uri)`

DMats (commented):
I'm leaving my review as a comment - neither approving nor requesting changes. I leave it to @alexcos20 to provide final approval.

alexcos20 (approved):

PR #435, opened 2021-08-10
Update the example_config.py

mariacarmina:
Fixes #331  .

Changes proposed in this PR:

- updated `example_config.py` to support ganache, ropsten, rinkeby, mainnet, polygon-mainnet and BSC;
- implemented unit tests  for the new config structure for each network which do not ask about security keys (only `BSC` and `ganache`);
- added more fields in the config object such as `chain_id`, `store_interval` and `network_name`.

PR #436, opened 2021-08-10
Add block confirmations

### LINKED ISSUE #393: Add block confirmations
Linked issue: https://app.zenhub.com/workspaces/backend-5e5e5a97e91ce03f380fb9b7/issues/oceanprotocol/ocean.py/227

mariacarmina:
Fixes #393 .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/ocean_lib/web3_internal/transactions.py
+++ b/ocean_lib/web3_internal/transactions.py
@@ -2,10 +2,13 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
+import time
 from typing import Optional, Union
 
 from enforce_typing import enforce_types
 from eth_account.messages import SignableMessage
+
+from ocean_lib.web3_internal.utils import get_network_timeout
 from ocean_lib.web3_internal.wallet import Wallet
 from web3.datastructures import AttributeDict
 from web3.main import Web3
@@ -43,8 +46,16 @@ def send_ether(
     tx["gas"] = web3.eth.estimate_gas(tx)
     raw_tx = from_wallet.sign_tx(tx)
     tx_hash = web3.eth.send_raw_transaction(raw_tx)
-    receipt = web3.eth.wait_for_transaction_receipt(tx_hash, timeout=30)
-    return receipt
+    receipt = web3.eth.wait_for_transaction_receipt(tx_hash)
+    if tx["chainId"] != 1337:
+        while web3.eth.block_number < receipt.blockNumber + 6:
+            time.sleep(get_network_timeout(tx["chainId"]))
+    else:
+        current_block = receipt.blockNumber
+        while current_block < receipt.blockNumber + 6:
+            time.sleep(get_network_timeout(1337))
+            current_block += 1
+    return web3.eth.get_transaction_receipt(tx_hash)
 
 
 @enforce_types
@@ -65,5 +76,13 @@ def cancel_or_replace_transaction(
     tx["gas"] = gas + 1
     raw_tx = from_wallet.sign_tx(tx, fixed_nonce=nonce_value, gas_price=gas_price)
     tx_hash = web3.eth.send_raw_transaction(raw_tx)
-    receipt = web3.eth.wait_for_transaction_receipt(tx_hash, timeout=30)
-    return receipt
+    receipt = web3.eth.wait_for_transaction_receipt(tx_hash)
+    if tx["chainId"] != 1337:
+        while web3.eth.block_number < receipt.blockNumber + 6:
+            time.sleep(get_network_timeout(tx["chainId"]))
+    else:
+        current_block = receipt.blockNumber
+        while current_block < receipt.blockNumber + 6:
+            time.sleep(get_network_timeout(1337))
+            current_block += 1
+    return web3.eth.get_transaction_receipt(tx_hash)


mariacarmina:
Instead of updating manually the block number for `ganache` network, I send some dummy transactions for generating new blocks creating a new thread that does it until it reaches the wanted block range.

ocean_lib/web3_internal/web3_overrides/contract.py
```diff
@@ -118,8 +123,16 @@ def transact_with_contract_function(
         txn_hash = web3.eth.send_transaction(transact_transaction)
 
     network_id = get_chain_id(web3)
-    web3.eth.wait_for_transaction_receipt(
+    receipt = web3.eth.wait_for_transaction_receipt(
         txn_hash, get_network_timeout(network_id=network_id)
     )
+    if transaction["chainId"] == 1337:
```
calina-c, line 129: I am not very familiar with this functionality but it seems to check out. I like it. However, I would suggest, because there is a lot of code duplication, to isolate this if/else block + the wait_for_transaction_receipt part into a utils function that takes a web3, block number and transaction hash as a parameters. More parameters might be needed, but it would be useful to have a single entry point for this logic. And you'd have to import DEFAULT_BLOCK_CONFIRMATIONS in just one place (single responsibility principle). How does that sound?

calina-c (changes requested):

calina-c:
@mariacarmina I also looked at the unit test times. Can we isolate the BLOCK CONFIRMATIONS logic into a single unit test with monkeypatch instead of applying it to the whole test suite through tox.ini and pytest.ini? I see the unit test time, which is already problematic around 20 mins, has increased to 38 mins. That is not something we can afford and either way we want to test block confirmations separately, not string it along the entire test suite. Can it be done?

ocean_lib/web3_internal/web3_overrides/contract.py
mariacarmina: Sounds great. I'll do it. Thanks for the suggestion

mariacarmina:
> @mariacarmina I also looked at the unit test times. Can we isolate the BLOCK CONFIRMATIONS logic into a single unit test with monkeypatch instead of applying it to the whole test suite through tox.ini and pytest.ini? I see the unit test time, which is already problematic around 20 mins, has increased to 38 mins. That is not something we can afford and either way we want to test block confirmations separately, not string it along the entire test suite. Can it be done?

I am not sure if it can be done separately in a unit test, because the changes are in `contract.py` as well and almost every test depends on it (`transact_with_contract_function`). If the changes were applicable only for `send_ether` and `cancel_or_replace_transaction` from `transactions.py`, just `test_transactions.py` would be affected by the **BLOCK CONFIRMATIONS**.

calina-c:
> > @mariacarmina I also looked at the unit test times. Can we isolate the BLOCK CONFIRMATIONS logic into a single unit test with monkeypatch instead of applying it to the whole test suite through tox.ini and pytest.ini? I see the unit test time, which is already problematic around 20 mins, has increased to 38 mins. That is not something we can afford and either way we want to test block confirmations separately, not string it along the entire test suite. Can it be done?
> 
> I am not sure if it can be done separately in a unit test, because the changes are in `contract.py` as well and almost every test depends on it (`transact_with_contract_function`). If the changes were applicable only for `send_ether` and `cancel_or_replace_transaction` from `transactions.py`, just `test_transactions.py` would be affected by the **BLOCK CONFIRMATIONS**.

Test all normally, with a shorter block confirmation. Just test one transaction with a larger number. Can we do this?

mariacarmina:
> > > @mariacarmina I also looked at the unit test times. Can we isolate the BLOCK CONFIRMATIONS logic into a single unit test with monkeypatch instead of applying it to the whole test suite through tox.ini and pytest.ini? I see the unit test time, which is already problematic around 20 mins, has increased to 38 mins. That is not something we can afford and either way we want to test block confirmations separately, not string it along the entire test suite. Can it be done?
> > 
> > 
> > I am not sure if it can be done separately in a unit test, because the changes are in `contract.py` as well and almost every test depends on it (`transact_with_contract_function`). If the changes were applicable only for `send_ether` and `cancel_or_replace_transaction` from `transactions.py`, just `test_transactions.py` would be affected by the **BLOCK CONFIRMATIONS**.
> 
> Test all normally, with a shorter block confirmation. Just test one transaction with a larger number. Can we do this?

The default block confirmations is 1, I cannot test with lower than this. 

> > > @mariacarmina I also looked at the unit test times. Can we isolate the BLOCK CONFIRMATIONS logic into a single unit test with monkeypatch instead of applying it to the whole test suite through tox.ini and pytest.ini? I see the unit test time, which is already problematic around 20 mins, has increased to 38 mins. That is not something we can afford and either way we want to test block confirmations separately, not string it along the entire test suite. Can it be done?
> > 
> > 
> > I am not sure if it can be done separately in a unit test, because the changes are in `contract.py` as well and almost every test depends on it (`transact_with_contract_function`). If the changes were applicable only for `send_ether` and `cancel_or_replace_transaction` from `transactions.py`, just `test_transactions.py` would be affected by the **BLOCK CONFIRMATIONS**.
> 
> Test all normally, with a shorter block confirmation. Just test one transaction with a larger number. Can we do this?

That means to add a parameter to the functions that I've created only for a unit test that has a different block confirmation number. I've discussed with @DMats about this and for every network you need to wait one block to see if the tx was fully settled. @DMats provided this [link] (https://ethereum.stackexchange.com/questions/100107/ethereum-transaction-confirmations-how-many-are-enough/100119#100119) which explains that 1 block is enough to confirm a transaction.

calina-c:
Update to let everyone know what is happening here: @mariacarmina  is currently working on a high-priority customer reported issue, but she will be back to fix conflicts and finish the test adaptation.

mariacarmina:
The thing that I need to integrate after merging #435 is the config number block confirmations for each network.

calina-c:
Let's discuss this with @DMats when we get the chance. I'd still rather have:
- the default set to 6
- 1 enforced as part of our bulk tests (just to speed up the process)
- one specific test with 6 or any other higher number

However, I see that even with the default configuration set to 1, the tests still amount to 38-40 mins so I'm not sure we can reduce this. Such a large test suite will prove problematic with our process, since it will take 2h at minimum to merge 3 separate PRs.

DMats:
Sure @calina-c I'll be available tomorrow to talk.

ocean_lib/web3_internal/web3_overrides/utils.py
```diff
@@ -0,0 +1,65 @@
...
+from enforce_typing import enforce_types
+from hexbytes import HexBytes
+from web3 import Web3
+
+from ocean_lib.config import Config
+from ocean_lib.web3_internal.wallet import Wallet
+
+
+@enforce_types
+def send_dummy_transactions(block_number: int, from_wallet: Wallet) -> None:
```
DMats, line 17: Please add a docstring. I think it should include a detailed description of what the `block_number` parameter is supposed to be.

ocean_lib/web3_internal/web3_overrides/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        from ocean_lib.example_config import (
+            CONFIG_NETWORK_HELPER,
+            NAME_BLOCK_CONFIRMATION_POLL_INTERVAL,
+        )
+
+        time.sleep(CONFIG_NETWORK_HELPER[1337][NAME_BLOCK_CONFIRMATION_POLL_INTERVAL])
+
+
+@enforce_types
+def fetch_transaction(
```
DMats, line 39: Please add a docstring.

ocean_lib/web3_internal/web3_overrides/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        if timeout
+        else web3.eth.wait_for_transaction_receipt(tx_hash)
+    )
+    from ocean_lib.example_config import (
+        CONFIG_NETWORK_HELPER,
+        NAME_BLOCK_CONFIRMATION_POLL_INTERVAL,
+    )
+
+    if tx["chainId"] == 1337:
+        send_dummy_transactions(receipt.blockNumber, from_wallet)
```
DMats, line 58: I'm not a fan of adding conditions to production code that would only ever be executed during tests. If there is a way for us to make the unit tests send dummy transactions, it would be preferable.

Combined with my other comment, later in this  method, this could eliminate the need to pass in `tx` or `chainId`.

ocean_lib/web3_internal/web3_overrides/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    else:
+        while web3.eth.block_number < receipt.blockNumber + config.block_confirmations:
+            time.sleep(
+                CONFIG_NETWORK_HELPER[tx["chainId"]][
```
DMats, line 62: Instead of passing in the transaction (`tx`), or the `chainId`, it would be better to pass in the poll interval. That would decouple this file from the CONFIG_NETWORK_HELPER struct and make this code more modular.

DMats (changes requested):
@calina-c I think it would be acceptable for the majority of the unit tests to use a `block_confirmations` value of `0`. This could be set in the `config` test fixture. 

Then there could be 1 or 2 unit tests that use other values like `1` or `6`.  Only these unit tests would need to send dummy transactions. 

For this to work, any unit test that specifies a block range would need access to the `config` fixture and define the range based on `block_confirmations` instead of using `DEFAULT_BLOCK_CONFIRMATIONS` or literals like `block - 3`.

ocean_lib/web3_internal/web3_overrides/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        tx = {
+            "from": from_wallet.address,
+            "to": "0xF9f2DB837b3db03Be72252fAeD2f6E0b73E428b9",
+            "value": Web3.toWei(0.001, "ether"),
+            "chainId": web3.eth.chain_id,
+        }
+        tx["gas"] = web3.eth.estimate_gas(tx)
+        raw_tx = from_wallet.sign_tx(tx)
+        web3.eth.send_raw_transaction(raw_tx)
+        from ocean_lib.example_config import (
```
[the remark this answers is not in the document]
mariacarmina: I put it here because of circular imports error.

DMats:
Hello @mariacarmina.  I meant to tinker with this to try and implement my recommendations, but I never got around to it. I'll be on vacation for a week, back on Sept 7th.

DMats:
Here's the design I'm aiming for:

1. `block_confirmations` is a parameter needed by ocean.py's lowest level API (sending transactions)
2. I want `block_confirmations` to be configurable via config.ini or envvar so I'm adding it as an instance variable in the `Config` class because it centralizes all the config.ini and envvar logic.
3. There needs to be a way to pass the block_confirmations param down from the high-level `Config` to the low-level `send_transaction` API.
4. High-Level APIs (`Ocean`, `OceanPool`, `OceanAssets`, `OceanCompute`, `OceanExchange`) are aware of the `Config`, and store it as an instance variable, so whenever they send transactions, they use `self.config.block_confirmations`
5. Mid-Level APIs (`BPool`, `DataToken`, `Dispenser`, etc.) are not aware of the `Config`, nor should they be. I could either (1) store `block_confirmations` as an instance variable when these classes are instantiated.  OR, I could (2) pass in `block_confirmations` as a parameter to each method that sends transactions. I chose to go with option (2) because different transactions may need more or less certainty that the transaction has been finalized on-chain.

ocean_lib/web3_internal/web3_overrides/utils.py
DMats: Marking resolved because function no longer exists.

ocean_lib/web3_internal/web3_overrides/utils.py
DMats: Marking resolved because function no longer exists.

mariacarmina:
> Here's the design I'm aiming for:
> 
> 1. `block_confirmations` is a parameter needed by ocean.py's lowest level API (sending transactions)
> 2. I want `block_confirmations` to be configurable via config.ini or envvar so I'm adding it as an instance variable in the `Config` class because it centralizes all the config.ini and envvar logic.
> 3. There needs to be a way to pass the block_confirmations param down from the high-level `Config` to the low-level `send_transaction` API.
> 4. High-Level APIs (`Ocean`, `OceanPool`, `OceanAssets`, `OceanCompute`, `OceanExchange`) are aware of the `Config`, and store it as an instance variable, so whenever they send transactions, they use `self.config.block_confirmations`
> 5. Mid-Level APIs (`BFactory`, `BPool`, `BToken`, `DataToken`, `Dispenser`, `DTFactory`, `FixedRateExchange`, `Metadata`, `ContractBase`) are not aware of the `Config`, nor should they be. I could either (1) pass  `block_confirmations` to the `__init__` method and store it as an instance variable when these classes are instantiated.  OR, I could (2) pass in `block_confirmations` as a parameter to each method that sends transactions. I chose to go with option (2) because different transactions may need more or less certainty that the transaction has been finalized on-chain.

I don't know what to think about point 1. Some of the functions exceeded the parameters number limit. 
For point 2, I think I managed to add `block_confirmations` just as an envvar for that moment when you requested the changes, so that aspect is fine by me. Overall, I didn't know at the beginning that it was needed to change almost all the files for this, but if this is the right approach and the rest agree on this, it's fine, but this workaround may require API breaking changes.

mariacarmina:
> > Here's the design I'm aiming for:
> > 
> > 1. `block_confirmations` is a parameter needed by ocean.py's lowest level API (sending transactions)
> > 2. I want `block_confirmations` to be configurable via config.ini or envvar so I'm adding it as an instance variable in the `Config` class because it centralizes all the config.ini and envvar logic.
> > 3. There needs to be a way to pass the block_confirmations param down from the high-level `Config` to the low-level `send_transaction` API.
> > 4. High-Level APIs (`Ocean`, `OceanPool`, `OceanAssets`, `OceanCompute`, `OceanExchange`) are aware of the `Config`, and store it as an instance variable, so whenever they send transactions, they use `self.config.block_confirmations`
> > 5. Mid-Level APIs (`BFactory`, `BPool`, `BToken`, `DataToken`, `Dispenser`, `DTFactory`, `FixedRateExchange`, `Metadata`, `ContractBase`) are not aware of the `Config`, nor should they be. I could either (1) pass  `block_confirmations` to the `__init__` method and store it as an instance variable when these classes are instantiated.  OR, I could (2) pass in `block_confirmations` as a parameter to each method that sends transactions. I chose to go with option (2) because different transactions may need more or less certainty that the transaction has been finalized on-chain.
> 
> I don't know what to think about point 1. Some of the functions exceeded the parameters number limit.
> For point 2, I think I managed to add `block_confirmations` just as an envvar for that moment when you requested the changes, so that aspect is fine by me. Overall, I didn't know at the beginning that it was needed to change almost all the files for this, but if this is the right approach and the rest agree on this, it's fine, but this workaround may require API breaking changes.

trentmc:
I hadn't followed this PR for a while. It felt like the change could be surgical, I was wondering why it was taking so many weeks to complete.

And I found the answer: It seems that every solidity wrapper has got an additional argument added `block_confirmations: int`. There are 33 files changed. For example models/bfactory.py now has:
```python
def newBPool(self, from_wallet: Wallet, block_confirmations: int) -> str:
```

This is not needed, nor wanted. (It also explains why it's taken so long.)
- **Not needed.** This is a parameter stored in config vars / envvars, as the github issue specifies, and David laid out nicely [in an earlier comment](https://github.com/oceanprotocol/ocean.py/pull/436#issuecomment-920945966). Given that it's stored there, there is no need to tack it onto every method. Rather, just grab it in the main method that does the blocking and tx call: eb3_internal/web3_overrides/utils.py::wait_for_transaction_receipt_and_block_confirmations(). Or maybe a level above if really needed. [Here's more context on blocking](https://github.com/oceanprotocol/ocean.py/issues/227#issuecomment-920997522).
- **Not wanted.** Tacking this arg to all these methods makes the methods' interfaces messier, opposite of simple.

Summary: `block_confirmations` should not be a parameter in the various model/*.py files. Take it out, do it surgically at the main method `wait_for_transaction_receipt_and_block_confirmations()`.

DMats:
Thank you @mariacarmina and @trentmc for your comments.

Based on feedback from the core team during today's daily stand-up, we concluded that one of my previous assumptions was false: Different transactions are **not** more important than others.

In light of this as well as @trentmc 's [comment](https://github.com/oceanprotocol/ocean.py/pull/436#issuecomment-921002444) above, I've decided to change the design for Mid-level APIs. Instead of option (2), I'll select option (1) pass `block_confirmations` to the `__init__` method and store it as an instance variable when these classes are instantiated. This will 

>5. Mid-Level APIs (`BFactory`, `BPool`, `BToken`, `DataToken`, `Dispenser`, `DTFactory`, `FixedRateExchange`, `Metadata`, `ContractBase`) are not aware of the `Config`, nor should they be. I could either (1) pass  `block_confirmations` to the `__init__` method and store it as an instance variable when these classes are instantiated. OR, I could (2) pass in `block_confirmations` as a parameter to each method that sends transactions. ~~I chose to go with option (2) because different transactions may need more or less certainty that the transaction has been finalized on-chain.~~

trentmc:
> pass block_confirmations to the __init__ method and store it as an instance variable when these classes are instantiated

To the __init__ method of each models/class? Eg a btoken class?

If yes, why? Why would each model object have to store the same config value? They should never have to deal with this. It only needs to be in one very surgical place, inside wait_for_transaction_receipt_and_block_confirmations().

DMats:
>If yes, why? Why would each model object have to store the same config value? They should never have to deal with this setting at all; not storing it; not sending it; nothing. block_confirmations only needs to be used in one very surgical place, inside wait_for_transaction_receipt_and_block_confirmations(). Have some config to have it set, apply it in that one surgical place, leave everything else the same. It should be like 3 files changed, not 33.

@trentmc "surgically" injecting the `block_confirmations` value at this low level without any affect on the higher level APIs would require global access to the block_confirmations option.  

If we read the envvar directly inside `wait_for_transaction_receipt_and_block_confirmations`, this would ignore users who adjust `block_confirmations` via `config.ini` or via a config dict. 

If we made the `Config` globally accessible (singleton) this would break users who want to connect to multiple chains at a time using separate instances of Ocean.

The solution I've proposed is built on the assumption that block_confirmations should be configurable by the user.
If we remove this assumption then, we can just hardcode the number of block_confirmations based on the chain_id. But I disagree with this approach.  I think some users will want to control the number of block confirmations because it affects user experience.

trentmc:
> f we read the envvar directly inside wait_for_transaction_receipt_and_block_confirmations, this would ignore users who adjust block_confirmations via config.ini or via a config dict.

To be clear, this is not what I proposed. We want to grab the config info from wherever it came from.

> If we made the Config globally accessible (singleton) this would break users who want to connect to multiple chains at a time using separate instances of Ocean.

Agree. And it's going exactly backwards on work done in previous months, work done for good reason (!) :)

> The solution I've proposed is built on the assumption that block_confirmations should be configurable by the user.
> If we remove this assumption then, we can just hardcode the number of block_confirmations based on the chain_id. But I disagree with this approach. I think some users will want to control the number of block confirmations because it affects user experience.

I think this is a reasonable goal. 

Then the question to ask is: is there a place where we can get the `block_confirmations` info (configurable by user) into `wait_for_transaction_receipt_and_block_confirmations()` without having to touch all the stuff in `models/`? 

There are likely a few answers. Here are a couple.
1. `Wallet` class gets a new attribute `block_confirmations`. Value is set from the value in `Config` object when a `Wallet` object is instantiated. Then `web3_internal/contract_base.py::send_transaction()` does _not_ take the `block_confirmations` argument, rather, inside `send_transaction()` it grabs the `block_confirmations` value from the `from_wallet` object (of class `Wallet`).
2. Or, like above except `Wallet` class gets a new attribute `config` which references the `config` object.
3. 

I'm on the fence whether (1) or (2) is best. (1) holds less but being an int it's a copy of the value, so if config changes then it won't. (2) is probably giving too much info to Wallet. Is there perhaps a way for `Wallet` class to be able to point to the `block_confirmations` value in `config`? If yes then that's ideal, call it option (3).

DMats:
> 1. Wallet class gets a new attribute block_confirmations, an int. Value is set from the value in Config object when a Wallet object is instantiated. Then web3_internal/contract_base.py::send_transaction() does not take the block_confirmations argument, rather, inside send_transaction() it grabs the block_confirmations value from the from_wallet object (of class Wallet).

> 3. Like (1) but don't store a copy of the block_confirmations value, rather reference it somehow.

> (3) combines the best of the other two; my q is how to do it, though it feels like there is a nice pythonic way to do it.

I like this idea.

Ideas for achieving (3):
A. Pass `block_confirmations` to `Wallet` inside a `list` or `dict` and store it as an attribute.
B. Pass `block_confirmations` to `Wallet` inside a custom `BlockConfirmations` object and store it as an attribute. Anytime Wallet sends a transaction, it reads `self.block_confirmations.value`

I prefer option (B) because I think it expresses intent more explicitly.  I'll start implementing and see if I can get it working.

trentmc:
OK to 3b above :)

DMats:
Huzzah! The existing tests have passed!  I think this PR could be ready for review.  
Part of me wants to add more unit tests, but I'm going to take a break for now.

READMEs/datatokens-flow.md
```diff
@@ -76,7 +76,7 @@ config = Config('config.ini')
 ocean = Ocean(config)
 
 print("create wallet: begin")
-wallet = Wallet(ocean.web3, private_key=private_key)
+wallet = Wallet(ocean.web3, private_key=private_key, block_confirmations=config.block_confirmations)
```
trentmc, line 79: I'm wondering if we can keep this shorter for the READMEs. My suggestion: don't specify the argument names, and just have 1 confirmation. That is:
'''python 
wallet = Wallet(ocean.web3, private_key, 1)
'''

Do this for all the READMEs, for the sake of readability.

READMEs/fixed-rate-exchange-flow.md
```diff
@@ -99,7 +99,7 @@ print(f"config.provider_url = '{config.provider_url}'")
 #Alice's wallet
 import os
 from ocean_lib.web3_internal.wallet import Wallet
-alice_wallet = Wallet(ocean.web3, private_key=os.getenv('TEST_PRIVATE_KEY1'))
+alice_wallet = Wallet(ocean.web3, private_key=os.getenv('TEST_PRIVATE_KEY1'), block_confirmations=config.block_confirmations)
```
trentmc, line 102: Make a separate line for:

private_key = os.getenv(..)

Then make Wallet(..) briefer like the above readme.

ocean_lib/web3_internal/wallet.py
```diff
@@ -32,18 +33,21 @@ class Wallet:
...
 
     """
 
     _last_tx_count = dict()
 
     @enforce_types
-    def __init__(self, web3: Web3, private_key: str) -> None:
+    def __init__(
+        self, web3: Web3, private_key: str, block_confirmations: Integer
+    ) -> None:
```
trentmc, line 45: Make this such that it can handle just a plain int, or the new Integer object.

trentmc: But if plain int comes in, convert to Integer objects before storing.

ocean_lib/web3_internal/web3_overrides/test/test_contract.py
```diff
@@ -9,16 +9,16 @@
 from ocean_lib.web3_internal.web3_overrides.contract import CustomContractFunction
 
 
-def test_main(web3, dtfactory_address):
+def test_main(web3, config, dtfactory_address):
```
trentmc, line 12: Instead of taking in config which is big, just take on block_confirmations

DMats: I think passing the config to unit tests is okay. Many unit tests do this. In this case `config` refers to the `config()` pytest fixture.

Passing only block_confirmations would require the creation of a new `block_confirmations()` pytest fixture. Please let me know if I should proceed with this change anyway.

ocean_lib/web3_internal/wallet.py
DMats: I love this idea. It'll be in my next push.

READMEs/datatokens-flow.md
DMats: I agree with shortening, but I disagree with the `1` literal.  Once the argument names are removed, the line is still relatively short when using `config.block_confirmations`. This is a more correct convention to show users because it accommodates any chain, not just the chains where 1 block confirmation is sufficient.

READMEs/fixed-rate-exchange-flow.md
DMats: Done.

ocean_lib/web3_internal/wallet.py
trentmc: +1

READMEs/fixed-rate-exchange-flow.md
```diff
@@ -99,15 +99,16 @@ print(f"config.provider_url = '{config.provider_url}'")
...
+alice_wallet = Wallet(ocean.web3, alice_private_key, config.block_confirmations)
 print(f"alice_wallet.address = '{alice_wallet.address}'")
 
 #Mint OCEAN for ganache only
 from ocean_lib.ocean.mint_fake_ocean import mint_fake_OCEAN
 mint_fake_OCEAN(config)
 
 assert alice_wallet.web3.eth.get_balance(alice_wallet.address) > 0, "need ETH"
-data_token = ocean.create_data_token('DataToken1', 'DT1', alice_wallet, blob=ocean.config.metadata_cache_uri)
+data_token = ocean.create_data_token('DataToken1', 'DT1', alice_wallet, config.metadata_cache_uri)
```
trentmc, line 111: I'd keep in the keyword argument `blob`, because "blob" is much more general than "metadata_cache_uri", unlike most other places we removed keyword arguments where it was a 1:1.

DMats: Done.

trentmc:
@DMats are you ready for a review? Or are there more changes to go?

DMats:
@trentmc Sorry, as I was reading through just now I realized the docstring was incomplete and I decided to add another small test case.  I have no further changes.

DMats:
It looks like the ranges I had previously selected for querying Events was not wide enough, especially when the `block_confirmations` param is set to 0. So I've widened by setting `from_block = block - (block_confirmations + 1)`

trentmc (approved):

DMats (approved):

PR #437, opened 2021-08-11
Publisher trusted algorithm publishers

### LINKED ISSUE #434: publisherTrustedAlgorithmPublishers
In order to use an algorithm with a dataset on C2D, we have 3 options now:
  - allowRaw
  - allowAnyPublished
  - publisherTrustedAlgorithms

When having tons of assets, it's hard to update them every time a trusted publisher is publishing a new algorithm. Or in Enterprise world, it's secure to trust your IT Department.

Proposal:
 - add publisherTrustedAlgorithmPublishers, which is an array of algorithm publishers (addresses) that are trusted by a specific dataset publisher.

When starting a compute job, we can check if the algorithm is published by a trusted provider, and if so, grant access

calina-c:
Closes #434


mariacarmina (approved):
LGTM, with a small tweak regarding code climate, maybe we can fix it.

alexcos20 (approved):
lgtm, but be sure that we are marking this as a breaking change (some functions args are changed)

PR #438, opened 2021-08-11
Wei interface

### LINKED ISSUE #231: Base 18 conversions

DMats:
Fixes #231

Changes proposed in this PR:

- 
- 
-


DMats:
I won't be able to continue work on this branch until I return from vacation on Friday, August 20th.

DMats:
Status:
`OceanPool` still uses `float`. I have yet to evaluate it but I suspect most of the times it uses `float`, it's encoding ether values, that could instead be expressed in wei and encoded as `int`.

Several other classes still use `float` but it's okay to not change them:

1. service_agreement `get_cost()` returns a `float` which it pulls from the blockchain.
2. ocean_assets `build_access_service()` assigns a `float` to a dictionary that eventually gets stored on-chain. 
3. EventListener uses `float` to represent timeout in seconds.

DMats:
A quick comment in case someone else picks this up while I'm on vacation.

My thought process for the remainder of this task was as follows:
1. Find all uses of `float`
2. Assess each occurrence to see if it's expressing token amounts in ether.  If so, modify it to express the same amount in wei, encoded as an `int`.
3. Consider removing the `_base` postfix anywhere it is used.

DMats:
@calina-c Per your request at today's sprint grooming meeting, I've tested this PR against all the READMEs. All README's completed without errors.

I discovered that `developers.md` section 8 references `remark` and `sphinx` for generating markdown docs. This conflicts with the `release-process.md` which references `oceanprotocol/readthedocs` to generate docs. I believe that `release-process.md` is correct and that section 8 of `developers.md` should be removed. I created issue #462 to address this.

I believe this PR is now ready for review.

PR #470, opened 2021-09-02
Compute result

### LINKED ISSUE #457: Secure C2D Outputs
Right now, the result of a C2D data (outputs/algo logs) are sent as a URL to the consumer.
Which means that anyone who has those URL has access to them, even if he/she is not the actual consumer.

Proposal:

- GET /compute    (get job results) hides the urls, exposing only an array of output file names, size & type
```json
[
  {
    "agreementId": "0x111111",
    "algoDID": "did:op:87bdaabb33354d2eb014af5091c604fb4b0f67dc6cca4d18a96547bffdc27bcf",
    "dateCreated": 1629281470.08931,
    "dateFinished": 1629281491.05138,
    "inputDID": [
      "did:op:1384941e6f0b46299b6e515723df3d8e8e5d1fb175554467a1cb7bc613f5c72e"
    ],
    "jobId": "6951125396154f2da719f097ae758bb5",
    "owner": "0xC41808BBef371AD5CFc76466dDF9dEe228d2BdAA",
    "removed": 0,
    "results": [
      {
        "filename": "output.log",
        "filesize": 85,
        "type": "output"
      },
      {
        "filename": "algorithm.log",
        "filesize": 127,
        "type": "algorithmLog"
      }
    ],
    "status": 70,
    "statusText": "Job finished",
    "stopreq": 0
  }
]
```

Provider exposes new endpoint:   GET /computeResult, which retrieves an output from a job (index based)
```
parameters:
      - name: consumerAddress
        in: query
        description: The consumer address.
        required: true
        type: string
      - name: jobId
        in: query
        description: JobId
        required: true
        type: string
      - name: index
        in: query
        description: Result index
        required: true
      - name: signature
        in: query
        description: Signature of (jobId+index+consumerAddress+nonce) to verify that the consumer has rights to download the result
    """
(see https://github.com/oceanprotocol/provider/pull/201)


Ocean.py should use that endpoint

### PR DESCRIPTION
**Pull request by calina-c**
Closes #457.

### CODE REVIEW

File: ocean_lib/data_provider/data_service_provider.py [hunk 1]
```diff
@@ -446,6 +446,46 @@ def compute_job_result(
             "get", did, job_id, service_endpoint, consumer_address, signature
         )
 
+    @staticmethod
+    @enforce_types
+    def compute_job_result_safe(
```
calina-c, line 451: compute_job_result_file

calina-c:
Updated for transparency: running this with a local provider from main branch, everything works perfectly. However, it does not work with the provider from barge. I have no explanation for this. I asked for @alexcos20 to help me out a bit.

calina-c:
Status: works with the latest provider version on barge, but waiting for Provider release of 0.4.16, which in turn depends on ocean-contracts 0.6.6. Will update and re-request review after dependencies are solved.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -446,6 +446,46 @@ def compute_job_result(
...
+        job_id: str,
+        index: int,
+        service_endpoint: str,
+        consumer_address: str,
+        signature: str,
+    ) -> Dict[str, Any]:
+        """
+
+        :param job_id: str id of compute job that was returned from `start_compute_job`
+        :param index: compute result index
```
DMats, line 461: Missing type.

```python
:param index: int compute result index
```

tests/integration/test_compute_flow.py
```diff
@@ -208,6 +209,21 @@ def run_compute_test(
         print(f"got job status after requesting result: {result}")
         assert "did" in result, "something not right about the compute job, no did."
 
+        succeeded = False
+        for _ in range(0, 200):
+            status = ocean_instance.compute.status(did, job_id, consumer_wallet)
+            if status["status"] > 60:
```
DMats, line 215: What is the significance of the number 60?

tests/resources/mocks/data_provider_mock.py
```diff
@@ -57,6 +57,10 @@ def compute_job_status(*args, **kwargs):
     def compute_job_result(*args, **kwargs):
         return True
 
+    @staticmethod
+    def compute_job_result_file(*args, **kwargs):
+        return True
```
DMats, line 62: Since the return value of `DataServiceProvider.compute_job_result_file()` is `Dict[str, Any]` shouldn't this `Mock` return something of the same type?

DMats (changes requested):

tests/resources/mocks/data_provider_mock.py
calina-c: I think all the methods are just dummies in place. We should change all of them, if that is the case. Maybe we can create a ticket for this? I think it's not strictly needed, but let's discuss this in its own.

DMats: Fair enough.

DMats (approved):

PR #475, opened 2021-09-06
Use c2d address in pay for service.

### LINKED ISSUE #469: Compute: dataset on provider 1 , algorithm on provider2
**Describe the bug**
Ocean.py MUST be able to start a compute job when dataset and algorithm are published using different providers.
This is not happening now, because the ordering flow is completely broken (consumer always equals payer)

More details:
Any provider will expose the 'computeAddress'  - see here:  https://provider.rinkeby.oceanprotocol.com/
(Ocean.py has no way of fetching this address)

When ordering assets for a C2D job(https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/ocean/ocean_assets.py#L518 , the consumer argument should be the compute address of the provider and not left blank 

Reason:   
For a C2d job you order the asset (data or algo), but actually you ARE NOT the consumer.  
But C2D is  (because C2D will download the file to make it available to job)
that's why the consumer MUST always be computeAddress, and never the user
otherwise, you have an order as consumer, which is useless, because C2D cannot download the dataset  (C2D doesn't have the consumer private key)

so :
1.for simple downloads, consumer can be = to payer  (although you can have the scenario:  I will pay for you)
2. for compute jobs, always consumer ! = payer

References:  https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts#L1233-L1289

Bonus: 
Make sure that ocean.py covers all flows from https://github.com/oceanprotocol/ocean.js/blob/main/test/integration/ComputeFlow.test.ts  (update permisions, working with multiple providers, etc) if they are not covered

(Label it Enhancement because we need new features, but BUG because we should already have them)

calina-c:
Closes #469.


READMEs/marketplace-flow.md
```diff
@@ -276,7 +276,14 @@ service = asset.get_service(ServiceTypes.ASSET_ACCESS)
...
 order_tx_id = ocean.assets.pay_for_service(
     ocean.web3,
-    quote.amount, quote.data_token_address, asset.did, service.index, fee_receiver, bob_wallet, None)
+    quote.amount,
+    quote.data_token_address,
+    asset.did,
+    service.index,
+    fee_receiver,
+    bob_wallet,
+    service.get_provider_address()
```
alexcos20, line 285: this is wrong. It should be the C2d Address, and not the provider address

ocean_lib/models/test/test_order.py
```diff
@@ -67,5 +68,7 @@ def test_order(web3, alice_ocean, alice_wallet):
             order[0] == asset.data_token_address
         ), "The order data token address is different."
         assert order[5] == alice_wallet.address, "The payer is not the supposed one."
-        assert order[6] == alice_wallet.address, "The consumer is not the supposed one."
+        assert (
+            order[6] == sa.get_provider_address()
```
alexcos20, line 72: this is wrong. It should be the C2d Address, and not the provider address

tests/integration/test_compute_flow.py
```diff
@@ -102,7 +102,7 @@ def process_order(ocean_instance, publisher_wallet, consumer_wallet, ddo, servic
 
     # Start the order on-chain using the `order` requirements from previous step
     service = ddo.get_service(service_type)
-    consumer = consumer_wallet.address
+    consumer = service.get_provider_address()
```
alexcos20, line 105: this is wrong. It should be the C2d Address, and not the provider address

alexcos20 (changes requested):

alexcos20:
Please add a test that starts a compute with dataset A publisher on provider 1, algo B published on provider 2

calina-c:
We already have this test as part of test_compute_multi_inputs.

And this morning, barge won't fund my accounts either, so I can't run any ocean.py tests (not even on main). I assume it's because I tried to run v4 last week and somehow messed up my addresses/artifacts. I'll continue to try debugging this.

calina-c:
Screenshot 2021-07-26 at 14 06 35

Barge has this error starting up, by the way...

calina-c:
Screenshot 2021-09-14 at 08 17 01

Barge had this startup error btw.

calina-c:
It seems the barge error was local and disappeared after pruning.

alexcos20 (approved):

PR #480, opened 2021-09-09
Optimise tests

### LINKED ISSUE #473: Reduce ocean.py test suite runs, using mocks and patches where possible
Right now, ocean.py tests take around 30 mins. This is a long time interval, especially since we are required to run this before every PR merge. Also, most of the ocean.py tests are in fact integration tests. I think a big chunk of time can be saved if we use proper unit tests in some places.

My suggestion would be to use mocks and patches. Python offers a flexible way of using patches and replacing any functionality with a mock. Previously in ocean.py, we overcomplicated the usage of mocks, IF we used them at all. You can look into the Aquarius test suite for inspiration and I am thinking of creating a presentation for the next sprint review. I can attach that too, when it's done.

calina-c:
Closes #473.

Prepared this Google Spreadsheet to see the results better: https://docs.google.com/spreadsheets/d/1nwQBMMsxMkZ8ekfML4V0FGNCAuxENNvfKfDL90vKT6g/edit?usp=sharing

I am only considering improvement on tests > 1 second for now.


calina-c:
Types of changes:

## Patch asset resolving (noisiest)
In most places, the "real" Aquarius (barge) is used to publish and wait for a ddo to be indexed. Where applicable, I patched the `resolve` or `resolve_asset` functions to return a sample ddo instead. This eliminates the need to wait for the publishing, indexing etc. and removes some of the extra load on Aquarius. In these cases, we were using published ddos, but the publishing itself was irrelevant to the tests.

I kept the functionality only where it was relevant (actually testing the publish)

What you will see:
- removal of `wait_for_ddo` and `get_registered_ddo_*` functions
- replacing the ddo in question with `get_sample_ddo_*`
- setting up the mock: `with patch("ocean_lib.assets.utils.resolve_asset") as mock:` and setting the return value e.g. `mock.return_value = algorithm_ddo_v2`
- the rest of the test is virtually unchanged

## Other changes
- if the "trustedAlgorithmPublishers" or "trustedAlgorithms" keys do not exist, consider them empty (possible bug)
- adapt create_publisher_trusted_algorithms to accept dids or ddos, just like the other functions of the module
- where a TimeExhausted function was expected, patch the original function to immediately throw it, instead of waiting for it "naturally"
- in `test_order`, remove the args list, since it was only useful back when we parameterised the test and we adjusted the args list
- previously `get_sample_algorithm_ddo()` returned a dict, so I renamed it to `get_sample_algorithm_ddo_dict`. I also created a `get_sample_algorithm_ddo` function to return an Asset. This is now consistent with the other function names.
- removed some unused fixtures
- created common fixtures in test_compute, to leverage the caching mechanisms for fixtures, instead of publishing the assets in each test function

ocean_lib/assets/test/test_utils.py
```diff
@@ -115,12 +108,13 @@ def test_fail_generate_trusted_algo_dict():
...
-    publisher = get_publisher_wallet()
-    ddo = get_registered_ddo_with_compute_service(
-        publisher_ocean_instance, publisher, trusted_algorithm_publishers=["0xabc"]
-    )
-    wait_for_ddo(publisher_ocean_instance, ddo.did)
-    assert ddo is not None, "DDO is not found in cache."
+    ddo = get_sample_ddo_with_compute_service()
+    compute_service = ddo.get_service(ServiceTypes.CLOUD_COMPUTE)
+    privacy_values = compute_service.attributes["main"].get("privacy")
+    if not privacy_values:
```
DMats, line 114: This `if` statement should be unnecessary.  Please remove it and update `ddo_with_compute_service.json` to contain a well-formed compute service with the `publisherTrustedAlgorithmPublishers` key defined.

DMats (approved):
Approving under the assumption that comments are addressed.

PR #482, opened 2021-09-15
V0.5.31

calina-c:


calina-c:
Proposed release notes:

Breaking changes:
* remove unnecessary properties across ocean.py models, unused and obsolete functions
* remove possibility of creating a wallet with encrypted key and password
* replace NETWORK_URL env var with OCEAN_NETWORK_URL
* add a wei interface and replace all amounts to wei. The ocean.py wei interface now supports custom number of decimals and helps with amount conversion.
* more strict variable type enforcement
* changed ganache keys in readmes, pytest.ini and tox.ini

Non-breaking changes:
* added supplementary check for provider URLs. Make sure there’s a Provider-like interface at the other end, not only that the link works.
* README improvements
* accomodate publisherTrustedAlgorithmPublishers in DataServiceProvider and OceanCompute, allowing algorithms to be trusted based on their publisher, not only on their individual DIDs
* add support for Moonbeam Alpha network
* add asset encryption
* implemented a function that searches the exchange by data token address
* replace underlying usage of get_event_logs with getLogs
* remove obsolete Sphinx documentation
* optimise tests

PR #496, opened 2021-09-21
Fix `OceanPool._get_all_liquidity_records method()`

### LINKED ISSUE #495: OceanPool._get_all_liquidity_records() fails due to incorrect argument lists
The `OceanPool._get_all_liquidity_records()` method fails for a couple reasons:  

1. The `get_dtfactory_address()` call is missing the `address_file` argument.
2. The `get_join/exit/swap_logs` call has extra `web3` argument that is unnecessary.

DMats:
Closes #495

Breaking changes proposed in this PR:

- Add `OceanPool.dtfactory_address` attribute

Changes proposed in this PR:

- Fix `OceanPool._get_all_liquidity_records method()`
- Add `OceanPool.get_liquidity_history()` to integration test. This will hopefully catch argument list errors in the future.

### DIFF
--- a/ocean_lib/ocean/ocean_pool.py
+++ b/ocean_lib/ocean/ocean_pool.py
@@ -13,7 +13,6 @@ from ocean_lib.models.bpool import BPool
 from ocean_lib.models.btoken import BToken
 from ocean_lib.models.data_token import DataToken
 from ocean_lib.models.dtfactory import DTFactory
-from ocean_lib.ocean.util import get_dtfactory_address
 from ocean_lib.web3_internal.currency import from_wei, to_wei
 from ocean_lib.web3_internal.wallet import Wallet
 from scipy.interpolate import interp1d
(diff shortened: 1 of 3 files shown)


ocean_lib/ocean/ocean_pool.py
```diff
@@ -655,22 +659,27 @@ def _get_all_liquidity_records(
...
-        action_to_fn = {
-            "join": "get_join_logs",
-            "exit": "get_exit_logs",
-            "swap": "get_swap_logs",
-        }
         current_block = to_block if to_block is not None else self.web3.eth.block_number
         pool = BPool(self.web3, pool_address)
         dt_address = token_address or self.get_token_address(pool_address, pool)
-        factory = DTFactory(self.web3, get_dtfactory_address(web3=self.web3))
+        factory = DTFactory(self.web3, self.dtfactory_address)
```
trentmc, line 665: Summary: I'm ok with this change. 

Details: When I first saw the chnage, I thought: why not just keep `get_dtfactory_address()` approach and add in the missing argument `address_file`? Since it avoids redundant storage of information. But then digging around, it's clear that it's gonna add complexities to get the `address_file` value. And there are questions of "which chain" too. By the time you've solved it, you've got something much more complex.

Whereas if you simply store the `dtfactory_address` attribute then everything stays clean. And it follows a pattern identical to the `bfactory_address` attribute.

It's the latter that's implemented here. Bravo!

trentmc (approved):

ocean_lib/ocean/ocean_pool.py
DMats: My thoughts exactly! :)

Issue #497, opened 2021-09-23
Release next ocean.py version


DMats:

DMats:
@calina-c Step 0 of the release-process.md instructs the dev to update the docs via the instructions at https://github.com/oceanprotocol/readthedocs. Then Step 1 instructs them to bump the version number.

This seems backward to me. Shouldn't we bump the version number first so that the updated docs show the new version number?

DMats:
I ended up doing everything in the following order.  I think we should update the ocean.py release procedure to be more like this:

1. Bump the version in ocean.py and merge to main (https://github.com/oceanprotocol/ocean.py/pull/498).  
2. Updated and merge readthedocs (https://github.com/oceanprotocol/readthedocs/pull/29). 
3. Deploy the new docs. (https://github.com/oceanprotocol/docs/pull/775)

DMats:
The v0.7.0 release can be found here: https://github.com/oceanprotocol/ocean.py/releases/tag/v0.7.0

DMats:
The Travis CI build that is triggered when new version tags are added, failed due to intermittent issues so the new v0.7.0 release didn't get pushed to PyPi.
https://app.travis-ci.com/github/oceanprotocol/ocean.py/builds/238388955

DMats:
I believe that the Travis CI build is failing because it does not export the `OPERATOR_SERVICE_URL` envvar which is required by the latest version of provider in the `main` branch of `barge`.

Unfortunately, this means that Travis CI will not be able to deploy ocean.py v0.7.0 to PyPi. And I can't deploy it manually because I don't know the PyPi username and password.

DMats:
# Status Update

:tada: ocean.py v0.8.1 is released in both [Github](https://github.com/oceanprotocol/ocean.py/releases/tag/v0.8.1) and [PyPi](https://pypi.org/project/ocean-lib/0.8.1/)

As soon as https://app.zenhub.com/workspace/o/oceanprotocol/docs/issues/779 is merged, the v0.8.1 docs will be deployed to docs.oceanprotocol.com

PR #500, opened 2021-09-25
Bump version: 0.7.0 → 0.8.0

DMats:
Towards #497 

Proposed changelog entry:

# Breaking Changes

- The following methods no longer return the compute job results URL. 
  - `OceanCompute.result()`
  - `DataServiceProvider.get_compute_status()`
  - `DataServiceProvider.get_compute_result()`

# Non-breaking changes

- Added support for the `computeResult` endpoint for getting the compute job results URL: 
  - `OceanCompute.result_file()`
  - `DataServiceProvider.compute_job_result_file()`
- Replace Travis with GitHub Actions

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.7.0
+current_version = 0.8.0
 commit = True
 tag = True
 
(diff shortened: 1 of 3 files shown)


trentmc:
When we ship this, we should also tweet from @shipping_ocean

calina-c:
> When we ship this, we should also tweet from @shipping_ocean

I'd like to test it first as part of the c2d readme and have a full, polished, up to date c2d sample.

PR #501, opened 2021-09-27
v0.8.1 with corrections

calina-c:
Proposed changelog:
- allow same-block confirmations


mariacarmina (approved):
It's ok as an intermediary tweak.

PR #516, opened 2021-10-01
Feature/re add get provider address

alexcos20:
Re-add DataServiceProvider.get_provider_address

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.3
+current_version = 0.8.4
 commit = True
 tag = True
 
--- a/ocean_lib/__init__.py
+++ b/ocean_lib/__init__.py
@@ -7,5 +7,5 @@
 
 __author__ = """OceanProtocol"""
 # fmt: off
-__version__ = '0.8.3'
+__version__ = '0.8.4'
 # fmt: on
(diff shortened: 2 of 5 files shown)


DMats (changes requested):
Since there's no issue to go with this, can you briefly explain why this method is needed?

Also, if we're adding the method back in, we should add the unit tests back in too.  

ocean_lib/data_provider/test/test_data_service_provider.py
```python
def test_provider_address(config):
    """Tests that a provider address exists on the DataServiceProvider."""
    provider_uri = DataSP.get_url(config)
    provider_address = DataSP.get_provider_address(provider_uri)
    assert provider_address, "Failed to get provider address."


def test_provider_address_with_url():
    """Tests that a URL version of provider address exists on the DataServiceProvider."""
    p_ocean_instance = get_publisher_ocean_instance()
    provider_address = DataSP.get_provider_address(
        DataSP.get_url(p_ocean_instance.config)
    )
    assert provider_address, "Failed to get provider address."
    assert DataSP.get_provider_address("not a url") is None
```

alexcos20:
How can you differentiate 2 assets, published by 2 providers which are using the same c2d backend ?

DMats (approved):
These changes look good to me so I'm approving.

Side note: The `test_provider_address_with_url` test is testing the `get_c2d_address` method, but the test name and the assertions still say "provider address". We should probably make an issue to fix this at some point.

DMats (approved):

PR #519, opened 2021-10-05
Added support for mumbai network.

### LINKED ISSUE #518: Add support for mumbai testnet
Add support for mumbai testnet in ExampleConfig class.

mariacarmina:
Fixes #518  .

Changes proposed in this PR:

- add support for mumbai network in `example-config`;
- update constants for mumbai network.

### DIFF
--- a/ocean_lib/example_config.py
+++ b/ocean_lib/example_config.py
@@ -87,6 +87,12 @@ CONFIG_NETWORK_HELPER = {
         NETWORK_NAME: "celoalfajores",
         NAME_BLOCK_CONFIRMATIONS: 3,
     },
+    80001: {
+        NAME_PROVIDER_URL: "https://provider.mumbai.oceanprotocol.com",
+        NAME_METADATA_CACHE_URI: "https://aquarius.oceanprotocol.com",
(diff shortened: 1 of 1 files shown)


ocean_lib/example_config.py
```diff
@@ -87,6 +87,12 @@
         NETWORK_NAME: "celoalfajores",
         NAME_BLOCK_CONFIRMATIONS: 3,
     },
+    80001: {
+        NAME_PROVIDER_URL: "https://provider.mumbai.oceanprotocol.com",
+        NAME_METADATA_CACHE_URI: "https://aquarius.oceanprotocol.com",
```
calina-c, line 92: @mariacarmina I'm thinking whether now, with universal aquarius, we can stop pasting this into the config. Can we set it independently?

mariacarmina: Ok, I have tested the marketplace flow and it works fine for ganache as well since it had the different aquarius URL.

mariacarmina: UPDATE: Created #525 to fix the duplication of the aquarius URL.

PR #551, opened 2021-11-02
Fixed rate exchange unit tests

### LINKED ISSUE #540: Implement unit tests for Fixed Rate Exchange in V4
Implement unit tests for Fixed Rate Exchange in V4 that are found here: https://github.com/oceanprotocol/contracts/blob/v4main/test/flow/FixedRateExchange.test.js

Translate methods from js to py and create functions for expect failure as well (See here some examples: https://github.com/oceanprotocol/ocean.py/pull/512/commits/1d9b8ccf8fe17a2326e1229e0433a19e45208522#diff-079668b3e23d8b89cadd6e91426901b67c828edce699278f410696f3c5147de0R102)

NOTE: The ones that repeat the same workflow, you can include them in one test, no need to create tests that do not have different flow.

maldo-ctrl:
Fixes #540 .

Changes proposed in this PR:

- Tests Fixed rate exchange creation workflow and interactions
- Covers `ocean_lib.models.v4.fixed_rate_exchange` module
- Coverage so far: 96%
- 
image

image

### DIFF
--- a/ocean_lib/models/v4/fixed_rate_exchange.py
+++ b/ocean_lib/models/v4/fixed_rate_exchange.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 from typing import List
+from enum import IntEnum
 
 from enforce_typing import enforce_types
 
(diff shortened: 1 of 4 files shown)


ocean_lib/models/v4/fixed_rate_exchange.py
```diff
@@ -3,18 +3,42 @@
...
 from typing import List
+from enum import IntEnum
 
 from enforce_typing import enforce_types
 
 from ocean_lib.web3_internal.contract_base import ContractBase
 from ocean_lib.web3_internal.wallet import Wallet
 
 
+class FixedRateExchangeV4Details(IntEnum):
```
mariacarmina, line 14: For me looks like a good approach to use enums classes here. @calina-c what do you think?

ocean_lib/models/v4/test/fixed_rate_exchange.py
```diff
@@ -0,0 +1,200 @@
...
+    web3, config, publisher_wallet, consumer_wallet, another_consumer_wallet
+):
+    """Test exchange with baseToken(OCEAN) 18 Decimals and dataToken 18 Decimals, RATE = 1"""
+    cap = web3.toWei("100000", "ether")
+    amount_dt_to_sell = web3.toWei("100", "ether")
+    no_limit = web3.toWei("100000000000000000000", "ether")
+    rate = web3.toWei("1", "ether")
+    market_fee = int(1e15)  # 0.1%
+    opf_fee = int(1e15)  # 0.1%
+    ocean = ERC20Token(web3, get_address_of_type(config, "Ocean"))
```
mariacarmina, line 82: can we specify here `ocean_token` instead of `ocean`?

tests/resources/helper_functions.py
```diff
@@ -213,11 +213,14 @@ def verify_supply(mint_amount=to_wei(50)):
 
 
 def deploy_erc721_erc20(
-    web3: Web3, config: Config, erc721_publisher: Wallet, erc20_minter: Wallet
+    web3: Web3,
+    config: Config,
+    erc721_publisher: Wallet,
+    erc20_minter: Wallet,
+    cap=500000000000000000,
```
mariacarmina, line 220: maybe we can use web3.toWei("0.05", "ether") instead of putting this large numbers.

tests/resources/helper_functions.py
```diff
@@ -282,4 +285,4 @@ def get_non_existent_nft_template(
         if address == ZERO_ADDRESS:
             return x
 
-    return -1
\ No newline at end of file
+    return -1
```
mariacarmina, line 288: In the helper function can you replace "x" with other name to be more specific. Other than that, @calina-c are you satisfied by using `range()` in the [for](https://github.com/oceanprotocol/ocean.py/pull/551/files#diff-8d3fd885d643390850dc0ddada99085063dfc0195d3f87c9f0fb4c163d144a1aR283) loop?

ocean_lib/models/v4/fixed_rate_exchange.py
calina-c: I am not a fan of Enums since they make the usage of the class more unintuitive, but my feelings are not particularly strong on this.

calina-c (approved):

Issue #520, opened 2021-10-05
[EPIC] Refactor ocean.py architecture: unify Asset and DDO classes, more


alexcos20:
Note: this ticket was initially a discussion about whether to re-boot ocean.py or refactor it. That discussion happened, and the choice is to refactor it. 

Here's the current list by Calina from [this comment](https://github.com/oceanprotocol/ocean.py/issues/520#issuecomment-938491658).

They are all currently _proposals_. It may be that we don't do some of them, it will be a judgement call as we proceed. 

* unify Asset and DDO classes. If we need to separate some code just for readability purposes, let's create mixins that deal with different aspects of the DDO. Right now, it is confusing to have two classes dealing with the same concept. E.g. Asset is a child of DDO, and it also defines trusted algo functions. What if I am using a DDO and I want to add trusted algorithms? I have to use the Asset base class instead, and it's cumbersome to figure this out.
* move create_checksum out of assets/utils.py and rename the file to trusted_algorithms.py (create_checksum is the only outlier) -> perfect place for create_checksum is in utilities.py, right NEXT TO checksum()
* move contents of common/agreemeents outside: it now contains fewer things (after Service refactoring), and if we unify Asset and DDO, it will be even lighter and easier to move
* remove AquariusProvider… simply retrieve the Aquarius class based on the config
* move ddo/constants.py contents to agreements
* convert AddressCredential class into a mixin -> keeps the code separate, but makes it more flexible
* common/ddo/README.md -> either move it to READMEs, or remove it entirely (it is leftover from the plecos, as its original README.md)
* investigate DID.py, it probably needs some attention
* Data Service Provider (there is already a ticket for this: #471)

trentmc:
This is new. And a big decision. Please share context / info. Specifically:
- what are the specific concerns of the current ocean.py
- Please document pros and cons of keeping the current codebase versus a new one.

I always fear "second system syndrome". It bit me once early in my career. I acknowledge that sometimes it is appropriate.

calina-c:
> This is new. And a big decision. Please share context / info. Specifically:
> 
> * what are the specific concerns of the current ocean.py
> * Please document pros and cons of keeping the current codebase versus a new one.
> 
> I always fear "second system syndrome". It bit me once early in my career. I acknowledge that sometimes it is appropriate.

Hi, Trent! We discussed this during our last backend grooming session. Here's are the main arguments for a new v4 version of ocean.py:
[...]

PR #601, opened 2021-12-02
Kill common

calina-c:
Related to #520


calina-c:
LGTM. I can not approve because I'm the original author.

PR #604, opened 2021-12-06
Implement publish flow

### LINKED ISSUE #489: Publish flow
- [ ] add functional tests

mariacarmina:
Fixes #489  .

Changes proposed in this PR:

- implemented `create` function into `OceanAssetV4`;
- implemented helper functions in ocean/models/v4 and in `OceanAssetV4` class;
- updated DDO sample from the ddo helpers;
- fixed encrypt endpoint;
- add unit tests in `tests/integration/test_publish_flow`;
- add docstrings.


ocean_lib/assets/v4/asset.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         nft: Optional[dict] = None,
         datatokens: Optional[list] = None,
         event: Optional[dict] = None,
+        purgatory: Optional[dict] = None,
```
calina-c, line 34: why would you need purgatory? Aquarius shouldn't serve files that are in purgatory, at all. So there's no way they would end up in ocean.py. If there was no purgatory in v3, don't add it in v4, either.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+        self._downloads_path = downloads_path
+
+    def validate(self, metadata: dict) -> Tuple[bool, list]:
+        """
+        Validate that the metadata is ok to be stored in aquarius.
+
+        :param metadata: dict conforming to the Metadata accepted by Ocean Protocol.
+        :return: (bool, list) list of errors, empty if valid
+        """
+        return self._get_aquarius(self._metadata_cache_uri).validate_metadata(metadata)
```
calina-c, line 64: Aquarius in v4 no longer validates metadata. It validates the full asset :)

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def _add_defaults(
+        self, services: list, data_token: str, files: str, provider_uri: str
+    ) -> list:
+
+        has_access_service = False
+        for service in services:
```
calina-c, line 71: please use a comprehension

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+        # copy metadata so we don't change the original
+        metadata_copy = copy.deepcopy(metadata)
+
+        asset_type = metadata_copy["type"]
+        assert asset_type in (
+            "dataset",
+            "algorithm",
+        ), f"Invalid/unsupported asset type {asset_type}"
+
+        validation_result, validation_errors = self.validate(metadata)
```
calina-c, line 116: please make adjustments here, as well. We validate the asset, not the metadata.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+                raise ContractNotFound(
+                    f"NFT address {nft_address} is not found in the ERC721Factory events."
+                )
+
+        assert nft_address, "nft_address is required for publishing a dataset asset."
+
+        # Generating the did and adding to the ddo.
+        did = f"did:op:{remove_0x_prefix(nft_address)}"
+        asset.did = did
+        logger.debug(f"Using datatoken address as did: {did}")
```
calina-c, line 168: we are not doing that anymore.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+            # verify nft address
+            if not erc721_factory.verify_nft(nft_address):
+                raise ContractNotFound(
+                    f"NFT address {nft_address} is not found in the ERC721Factory events."
+                )
+
+        assert nft_address, "nft_address is required for publishing a dataset asset."
+
+        # Generating the did and adding to the ddo.
+        did = f"did:op:{remove_0x_prefix(nft_address)}"
```
calina-c, line 166: did has a new format, see in the docs, and implemented in the provider and aquarius tests.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,200 @@
...
+            asset.add_service(service)
+
+        # TODO: provider endpoint encrypt urls
+        # TODO: update this call to setMetaData of the NFT -> Store Metadata role
+        # TODO: wait until asset appears in Aqua -> verify if aqua receives it
+
+        return asset
+
+    @staticmethod
+    def build_access_service(
```
calina-c, line 185: is this used anywhere else? If it's just for the content in line 76 onwards, then call it directly there.

calina-c (changes requested):

ocean_lib/assets/v4/asset.py
mariacarmina: I saw in the docs that it was added and wanted to reproduce the whole structure. I will remove it then.

ocean_lib/ocean/v4/ocean_assets.py
mariacarmina: Instead using a list comprehension and after a filter, I prefer using `any(map(lambda s: s.type == ServiceTypesV4.ASSET_ACCESS, services))`. Is it ok like that?

ocean_lib/ocean/v4/ocean_assets.py
mariacarmina: It is also implemented in ocean_assets V3.

mariacarmina:
Screenshot from 2021-12-21 09-31-01

ocean_lib/aquarius/aquarius.py
```diff
@@ -156,6 +156,28 @@ def validate_metadata(self, metadata: dict) -> Tuple[bool, Union[list, dict]]:
...
+        """
+        asset_dict = asset.as_dictionary()
+
+        response = self.requests_session.post(
+            f"{self.base_url.replace('/v1/', '/')}/ddo/validate-remote",
+            data=json.dumps(asset_dict),
+            headers={"content-type": "application/json"},
+        )
+
+        if response.content == b"true\n":
```
calina-c, line 175: please use the decoded the binary response instead, which should be plain `True`. I think using response.json() is enough but you should be able to figure that out with pdb.

calina-c: i.e. parsed_response = response.json()
if parsed_response is True: return True, []
return False, parsed_response

ocean_lib/assets/v4/test/test_asset.py
```diff
@@ -92,7 +92,10 @@ def test_asset_utils(web3):
     }
     event = ddo_dict["event"]
 
-    assert ddo_dict["stats"] == {"consumes": 4, "isInPurgatory": "false"}
+    assert ddo_dict["purgatory"] == {"state": False}
+    purgatory = ddo_dict["purgatory"]
+
+    assert ddo_dict["stats"] == {"consumes": 4}
```
calina-c, line 98: how did "4" happen here?

ocean_lib/assets/v4/test/test_asset.py
```diff
@@ -105,6 +108,7 @@ def test_asset_utils(web3):
         nft=nft,
         datatokens=data_tokens,
         event=event,
+        purgatory=purgatory,
```
calina-c, line 111: no, assets that are in purgatory are never returned by Aquarius, so you don't need to include this in the asset.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -109,21 +109,21 @@ def encrypt_files_dict(
     @staticmethod
     @enforce_types
     def encrypt(
-        objects_to_encrypt: Union[list, str], encrypt_endpoint: str
+        objects_to_encrypt: Union[list, str, bytes], encrypt_endpoint: str
     ) -> Response:
         if isinstance(objects_to_encrypt, list):
             data = str(list(map(lambda file: file.from_dict(), objects_to_encrypt)))
```
calina-c, line 115: are you sure this results in a stripped json? Without spaces?

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+            tx_receipt.blockNumber,
+            self._web3.eth.block_number,
+            None,
+        )
+        assert registered_token_event, "Cannot find TokenCreated event."
+
+        return registered_token_event[0].args.newTokenAddress
+
+    def find_service_by_data_token(self, data_token: str, services: list) -> str:
+        for service in services:
```
calina-c, line 136: please use the next() syntax

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+            if service.data_token == data_token:
+                return service.id
+
+    def build_data_tokens_list(
+        self, services: list, deployed_erc20_tokens: list
+    ) -> list:
+        data_tokens = []
+        for erc20_token in deployed_erc20_tokens:
```
calina-c, line 144: remove. Alongside the entire "nft" key. It is not your job to build this.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+                f"Asset id {did} is already registered to another asset."
+            )
+        asset.chain_id = self._web3.eth.chain_id
+        asset.metadata = metadata
+
+        asset.credentials = credentials if credentials else build_credentials_dict()
+
+        erc20_addresses = []
+        services = services or []
+        deployed_erc20_tokens = deployed_erc20_tokens or []
```
calina-c, line 274: I'm sorry but this is very confusing and I'm not sure what is happening here. I do not understand the connection between the tokens and the services. Why are we "guessing" correspondences? Why do we have a list of data_tokens which is different from the deployed list? I know we want to allow for multiple options but I'm not sure this is it. And nobody seems to know what we are trying to build.

The basic flow with one datatoken, one access service? Ok, we are customising this a lot, but we are not organised in any way. @AnaLoznianu who should I ask for help here? This is beyond my knowledge of what ocean.py should do. I need help ASAP.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+            data_tokens = self.build_data_tokens_list(
+                services=services, deployed_erc20_tokens=deployed_erc20_tokens
+            )
+
+        asset.datatokens = data_tokens
+
+        for service in services:
+            asset.add_service(service)
+
+        nft = {
```
calina-c, line 314: it is not ocean.py's job to build the nft key. That is added by aquarius AFTER the MetadataCreated event is processed. It has nothing to do with validation and you shouldn't build it.

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+
+        # Validation by Aquarius
+        validation_result, validation_errors = self.validate(asset)
+        if not validation_result:
+            msg = f"Asset has validation errors: {validation_errors}"
+            logger.error(msg)
+            raise ValueError(msg)
+
+        # Process the DDO
+        asset_dict = asset.as_dictionary()
```
calina-c, line 332: where are the separators?

ocean_lib/ocean/v4/ocean_assets.py
```diff
@@ -0,0 +1,408 @@
...
+        ddo_string = json.dumps(asset_dict)
+        ddo_bytes = ddo_string.encode("utf-8")
+        ddo_hash = create_checksum(ddo_string)
+
+        # Plain asset
+        if not encrypt_flag and not compress_flag:
+            encrypted_ddo = ddo_bytes
+            _ = erc721_token.set_metadata(
```
calina-c, line 340: there are multiple calls of this function that differ only by flags, data and data_hash. Do you think we can reuse the rest of the params?

tests/integration/test_publish_flow.py
```diff
@@ -2,31 +2,36 @@
...
 def test_publish_flow(web3, config, publisher_wallet):
-
+    """Tests publish flow in different situations."""
     erc721_factory_address = get_address_of_type(
         config, ERC721FactoryContract.CONTRACT_NAME
     )
     erc721_factory = ERC721FactoryContract(web3, erc721_factory_address)
 
     # Publisher deploys NFT contract
     tx = erc721_factory.deploy_erc721_contract(
```
calina-c, line 26: I think this call was was changed, so we need to rebase with v4main after merging Pablo's new artifacts work.

calina-c (changes requested):

ocean_lib/assets/v4/test/test_asset.py
mariacarmina: It's a sample asset, it's just for testing.

calina-c:
I updated my review and resolved appropriate issues. Only separator things and the rebase are pending.

calina-c (approved):

PR #609, opened 2021-12-08
Integrate encrypt endpoint

### LINKED ISSUE #606: Integrate encrypt endpoint into ocean.py
In order to encrypt the files for the DDO, ocean.py needs to add support for the encrypt endpoint from provider. Need for create asset function that is used in publish flow.

Linked provider issue: https://github.com/oceanprotocol/provider/issues/283

mariacarmina:
Fixes #606  .

Changes proposed in this PR:

- add classes for file objects;
- integrate the endpoint encrypt in `data_service_provider`;
- add test for it.


ocean_lib/data_provider/data_service_provider.py
```diff
@@ -105,6 +106,37 @@ def encrypt_files_dict(
...
+        encrypt_endpoint: str,
+    ) -> str:
+        data = str(list(map(lambda file: file.from_dict(), files)))
+        response = DataServiceProvider._http_method(
+            "post",
+            encrypt_endpoint,
+            data=data,
+            headers={"content-type": "application/octet-stream"},
+        )
+        if response and hasattr(response, "status_code"):
```
calina-c, line 122: please remove the multiple if imbrication by using my favourite tactic :D

ocean_lib/agreements/file_objects.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def to_dict(self) -> dict:
+        return {"type": self.type, "hash": self.hash}
+
+    def from_dict(self) -> str:
+        return json.dumps(self.to_dict(), separators=(",", ":"), indent=4)
+
+
+@enforce_types
+def Factory(file_obj: dict) -> Union[UrlFile, IpfsFile]:
```
calina-c, line 39: please do not name a Factory Factory, be more specific. FileTypeFactory, perhaps? I like the approach, though! 🥳

mariacarmina: FileTypeFactory works for me as well. 👍🏻

ocean_lib/data_provider/data_service_provider.py
calina-c: ??? the other way around! This will return "" if there is a status code. It will never reach the next if!

calina-c: I don't know how this was resolved. And I don't know how it is working.

calina-c (changes requested):
see comment.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -105,6 +106,44 @@ def encrypt_files_dict(
...
+        )
+        response = DataServiceProvider._http_method(
+            "post",
+            encrypt_endpoint,
+            data=payload,
+            headers={"content-type": "application/octet-stream"},
+        )
+
+        if not response or not hasattr(response, "status_code"):
+            return ""
```
DMats, line 130: I think this function should throw an exception in this case.

DMats (approved):

ocean_lib/data_provider/data_service_provider.py
calina-c: I agree

ocean_lib/data_provider/test/test_data_service_provider.py
```diff
@@ -161,6 +163,21 @@ def test_delete_job_result(with_nice_client):
...
+    file2_dict = {"type": "url", "url": "https://url.com/file2.csv", "method": "GET"}
+    file1 = FilesTypeFactory(file1_dict)
+    file2 = FilesTypeFactory(file2_dict)
+
+    result = DataSP.encrypt_urls(
+        [file1, file2], "http://localhost:8030/api/services/encrypt"
+    )
+
+    assert isinstance(result, bytes)
+    assert result.decode("utf-8").startswith("0x")
```
calina-c, line 178: I don't think this is enough to assert. Can you assert some things about the content itself, or maybe decrypt and check the content is the same?

calina-c (changes requested):

calina-c (approved):

PR #636, opened 2022-01-05
Consume flow

### LINKED ISSUE #490: Consume flow

mariacarmina:
Fixes #490  .

Changes proposed in this PR:

- implemented consume flow;
- add `download` and `initialize` endpoints to the DataServiceProivder class;
- updated mock DSP;
- fixed `asset_downloader tests`.


ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,100 @@ def encrypt(
...
-    def sign_message(wallet: Wallet, msg: str) -> str:
+    def initialize(
+        did: str,
+        service_id: str,
+        file_index: int,
+        consumer_address: str,
+        service_endpoint: str,
+        userdata: Optional[Dict] = None,
+    ) -> Response:
+        initialize_endpoint = service_endpoint
```
calina-c, line 113: please use either preparedUrl or pass the GET query as a parameter. We are not doing '90s PHP here, building URLs manually.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,100 @@ def encrypt(
...
+
+        return response
+
+    @staticmethod
+    @enforce_types
+    def download(
+        did: str,
+        service_id: str,
+        tx_id: str,
+        file_index: int,
```
calina-c, line 146: I think here we should also have the None option (that downloads all files)

tests/integration/test_consume_flow.py
```diff
@@ -0,0 +1,147 @@
...
+            tx_id=tx_id,
+            file_index=i,
+            consumer_wallet=consumer_wallet,
+            download_endpoint="http://172.15.0.4:8030/api/services/download",
+            destination_folder=destination,
+        )
+        assert response
+        assert response.status_code == 200
+
+    assert len(os.listdir(destination)) >= 1, "The asset folder is empty."
```
calina-c, line 147: please make this assertion tighter: count the files before and after downloading, and check that the difference is == to the number of files.

tests/resources/helper_functions.py
```diff
@@ -345,3 +346,37 @@ def transfer_ocean_if_balance_lte(
         )
 
     return ocean_token.balanceOf(recipient) - initial_recipient_balance
+
+
+def get_provider_fees() -> Dict[str, Any]:
```
calina-c, line 351: Shouldn't you get the provider fees from the provider itself?

tests/resources/mocks/data_provider_mock.py
```diff
@@ -72,55 +75,60 @@ def build_download_endpoint(provider_uri=None):
...
+        userdata: Optional[Dict] = None,
     ):
+        initialize_endpoint = service_endpoint
+        initialize_endpoint += f"?documentId={did}"
+        initialize_endpoint += f"&serviceId={service_id}"
+        initialize_endpoint += f"&fileIndex={file_index}"
+        initialize_endpoint += f"&consumerAddress={consumer_address}"
+        if userdata:
+            initialize_endpoint += f"&userdata={json.dumps(userdata)}"
+        DataServiceProvider._http_method("get", initialize_endpoint)
```
calina-c, line 93: I don't understand the point. Why use a mock when you eventually use the main DataServiceProvider? This mock is just a passthrough.

calina-c:
One more thing you forgot: reinstating the download function in OceanAssets and unskipping + correcting the test in test_ocean_assets.

ocean_lib/assets/asset_downloader.py
```diff
@@ -16,77 +16,69 @@
...
         raise AssertionError(
             'Consume asset failed, service definition is missing the "serviceEndpoint".'
         )
+    service_id = service.id
+
+    assert isinstance(file_index, int), logger.error("index has to be an integer.")
+    assert file_index >= 0, logger.error("index has to be 0 or a positive integer.")
+
+    service_endpoint += "/api/services/initialize"
+    data_provider.initialize(
```
DMats, line 55: Why don't we store the initialize response?

calina-c: yes, that's the next step. Please use the initialize response in the download inputs.

ocean_lib/assets/asset_downloader.py
```diff
@@ -16,77 +16,69 @@
...
-            _files,
-            asset_folder,
-            service_index,
-            token_address,
-            order_tx_id,
-            i,
-            userdata,
-        )
+    service_endpoint.replace("initialize", "download")
+    data_provider.download(
```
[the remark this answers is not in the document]
calina-c: if file index is None, please download all the files. I saw you made it optional in the params list, so please add the logic, as well.

ocean_lib/data_provider/test/test_data_service_provider.py
```diff
@@ -173,6 +181,51 @@ def test_encrypt(web3, provider_wallet):
...
+        did=ddo.did,
+        service_id=access_service.id,
+        file_index=0,
+        consumer_address=publisher_wallet.address,
+        service_endpoint="http://172.15.0.4:8030/api/services/initialize",
+    )
+    assert initialize_result
+    assert initialize_result.status_code == 200
+    response_json = initialize_result.json()
+    assert response_json["providerFee"] == get_provider_fees()
```
calina-c, line 226: this is good. The providerFees returned from our test provider will probably always correspond to our test version. If this changes, we'll adapt the test accordingly. Leave it as is for now.

tests/resources/helper_functions.py
calina-c: To clarify: when testing the initialize function separately, this is perfect. So leave it in the code.
But when actually doing the flow (initialize+download), please use the provider fees you get from the initialize step.

calina-c (changes requested):
I left the old unresolved issues and added some new comments.

tests/integration/test_consume_flow.py
mariacarmina: == len(files)

tests/resources/mocks/data_provider_mock.py
mariacarmina: I will delete those methods from the mock class.

ocean_lib/assets/asset_downloader.py
```diff
@@ -16,77 +16,62 @@
...
 
 
-# TODO: fix for v4
-
-
 @enforce_types
 def download_asset_files(
-    service_index: int,
     asset: Asset,
+    provider_uri: str,
```
calina-c, line 22: the DataProvider must have its own provider_uri, we shouldn't pass it separately.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,108 @@ def encrypt(
...
+
+        return response
+
+    @staticmethod
+    @enforce_types
+    def download(
+        did: str,
+        service_id: str,
+        tx_id: str,
+        files: List[Dict[str, Any]],
```
calina-c, line 151: please revert this change of passing the files dict. We are just blindly downloading files from an asset. I am not the publisher, files are encrypted, I don't want to call fileinfo. I bought the asset and I want to download it. I want to have the option of not passing in files dicts, files indices, file lengths etc.

-> Yes, I _could_ have knowledge and _choose_ to pass in the file index, as this was previously implemented. But not file objects and stuff like that. This is a library to help people use OCEAN, not make them use our File Object etc. Those are for our testing purposes and for "advanced" users. In my previous review, I asked to have fileIndex as an int or as None. Not to make this a file object, dict or something else.

calina-c (changes requested):
Please address the two issues added now before moving forward.

ocean_lib/assets/asset_downloader.py
mariacarmina: I need provider_uri for building the URL for download and if I don't pass provider_uri as parameter, I need to pass a config object from where I retrieve provider_url as I used in the tests with a fixture config.

ocean_lib/data_provider/data_service_provider.py
mariacarmina: I have the following problem: provider download endpoint requires `fileIndex` field, otherwise it will throw an error (https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/consume.py#L282). As you mentioned above, files are encrypted and we don't want to pass a list of files. In this case, I suggest to implement the `fileinfo` endpoint and call it inside the download function from `DataServiceProvider` class, so that the user is not obliged to call it separately. I think this is the easiest way to implement the proposed functionality avoiding passing the file list.
BTW, in V3, the file list was passed to the download service as it is shown [here](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/data_provider/data_service_provider.py#L197) and there were only 2 fields from the payload that differs from the V4 one. @DMats, @alexcos20, what do you think?

mariacarmina: UPDATE: The solution will be to implement `fileinfo` endpoint in ocean.py and use it inside the download endpoint when the parameter `index` is None.

ocean_lib/assets/asset_downloader.py
calina-c: DataServiceProvider already has a property set. If the property is not set, it will grab the default provider_uri from config. I assure you there is no need for provider_uri if there is a DataServiceProvider object.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,137 @@ def encrypt(
 
     @staticmethod
     @enforce_types
-    def sign_message(wallet: Wallet, msg: str) -> str:
+    def fileinfo(did: str, service_id: str, service_endpoint: str) -> Response:
```
calina-c, line 105: there's no fileinfo?!?! Why do we keep adding endpoints that most certainly already exist?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,137 @@ def encrypt(
...
+
+        logger.info(
+            f"Retrieved asset files successfully"
+            f" FileInfoEndpoint {service_endpoint}"
+        )
+        return response
+
+    @staticmethod
+    @enforce_types
+    def initialize(
```
calina-c, line 131: by the way, initialize is a duplicate of the order_requirements function, I think. Please delete the previous version and make sure to adapt any of its usages.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -103,7 +102,137 @@ def encrypt(
...
+        )
+        files = fileinfo_response.json()
+        indexes = range(len(files))
+        if index is not None:
+            assert isinstance(index, int), logger.error("index has to be an integer.")
+            assert index >= 0, logger.error("index has to be 0 or a positive integer.")
+            assert index < len(files), logger.error(
+                "index can not be bigger than the number of files"
+            )
+            indexes = [index]
```
calina-c, line 195: btw the correct plural of "index" is "indices"

ocean_lib/data_provider/test/test_data_service_provider.py
```diff
@@ -173,6 +182,59 @@ def test_encrypt(web3, provider_wallet):
     assert decrypted_document_string == test_string
 
 
+def test_fileinfo(web3, config, publisher_wallet, publisher_ocean_instance):
```
calina-c, line 185: there's also another test for this, using the former version of fileinfo...

ocean_lib/data_provider/data_service_provider.py
mariacarmina: https://stackoverflow.com/questions/1378781/proper-terminology-should-i-say-indexes-or-indices both are correct :))

ocean_lib/assets/asset_downloader.py
mariacarmina: I will delete DataServiceProvider object and keep just `provider_uri`.

ocean_lib/data_provider/data_service_provider.py
mariacarmina: There are only check functions which implement the fileinfo request inside them. There is not a specific method of fileinfo in ocean.py.

ocean_lib/assets/test/test_asset_downloader.py
```diff
@@ -5,98 +5,139 @@
...
-            "",
             "test_order_tx_id",
-            data_provider,
-            index,
+            index=index,
         )
+
     index = 4
-    with pytest.raises(AssertionError):
+    with pytest.raises(Exception):
```
calina-c, line 78: we never assert raising a plain Exception. If this Exception comes from provider, make sure you catch it and rethrow an AssertionError instead.

ocean_lib/assets/test/test_asset_downloader.py
```diff
@@ -5,98 +5,139 @@
...
 
     written_path = download_asset_files(
-        sa.index,
-        ddo,
-        publisher_wallet,
-        tmpdir,
-        "0x1",
-        "test_order_tx_id",
-        data_provider,
+        ddo, config.provider_url, publisher_wallet, tmpdir, tx_id
```
calina-c, line 140: what is the difference between this test and the consume flow? I would rather keep the tests here, but please make sure you assert more than the path existence, rather copy the part with len(dir) here.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -205,33 +187,47 @@ def download_service(
             indexes = [index]
 
         req = PreparedRequest()
-        params = {
-            "documentId": did,
+        service_endpoint += f"?documentId={did}"
```
calina-c, line 190: why do we have some parameters added like this, and some by using payload?

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -205,33 +187,47 @@ def download_service(
...
+            if not response or not hasattr(response, "status_code"):
+                raise Exception("Response not found!")
+
+            if response.status_code != 200:
+                msg = (
+                    f"Download asset failed at the downloadEndpoint "
+                    f"{req.url}, reason {response.text}, status {response.status_code}"
+                )
+                logger.error(msg)
+                raise Exception(msg)
```
calina-c, line 218: we do not raise a plain Exception. Never, ever. Please see my other comment.

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -237,18 +211,23 @@ def test_ocean_assets_algorithm(publisher_ocean_instance, publisher_wallet, conf
     assert ddo, "DDO None. The ddo is not cached after the creation."
 
 
-@pytest.mark.skip(reason="TODO: download function on OceanAssets class")
 def test_download_fails(publisher_ocean_instance, publisher_wallet):
     """Tests failures of assets download function."""
     with patch("ocean_lib.ocean.ocean_assets.OceanAssets.resolve") as mock:
-        mock.return_value = get_sample_ddo()
+        mock.return_value = Asset.from_dict(get_sample_ddo())
+        asset = mock.return_value
```
calina-c, line 218: please switch these two statements. Assign the asset first, then set it as the mock.return_value.

tests/integration/test_consume_flow.py
```diff
@@ -0,0 +1,159 @@
...
+        value=to_wei(50),
+        from_wallet=publisher_wallet,
+    )
+
+    # Initialize service
+    response = data_provider.initialize(
+        did=ddo.did,
+        service_id=service.id,
+        consumer_address=consumer_wallet.address,
+        service_endpoint=data_provider.build_initialize_endpoint(config.provider_url)[
```
calina-c, line 115: please assign the service_endpoint separately, using a _, variable = build() syntax. It is clearer if we switch the tuple count at some point. The [1] selector can apply to a list too, and it's a bit unclear.

tests/resources/helper_functions.py
```diff
@@ -345,3 +348,67 @@ def transfer_ocean_if_balance_lte(
...
+        # make it compatible with last openzepellin https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1622
+        "v": signature.v,
+        "r": signature.r,
+        "s": signature.s,
+    }
+
+    return provider_fee
+
+
+def create_basics(config, web3, data_provider):
```
calina-c, line 387: don't we have create_basics in test_ocean_assets.py? If you move it here, please remove it from test_ocean_assets.py.

calina-c (commented):
A few comments left

tests/resources/helper_functions.py
mariacarmina: I have moved it to helper functions since I used it in many places.

ocean_lib/data_provider/data_service_provider.py
mariacarmina: I recycled some chunks of the previous code from V3.

ocean_lib/data_provider/data_service_provider.py
mariacarmina: prepare_url function transforms ':' from "did:op:" into "%3".

calina-c (approved):

PR #641, opened 2022-01-10
Ocean class

### LINKED ISSUE #633: Adapt Ocean class for v4
Currently there is a lot of outdated v3 code in the Ocean class (ocean/ocean.py). We need to fix that :)

calina-c:
Closes #633.

### DIFF
--- a/ocean_lib/ocean/ocean.py
+++ b/ocean_lib/ocean/ocean.py
@@ -24,10 +24,10 @@ from ocean_lib.ocean.ocean_pool import OceanPool
 from ocean_lib.ocean.util import (
     get_address_of_type,
     get_bfactory_address,
-    get_dtfactory_address,
     get_ocean_token_address,
     get_web3,
 )
+from ocean_lib.web3_internal.constants import ZERO_ADDRESS
 from ocean_lib.web3_internal.utils import get_network_name
 from ocean_lib.web3_internal.wallet import Wallet
 from web3.datastructures import AttributeDict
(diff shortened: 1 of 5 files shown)


ocean_lib/ocean/ocean.py
```diff
@@ -117,13 +119,14 @@ def OCEAN_address(self) -> str:
...
         self,
         name: str,
         symbol: str,
+        token_uri: str,
         from_wallet: Wallet,
-        cap: int = DataToken.DEFAULT_CAP,
-        blob: str = "",
+        template_index: Optional[int] = 1,
+        additional_erc20_deployer: Optional[str] = None,
     ) -> DataToken:
```
mariacarmina, line 130: This should return ERC721Token. I have an example for this function [here](https://github.com/oceanprotocol/ocean.py/blob/simple-flow/ocean_lib/ocean/ocean.py#L122)

ocean_lib/ocean/ocean.py
```diff
@@ -149,34 +152,57 @@ def create_data_token(
...
-        return DataToken(self.web3, token_address)
+        return ERC20Token(self.web3, token_address)
 
     @enforce_types
-    def get_dtfactory(self, dtfactory_address: str = "") -> DTFactory:
+    def get_nft_factory(self, nft_factory_address: str = "") -> ERC721FactoryContract:
         """
         :param dtfactory_address: contract address, str
 
         :return: `DTFactory` instance
```
mariacarmina, line 196: Can we update docstrings here?

ocean_lib/ocean/ocean.py
calina-c: you're right, I'll change the return type and handle your docstrings comment too.

mariacarmina (approved):

PR #646, opened 2022-01-11
Reinstate update asset functionality in OceanAssets

### LINKED ISSUE #628: Reinstate the update function in OceanAssets
Right now we have flows for publishing (creating) an asset. But the update flows are missing.
- reinstate the update function in OceanAssets. Here is how it looked in v3: https://github.com/oceanprotocol/ocean.py/blob/e914696b126e2a4304752ca4bd4c9b32380936c2/ocean_lib/ocean/ocean_assets.py#L333
- reinstate and adapt the `test_update` function in `ocean_lib/ocean/test/test_ocean_assets.py`

I have marked #627 as a blocker, but it is more like a dependency. You can start from that branch and later cherry-pick if it is not merged by the time you start.

md0x:
Signed-off-by: Pablo Maldonado <pablo@oceanprotocol.com>

Fixes #628 .

Changes proposed in this PR:

- Reinstate update asset function in OceanAssets
- Create `process_ddo` function for shared processes between `create` and `update` function 
- Update tests


ocean_lib/ocean/ocean_assets.py
```diff
@@ -320,46 +370,9 @@ def create(
...
-            compressed_document = lzma.compress(ddo_bytes)
-
-            # Encrypt DDO
-            encrypt_response = DataServiceProvider.encrypt(
-                objects_to_encrypt=compressed_document,
-                encrypt_endpoint=f"{provider_uri}/api/services/encrypt",
-            )
-
-            document = encrypt_response.text
+        [document, flags, ddo_hash] = self.process_ddo(
```
calina-c, line 373: you don't need the brackets, Python supports multiple returns and will treat them as a tuple by default

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -39,43 +40,56 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+            publisher_wallet.address,
+            ZERO_ADDRESS,
+            get_address_of_type(config, "Ocean"),
+        ],
+        uints=[ocean.web3.toWei("0.5", "ether"), 0],
+        bytess=[b""],
+    )
 
-    ddo = create_asset(ocn, alice, config)
+    if not metadata:
```
calina-c, line 59: well, we are testing the update. When updating metadata, metadata should be required, so the option of metadata being empty/None is moot.

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -39,43 +40,56 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+            "created": "2020-11-15T12:27:48Z",
+            "updated": "2021-05-17T21:58:02Z",
+            "description": "Sample description",
+            "name": "Sample asset",
+            "type": "dataset",
+            "author": "OPF",
+            "license": "https://market.oceanprotocol.com/terms",
+        }
+    data_provider = DataServiceProvider
+    file1_dict = {"type": "url", "url": "https://url.com/file1.csv", "method": "GET"}
```
calina-c, line 70: I think we are overusing this in tests. You can just create a UrlFile with url_file.url = ...
instead of creating this from_dict everytime.

but what if we update the files themselves? Or the service structure? Then it's not so simple. Here we are basically attaching a random file, but what if we want to change the files as part of the update? Or the service structure? I don't think the task is complete just by setting the metadata. Can you address this on Slack?

calina-c (changes requested):

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,185 @@ def create(
...
-                objects_to_encrypt=ddo_string,
-                encrypt_endpoint=f"{provider_uri}/api/services/encrypt",
+        return asset
+
+    def update(
+        self,
+        did: str,
+        new_metadata: dict,
+        publisher_wallet: Wallet,
+        delete_erc20_tokens: Optional[List[str]] = [],
```
calina-c, line 403: can't you figure this parameter out by comparing the current list against the new list?

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,185 @@ def create(
...
+        :param new_erc20_tokens_data: list of ERC20CreateData necessary for deploying ERC20 tokens for different services.
+        :param new_deployed_erc20_tokens: list of ERC20 tokens which are already deployed.
+        :param encrypt_flag: bool for encryption of the DDO.
+        :param compress_flag: bool for compression of the DDO.
+        :return: update tx_result
+        """
+
+        old_asset = self._get_aquarius(self._metadata_cache_uri).wait_for_asset(did)
+
+        assert old_asset, "Asset not found."
```
calina-c, line 432: actually this could be a possibility: what if the old metadata state is expired, and we are updating it to active? Aquarius won't serve this asset, but we still want to update it. Not sure if the case is valid, but it's a discussion worth having.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,185 @@ def create(
...
 
-        # Encrypted & compressed
+        assert erc721_address, "nft_address is required for publishing a dataset asset."
+        erc721_token = ERC721Token(self._web3, erc721_address)
+
+        # Copy original asset
+        asset = copy.deepcopy(old_asset)
+
+        # Check that the ddo is registered
+        if not self._get_aquarius().ddo_exists(did):
```
calina-c, line 476: you just got this asset from aquarius, there's no point in checking again if it exists.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,185 @@ def create(
...
+            assert encrypted_files, "Encrypted_files is required for adding new tokens."
+
+        # Check if datatokens in delete_erc20_tokens are in the old_asset
+        for delete_erc20_tokens_data in delete_erc20_tokens:
+            assert delete_erc20_tokens_data in [
+                dt.get("address") for dt in old_asset.datatokens
+            ], f"{delete_erc20_tokens_data} to delete is not in the old asset datatokens."
+
+        # Metadata sanity check
+        assert isinstance(
```
calina-c, line 444: I am not sure about separating the parts for the new asset. i.e. checking metadata, files etc. separately? This is an approach that works, but I think Alex's suggestion would be better. I agree that we offer utillities for creating the asset but I think in the case of update we should just grab the changed asset and update it on chain, using all its new features.

Here is a use case:
we update the trusted algorithms on the Asset object, then call update. Do we have to do the trusted algorithms part all over again and put it in the metadata argument instead? Our object already contains the necessary changes.

The drawback on this is flexibility:
- for create, we offer the option to auto-deploy some tokens, which will be limited on update. But I would treat that as a "nothing changed" (same tokens as before).
- in case tokens HAVE changed, we check their addresses and do not offer autodeploy options. Ok, maybe we can work in some ErcCreateData, but after we get the basic flow working.

Bottom line:
We shouldn't update the asset piecemeal, we should use the Asset object.

This is my personal opinion. I will tag @trentmc @alexcos20 for a second opinion though.

calina-c (changes requested):

ocean_lib/ocean/ocean_assets.py
md0x: Yes, I guess it makes sense, I wanted to be sure the deletion of an datatoken from the NFT was aware. With the change you  proposed of passing the modified Asset it makes sense to me to get rid of this.
Thank you.

ocean_lib/ocean/ocean_assets.py
md0x: Sounds good to me, 
I am going to refactor the code to pass directly as a parameter to update the updated asset and thus make the update easier (remove for now the token deployment with erc20 data among other things).

ocean_lib/ocean/ocean_assets.py
```diff
@@ -141,24 +141,77 @@ def build_data_tokens_list(
         self, services: list, deployed_erc20_tokens: list
     ) -> list:
         data_tokens = []
+        # (1-n) service per datatoken, 1 datatoken per service
         for erc20_token in deployed_erc20_tokens:
             for service in services:
```
calina-c, line 145: let's make a list comprehension, I think the current structure warrants it. E.g. datatokens = datatokens + [{... that dict} for service in services if s.address == dt.address]

ocean_lib/ocean/ocean_assets.py
```diff
@@ -141,24 +141,77 @@ def build_data_tokens_list(
...
+                            "address": erc20_token.address,
+                            "name": erc20_token.contract.caller.name(),
+                            "symbol": erc20_token.symbol(),
+                            "serviceId": service.id,
+                        }
+                    )
         return data_tokens
 
+    @staticmethod
+    def process_ddo(
```
calina-c, line 159: I think a better name for this function would be _encrypt_ddo_locally, or _encrypt_locally_if_needed, or even _apply_ddo_flags. It is unclear otherwise what "processing" is happening, without looking at the code

ocean_lib/ocean/ocean_assets.py
```diff
@@ -141,24 +141,77 @@ def build_data_tokens_list(
...
+        compress_flag: Optional[bool] = False,
+    ):
+        # Process the DDO
+        asset_dict = asset.as_dictionary()
+        ddo_string = json.dumps(asset_dict, separators=(",", ":"))
+        ddo_bytes = ddo_string.encode("utf-8")
+        ddo_hash = create_checksum(ddo_string)
+
+        # Plain asset
+        if not encrypt_flag and not compress_flag:
```
calina-c, line 172: I also think we could use some "return" statements here, instead of doing a huge `elif` thing. This was previously an `elif` because it was part of a larger logic, but now as a function it *screams* "force returns". E.g. if the asset is plain you can return it directly...

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,88 @@ def create(
...
-        ddo_hash = create_checksum(ddo_string)
+        document, flags, ddo_hash = self.process_ddo(
+            asset, provider_uri, encrypt_flag, compress_flag
+        )
 
-        # Plain asset
-        if not encrypt_flag and not compress_flag:
-            flags = bytes([0])
-            document = ddo_bytes
+        _ = erc721_token.set_metadata(
```
calina-c, line 382: if you don't need the assignment, then don't assign it. We use the _, something assignment when we have a tuple return type, to get "half" of the return, but in this case you don't need to assign at all, if the only return value is unused

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,88 @@ def create(
...
+        """
+
+        # Metadata sanity check
+        assert isinstance(
+            asset.metadata, dict
+        ), f"Expected asset.metadata of type dict, got {type(asset.metadata)}"
+
+        asset_type = asset.metadata.get("type")
+
+        assert asset_type in (
```
calina-c, line 423: these asserts seem identical to what we have in create. How about a function called _assert_ddo_metadata to keep it DRY?

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +375,88 @@ def create(
...
+        if not validation_result:
+            msg = f"Asset has validation errors: {validation_errors}"
+            logger.error(msg)
+            raise ValueError(msg)
+
+        document, flags, ddo_hash = self.process_ddo(
+            asset, provider_uri, encrypt_flag, compress_flag
+        )
+
+        return erc721_token.set_metadata(
```
calina-c, line 459: I think we should return the asset (and wait for it) instead. What was the return type for v3?

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -38,43 +42,166 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+        publisher_ocean_instance.web3, config, publisher_wallet, publisher_wallet
+    )
+
+    ocean = publisher_ocean_instance
 
-    ddo = create_asset(ocn, alice, config)
+    data_provider = DataServiceProvider
+    file1_dict = {"type": "url", "url": "https://url.com/file1.csv", "method": "GET"}
+    file1 = FilesTypeFactory(file1_dict)
+    encrypt_response = data_provider.encrypt(
```
calina-c, line 58: there is an encrypt utility on ocean_assets itself. Maybe we can use that for some concision.

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -38,43 +42,166 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
-    log = ddo_reg.get_event_log(
-        ddo_reg.EVENT_METADATA_UPDATED, block - block_confirmations, asset_id, 30
+    ocean.assets.update(
+        asset=ddo,
+        publisher_wallet=publisher_wallet,
     )
-    assert log, "no ddo updated event"
-    _asset = wait_for_update(ocn, ddo.did, "name", _name)
+
+    _asset = wait_for_update(ocean, ddo.did, "description", _description)
```
calina-c, line 78: we made the decision of waiting for the creation inside create, so I think we should mirror that in update as well. From my POV, there's no point in separating the update and wait_for_update functions. But you can leave this comment last, and we can get some second opinions (applicable to create, too)

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -38,43 +42,166 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+
+    # Update credentials
+    new_asset = new_metadata = copy.deepcopy(_asset)
+    new_metadata = copy.deepcopy(_asset.metadata)
+    _description = "Updated credentials"
+    _new_credentials = {
+        "allow": [{"type": "address", "values": ["0x123", "0x456"]}],
+        "deny": [{"type": "address", "values": ["0x2222", "0x333"]}],
+    }
+    new_metadata["description"] = _description
```
calina-c, line 93: let's separate the tests a bit. If we are testing for metadata update above, please only check credentials update now. If we always test metadata + something, we might get false results. I want to see that editing the credentials without any metadata updates works properly. Same for the other cases.

calina-c (changes requested):
On my previous review, I focused on the architecture of the update function itself, which was a prerequisite of sorts. This time, I added a more comprehensive review of the form and components. This looks like it is progressing well! Good job, Pablo.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -374,10 +459,9 @@ def create(
             from_wallet=publisher_wallet,
         )
 
-        # Fetch the asset on chain
-        asset = self._get_aquarius(self._metadata_cache_uri).wait_for_asset(did)
+        self._web3.eth.wait_for_transaction_receipt(tx_result)
 
-        return asset
+        return wait_for_asset_update(self._get_aquarius(), asset, tx_result)
```
calina-c, line 464: I think you can assign an aquarius instance to avoid repetition of the get aquarius part here and above.

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -38,43 +41,145 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+
+    _asset = publisher_ocean_instance.assets.update(
+        asset=ddo,
+        publisher_wallet=publisher_wallet,
+    )
+
+    assert _asset.datatokens == ddo.datatokens
+    assert len(_asset.services) == len(ddo.services)
+    assert _asset.services[0].as_dictionary() == ddo.services[0].as_dictionary()
+    assert _asset.credentials == ddo.credentials
```
calina-c, line 64: ok let's assert that the updated values truly corresponds ("updated" and "description" keys in this case)

ocean_lib/ocean/test/test_ocean_assets.py
```diff
@@ -38,43 +41,145 @@ def test_register_asset(publisher_ocean_instance, publisher_wallet, consumer_wal
...
+
+
+def test_update_credentials(publisher_ocean_instance, publisher_wallet, config):
+
```
calina-c, line 68: you have an empty line at the beginning of each test. Please remove it or populate it with docstrings

tests/resources/ddo_helpers.py
```diff
@@ -297,7 +297,7 @@ def wait_for_update(ocean, did, updated_attr, value, timeout=30):
 
         if not ddo:
             time.sleep(0.2)
-        elif ddo.metadata["main"].get(updated_attr) == value:
```
calina-c, line 300: can we remove the entire wait_for_update function? is it not deprecated now?

calina-c (changes requested):
Excellent work! I left a few nitpicks but I'm confident we can merge this tomorrow.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -140,25 +137,96 @@ def find_service_by_data_token(self, data_token: str, services: list) -> str:
...
-                data_tokens.append(data_token)
-        return data_tokens
+                for service in services
+                if service.data_token == erc20_token.address
+            ]
+
+        return datatokens
+
+    @staticmethod
+    def _encrypt_ddo_locally(
```
DMats, line 157: Please consider renaming to `_encrypt_ddo`.  The `_locally` seems like a misnomer given that you're still making a call to the provider encrypt endpoint.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -322,48 +377,76 @@ def create(
...
-            # Compress DDO
-            compressed_document = lzma.compress(ddo_bytes)
+    def update(
+        self,
+        asset: Asset,
+        publisher_wallet: Wallet,
+        provider_uri: Optional[str] = None,
+        encrypt_flag: Optional[bool] = False,
+        compress_flag: Optional[bool] = False,
+    ) -> str:
```
DMats, line 407: This returns an `Optional[Asset]`, not a `str`

ocean_lib/ocean/util.py
```diff
@@ -97,3 +100,24 @@ def get_ocean_token_address(
         address_file, network or get_network_name(web3=web3)
     )
     return addresses.get("Ocean") if addresses else None
+
+
+@enforce_types
+def wait_for_asset_update(aquarius: Aquarius, asset: Asset, tx: str):
```
DMats, line 106: I think this function should be moved to `aquarius.py` and adapted to be a method of `Aquarius` class. I'd place it right under the existing `Aquarius.wait_for_asset` method.

ocean_lib/ocean/util.py
```diff
@@ -97,3 +100,24 @@ def get_ocean_token_address(
...
+    ddo = None
+    while True:
+        try:
+            ddo = aquarius.get_asset_ddo(asset.did)
+        except ValueError:
+            pass
+        if not ddo:
+            time.sleep(0.2)
+        elif ddo.event.get("tx") == tx:
+            print(ddo.event)
```
DMats, line 117: Please use a logger add a descriptive log message. Something like:

```
logger.debug(f"Transaction matching the given tx id detected in metadata store. asset.event = {ddo.event}"
```

DMats (changes requested):
Great work @md0x.  I left a few minor comments.

DMats (approved):

calina-c (approved):

PR #656, opened 2022-01-14
Test simultaneous deploy of NFT and ERC.

### LINKED ISSUE #644: Deploy NFT and datatoken in the same transaction
There is a function in smart contracts to do both from the same transaction. Please offer this option from ocean.py to save gas fees. We need to have both options: separate and simultaneous deploy.

calina-c:
Closes #644


calina-c:
As it turns out, this is already tested. However, a good thing came out of this: we discovered a bug in contracts which will soon be fixed.

PR #663, opened 2022-01-18
Implements user orders.

### LINKED ISSUE #642: Get user orders in v4
Currently in the Ocean class (ocean/ocean.py) there is a function called `get_user_orders`, which is outdated. We need to have a way of retrieving all user orders on a token. I discussed this with Trent and it is not high priority, so it's been taken out of #633.

TODOs for completion:
- adapt the `get_user_orders` function for v4
- add the get_start_order_logs (not sure if in ERC721Token or ERC20Token at the point I'm writing this) function, similar to what was previously in the DataToken class/contract.
- add a test in `ocean/test/test_ocean.py` to verify the behaviour of `get_user_orders`
- potentially use in the readmes as a way to assert that the order has been placed, in conjunction with the asserts for downloaded files (see marketplace_flow in v3)

calina-c:
Closes #642.


ocean_lib/models/side_staking.py
```diff
@@ -14,17 +14,20 @@ class SideStaking(ContractBase):
...
     def get_base_token_address(self, data_token: str) -> str:
-        return self.contract.caller.getBaseTokenAddress(data_token)
+        return self.contract.caller.getBasetokenAddress(data_token)
 
     def get_pool_address(self, data_token: str) -> str:
         return self.contract.caller.getPoolAddress(data_token)
 
     def get_base_token_balance(self, data_token: str) -> int:
-        return self.contract.caller.getBaseTokenBalance(data_token)
+        return self.contract.caller.getbaseTokenBalance(data_token)
```
calina-c, line 30: in the ABIs, is it getBasetokenAddress and getbaseTokenBalance? The casing is different...

mariacarmina: In barge it was getBasetokenBalance and in contracts it's `getbaseTokenBalane` https://github.com/oceanprotocol/contracts/blob/v4main_postaudit/contracts/pools/ssContracts/SideStaking.sol#L22

mariacarmina: Forgot to modify, it should match the ABI

mariacarmina (approved):
Great job, @calina-c !

PR #667, opened 2022-01-19
Replaced web3.toWei with utility to_wei.

### LINKED ISSUE #662: Replace web3.toWei with to_wei utility
It seems in some places we are using to_wei utility, while in others we use the web3 one. Replace web3.toWei with to_wei utility for consistency. See [comment](https://github.com/oceanprotocol/ocean.py/pull/659#discussion_r786496854)

mariacarmina:
Fixes #662  .

Changes proposed in this PR:

- replaced web3.toWei with `to_wei`;
- removed unused lines of code from tests.


ocean_lib/models/erc20_enterprise.py
```diff
@@ -14,14 +14,14 @@ class ERC20Enterprise(ERC20Token):
     CONTRACT_NAME = "ERC20TemplateEnterprise"
 
     def buy_from_fre_and_order(
-        self, order_params: dict, fre_params: dict, from_wallet: Wallet
+        self, order_params: tuple, fre_params: tuple, from_wallet: Wallet
```
calina-c, line 17: How is this change related to the task?

ocean_lib/models/erc20_enterprise.py
```diff
@@ -14,14 +14,14 @@ class ERC20Enterprise(ERC20Token):
...
     ) -> str:
         return self.send_transaction(
             "buyFromFreAndOrder", (order_params, fre_params), from_wallet
         )
 
     def buy_from_dispenser_and_order(
-        self, order_params: dict, dispenser_address: str, from_wallet: Wallet
+        self, order_params: tuple, dispenser_address: str, from_wallet: Wallet
```
calina-c, line 24: In cases where we have a tuple type we should be able to accept both regular tuples, dictionaries and our own variety of NamedTuples defined in models_structures.py. Web3.py supports both tuples and dictionaries and our tests are mixed in this aspect. If the web3.py PR for named tuples is approved, I will start refactoring on all these parameters to allow this. However, as long as the tests pass I don't want to handle this right now. We should stick to the toWei changes for now.

READMEs/marketplace-flow.md
```diff
@@ -192,14 +193,14 @@ erc20_token = ocean.get_datatoken(asset.get_service("access").datatoken)
 OCEAN_token = ocean.get_datatoken(ocean.OCEAN_address)
 
 ss_params = [
-    ocean.web3.toWei(1, "ether"),
```
calina-c, line 195: In this particular case, we want a to_wei function inside Ocean class, to avoid imports. The main advantage of the toWei was that we could call it directly from ocean. When we discussed this refactor, we also discussed placing a to_wei utility inside the Ocean class and using that where convenient. Please add that as discussed.

calina-c (changes requested):
Good job. However, the task is not 100% complete and there are some extra changes that don't belong here.

calina-c (approved):

PR #676, opened 2022-01-25
Added checks for consumable asset.

### LINKED ISSUE #673: Check if the asset is consumable for download
For download service, ocean.py doesn't handle error logs if the asset is not consumable for this type of service.

Steps:

- in download function, resolve the asset;
- check the `stats` of that DDO and throw a specific error if is not consumable.

mariacarmina:
Fixes #673  .

Changes proposed in this PR:

- add checks if the asset is consumable before calling provider download endpoint;
- add unit tests for this functionality.


calina-c (approved):
Approved but black has to pass too.

Issue #670, opened 2022-01-20
DataServiceProvider.download contains hardcoded provider URI


DMats:
This issue is to fix `DataServiceProvider.download` so that it doesn't use a hardcoded provider URI.

```
        fileinfo_response = DataServiceProvider.fileinfo(
            did, service_id, "http://172.15.0.4:8030/api/services/fileinfo"
        )
```

PR #677, opened 2022-01-25
Remove hardcoded URL from download function

mariacarmina:
Fixes # .

Changes proposed in this PR:

- remove hardcoded URL from download function

PR #678, opened 2022-01-25
Remove hardcoding in DSP with fileinfo.

calina-c:
Closees #670

### DIFF
--- a/ocean_lib/data_provider/data_service_provider.py
+++ b/ocean_lib/data_provider/data_service_provider.py
@@ -180,10 +180,11 @@ class DataServiceProvider:
         index: Optional[int] = None,
         userdata: Optional[Dict] = None,
     ) -> None:
-
+        _, fileinfo_endpoint = DataServiceProvider.build_fileinfo(service_endpoint)
         fileinfo_response = DataServiceProvider.fileinfo(
-            did, service_id, "http://172.15.0.4:8030/api/services/fileinfo"
+            did, service_id, fileinfo_endpoint
         )
+
         files = fileinfo_response.json()
         indexes = range(len(files))
         if index is not None:


ocean_lib/data_provider/data_service_provider.py
```diff
@@ -180,10 +180,11 @@ def download(
         index: Optional[int] = None,
         userdata: Optional[Dict] = None,
     ) -> None:
-
+        _, fileinfo_endpoint = DataServiceProvider.build_fileinfo(service_endpoint)
```
mariacarmina, line 183: The service_endpoint is the download one. When we call build_fileinfo, we need to pass the root uri of the provider. I suggest to use `get_root_uri` which retruns the provider uri of the service_endpoint => http://172.15.0.4:8030/ and after use the result to call `build_fileinfo`. Maybe #677 can be useful here.

mariacarmina: This is from asset_downloader file
image

calina-c: actually, this does open a can of worms: we don't need to be passing in the service endpoint. We can retrieve it from the service inside the download function. We NEVER need provider root uri. We can always infer the root and all related service_endpoints based on the service endpoint. This is applicable to all the functions inside the DSP.

Do you want to take this task or should I work on it tomorrow? I know you are low on tasks.

calina-c: Look at the snippet you paste it. We are calculating the download endpoint OUTSIDE the download function, then feeding it to the download function as a parameter. But the download function "knows" it's a download function and that it should build the download endpoint. So move that responsibility there. Please get rid of all service_endpoint and provider_root_uri parameters being sent to functions in the DataServiceProvider. We have the service, we can get that on the fly.

calina-c:
@Maria I've created a different ticket for the follow-up: https://github.com/oceanprotocol/ocean.py/issues/680

Issue #639, opened 2022-01-07
Use NamedTuples properly (formerly task for startMultipleTokenOrders)


AnaLoznianu:
Currently we have an issue with startMultipleTokenOrders function in ERC721Factory. The issue comes from the "tuple[]" encoding, which web3.py does not support correctly. This means we have to do manual encoding, and possibly send the transaction without web3.py.

We have the following options:
- figure out a way to send the transaction with encoded parameters using web3.py, but skipping ContractFunction logic
- if we are unable to do so, we must broadcast the transaction completely raw, using no tools/libraries, in which case we also have to handle the confirmation logic that the transaction took place (what web3.py handles out of the box)

### What we have until now:
- using a tuple instead of the previously-used dictionary
- we are converting providerData to bytes (it is originally a string, but we can't send it like that)
- encoding with the following signature

```
encd = encode_single("(address,address,uint256,address,address,uint256,uint8,bytes32,bytes32,bytes32)[]", [base_tuple])
```
- there's clues on how to send the transaction here: https://www.programcreek.com/python/example/116802/eth_abi.encode_abi

specifically, see
```
transaction = {
        "from": eth_tester.get_accounts()[0],
        "to": contract_address,
        "gas": 500000,
        "data": encode_hex(fn_selector + encode_abi(arg_types, fn_args)),
    }
```

calina-c:
Merged in two fixes. The library now allows for uniform passing of tuples, dicts and our specific NamedTuples. However, we handle an extra conversion before sending it to web3.py, conversion which will happen on web3.py once https://github.com/ethereum/web3.py/pull/2312 is merged in. Then we can erase the conversions.

PR #684, opened 2022-01-26
Uniform and flexible list, dict and named tuples

calina-c:
Work on #639.

Marked as draft until we get the right PoolData ordering in ERC721Factory (dep. contracts). The remaining task then is to remove the PoolData2 structure and its usages.


mariacarmina (approved):
Looks nice & clean!

PR #695, opened 2022-02-01
Issue694 remove compute status filtering

### LINKED ISSUE #694: Remove compute 'results' and 'status' filter from ocean.py
Currently `OceanCompute.status` and `OceanCompute.results` both call the `computeStatus` endpoint in provider.  On top of that, they apply to filter to the returned dict:

`OceanCompute.status` returns a dict containing only `ok`, `status` and `statusText`.
`OceanCompute.results` returns a dict containing only `did`, `urls`, and `logs`.

This issue is to 

1. Remove the filtering so that `OceanCompute` returns the full computeStatus response.
2. Refactor the OceanCompute methods so that there's only 1 for calling computeStatus endpoint.

DMats:
Fixes #694.

Changes proposed in this PR:

- Remove compute status filters
- Refactor `OceanCompute` `status()`, `result()`, and `result_file()`.
- Refactor `DataServiceProvider` `compute_job_status()`, `compute_job_result()`, and `compute_job_result_file()`

### DIFF
--- a/ocean_lib/ocean/ocean_compute.py
+++ b/ocean_lib/ocean/ocean_compute.py
@@ -34,17 +34,14 @@ class OceanCompute:
 
     @staticmethod
     @enforce_types
-    def _status_from_job_info(job_info: Dict[str, Any]) -> Dict[str, Any]:
+    def _add_ok_to_job_info(job_info: Dict[str, Any]) -> Dict[str, Any]:
         """
-        Helper function to extract the status dict with an added boolean for quick validation
-        :param job_info: dict having status and statusText keys
-        :return:
+        Helper method to add the "ok" key to the job info dict for quick validation.
+        :param job_info: dict returned by computeStatus endpoint in provider
+        :return: dict with the "ok" key added
         """
-        return {
-            "ok": job_info["status"] not in (31, 32),
-            "status": job_info["status"],
-            "statusText": job_info["statusText"],
-        }
+        job_info.update({"ok": job_info["status"] not in (31, 32)})
+        return job_info
 
     @enforce_types
     def start(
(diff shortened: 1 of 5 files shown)


ocean_lib/ocean/ocean_compute.py
```diff
@@ -94,38 +80,17 @@ def status(self, did: str, job_id: str, wallet: Wallet) -> Dict[str, Any]:
...
-        info_dict = self._data_provider.compute_job_result(
+        job_info = self._data_provider.compute_job_status(
             did, job_id, service_endpoint, wallet
         )
-        return {
-            "did": info_dict.get("resultsDid", ""),
-            "urls": info_dict.get("resultsUrl", []),
-            "logs": info_dict.get("algorithmLogUrl", []),
-        }
+        job_info.update({"ok": job_info["status"] not in (31, 32)})
```
calina-c, line 89: I would add status code checks and use job_info.get("status") syntax to avoid errors

calina-c (changes requested):

ocean_lib/ocean/ocean_compute.py
DMats: The status code check is performed inside `_send_compute_request` which is called by `compute_job_status`, so it's unnecessary and impossible to repeat here because at this point we only have the `dict`, not the `Response`.

For job_info.get("status") I'm fine with that, but I'll need to add `None` to the failure conditions.  `(31, 32, None)`

P.S. As a result of this comment, I realized that I forgot to refractor the `DataServiceProvider` `compute_job_result` and `compute_job_result_file` methods the same way I did for the `OceanCompute` `result` and `result_file`.  So thank you!

DMats: Done.  Please resolve.

calina-c (approved):

PR #704, opened 2022-02-03
Add a separate section in each README for using createNftErc type functions

### LINKED ISSUE #693: Add a separate section in each README for using createNftErc type functions
There are couple of functions in `erc721_factory` that can be added in READMEs:

- `create_nft_with_erc` for `datatokens.md`
- `create_nft_erc_with_pool` for `marketplace-flow.md`
- `create_nft_erc_with_fixed_rate` for `fixed-rate-exchange.md`
- `create_nft_erc_with_dispenser` for `dispenser-flow.md`

mariacarmina:
Fixes #693  .

Changes proposed in this PR:

- 
- 
-


ocean_lib/models/erc721_factory.py
```diff
@@ -124,12 +126,16 @@ def start_multiple_token_order(self, orders: list, from_wallet: Wallet) -> str:
...
-        nft_create_data: Union[dict, tuple, CreateERC721DataNoDeployer],
+        nft_create_data: Union[
+            dict, tuple, CreateERC721DataNoDeployer, CreateERC721Data
+        ],
         erc_create_data: Union[dict, tuple, CreateErc20Data],
         from_wallet: Wallet,
     ) -> str:
         # TODO: this will be handled in web3 py
-        if isinstance(nft_create_data, CreateERC721DataNoDeployer):
+        if isinstance(nft_create_data, CreateERC721DataNoDeployer) or isinstance(
```
calina-c, line 136: there is no argument for additional erc20 deployer in that ABI, we should not support regular CreateERC721Data, just CreateERC721DataNoDeployer. This applies everywhere where it was changed. I wouldn't have created a different named tuple if both would have the same ABI structure.

ocean_lib/models/erc721_factory.py
```diff
@@ -263,3 +281,31 @@ def get_token_address(self, tx_id: str):
         )
 
         return registered_event[0].args.newTokenAddress
+
+    def create_nft_erc_tokens_once(
```
calina-c, line 285: what is this? Is this a function that creates the nft and erc at the same time, except you are doing 2 transactions? This function already exists if you scroll up, with just one transaction, calling it directly from the contracts.

READMEs/datatokens-flow.md
```diff
@@ -123,4 +123,60 @@ print(f"datatoken name: {erc20_token.token_name()}")
 print(f"datatoken symbol: {erc20_token.symbol()}")
 ```
 
+## 3. Publish datatokens using one helper
```
calina-c, line 126: No point in adding a different section and the entire flow all over again. Just say something like "as an alternative to publishing the nft and the datatoken using separate transactions, you can use the ... function to deploy them within a single transaction ... ocean.py also offers the option of creating the NFT, the ERC20 and a pool, exchange or dispenser within the same transaction, using the PoolData, FixedData or DispenserData as arguments".

Then just show samples of the data involved. The user is not stupid, we don't need to copy the whole flow from start to end just to showcase a functionality almost identical to what we already showed them.

calina-c (changes requested):

ocean_lib/models/erc721_factory.py
mariacarmina: This function calls `create_nft_with_erc` and handles both events in order to retrieve the address for both token. We did the same for NFT creation and for datatoken as well.

calina-c:
Added the "do not merge" label since this is still under discussion.

trentmc (changes requested):
I have given my feedback in the github issue itself, here:
https://github.com/oceanprotocol/ocean.py/issues/693#issuecomment-1030577857

I am _not_ supportive of doing this. It kills DRY.

READMEs/datatokens-flow.md
```diff
@@ -141,4 +141,7 @@ print(f"datatoken name: {erc20_token.token_name()}")
 print(f"datatoken symbol: {erc20_token.symbol()}")
 ```
 
+As an alternative for publishing a NFT and a datatoken at once, you can use `create_nft_with_erc20`.
```
calina-c, line 144: I think you can move this to the very end. Give the user their "congrats" first 😅

calina-c: Also, make sure you mention that the arguments for NFT creation are a bit different for the "at once" functions, since the additional erc 20 deployer will automatically be set as the factory router (use the ...NoDeployer named tuple)

READMEs/dispenser-flow.md
```diff
@@ -44,12 +44,11 @@ Please refer to [datatokens-flow](datatokens-flow.md) and complete the following
...
 In the same python console:
 ```python
-from ocean_lib.models.dispenser import Dispenser
 from ocean_lib.models.models_structures import DispenserData
-from ocean_lib.ocean.util import get_address_of_type
+from ocean_lib.web3_internal.constants import ZERO_ADDRESS
 
 # Get the dispenser
-dispenser = Dispenser(ocean.web3, get_address_of_type(config, "Dispenser"))
+dispenser = ocean.dispenser
```
calina-c, line 51: 👍👍👍👍

READMEs/datatokens-flow.md
calina-c: This applies to all "at once" functions btw

tests/readmes/test_dispenser_flow.py
```diff
@@ -5,11 +5,13 @@
 import os
 
 from ocean_lib.example_config import ExampleConfig
-from ocean_lib.models.dispenser import Dispenser
-from ocean_lib.models.models_structures import CreateErc20Data, DispenserData
+from ocean_lib.models.models_structures import (
+    CreateErc20Data,
+    DispenserData,
+    CreateERC721DataNoDeployer,
```
calina-c, line 11: Are the nodeployer imports used in readmes? I thought the flow remains the simple one, and the "at once" functions are only showcased as an alternative

calina-c (changes requested):

calina-c:
@mariacarmina tests are failing, can you please look into it?

calina-c (approved):

trentmc (approved):

PR #722, opened 2022-02-16
Coverage

calina-c:
Works on https://github.com/oceanprotocol/multi-repo-issue/issues/134

I also moved and renamed some things:
-> models_structures are actually helpers for ABIs which we offer, so I renamed the file abi_tuples.py and moved it to a new structures folder
-> algorithm metadata -> moved to structures folder
-> file objects are also helpers that we use to help out users, so I moved it to the structures folder
-> I remade AddressCredential into a parent class instead of injecting the asset inside it. Later on we can use dynamic inheritance or other forms to accomodate more credential type, but my goal was to remove separate and clumsy usages of the AddressCredential and use the asset object directly

Why did I move stuff to `structures` when it could have been fine in models and agreements? Simply because these things are different.

- Models should only contain contract models and constants that we use as part of dealing with the contracts.
- Agreements should only contain constants, types and agreeements that we use when "talking" to other Ocean Protocol components. i.e. they are "agreed" with other components.
- anything related to helpers that we use internally to build something (like the ABI helpers, file structure generators, algorithm metadata builders) should reside in structures.

Coverage improvements made:
- address credential testing
- removed obsolete functions from Aquarius (left from v3), add a few Aquarius class tests
- test failures of DataServiceProvider for both empty response and non-200 status codes, a few improvements on the mocks we use for this purpose
- add tests for file objects
- add a small test for example config and others


DMats (commented):
Part of me wishes we used composition instead of inheritance - where the `AddressCredential` was a member of the `Asset`, created in the `init` method.  

But aside from that this is great work.  Nice job @calina-c

Issue #734, opened 2022-02-22
Add a slow pytest marker and configure test to run from time to time


AnaLoznianu:
- [ ] decimal point fuzzing 
- [ ] stress testing

calina-c:
The marker works and the actions starts, but it is timed out. I think we should wait for the stress test refactoring branch to be merged and then recheck the action.

PR #737, opened 2022-02-23
Issue705 image processing c2d example

### LINKED ISSUE #705: Update image processing c2d example to v4, and bring into ocean.py repo
Currently, the c2d readme mentions two external resources for learning to use other algorithms with Compute-to-data.

```markdown
TODO: Update these examples based on v4
1. simple image processing at [ocean-lena](https://github.com/calina-c/ocean-lena/blob/main/c2d-flow.md)
2. logistic regression for classification [blog post](https://medium.com/ravenprotocol/machine-learning-series-using-logistic-regression-for-classification-in-oceans-compute-to-data-18df49b6b165) with both GUI and CLI flows.
```

This issue is to either:

1. Update these examples to v4 and bring them into the ocean.py repo as new READMEs
2. Remove them

DMats:
Fixes #705

Changes proposed in this PR:

- Add `READMEs/c2d-flow-more-examples.md` which includes image processing and linear regression c2d examples
- Parameterize `test_c2d_flow_readme` integration test to include image processing and linear regression examples

TODO: Update dataset and algorithm URLs to `main` after https://github.com/oceanprotocol/cdt-algorithms/pull/1 is merged.


tests/resources/algorithms/logistic_regression.py
```diff
@@ -0,0 +1,118 @@
...
+# Modified by Ocean Protocol Foundation
+# License: BSD 3 clause
+
+import json
+import os
+import pickle
+import sys
+
+# Uncomment if running locally. g not included in oceanprotocol/algo_dockers:python-panda image
+# import matplotlib.pyplot as plt
```
DMats, line 29: I chose not to update the [python-panda docker image requirements.txt] (https://github.com/oceanprotocol/algo_dockers/blob/main/python-panda/requirements.txt) to include `matplotlib` because I don't know if I have the authority to push a new image to docker hub.

DMats:
Just realized that this PR is still missing a few needed edits, so moving it back to Draft status for now.

tests/resources/algorithms/logistic_regression.py
DMats: PR for adding `matplotlib` to `python-panda` docker image was created.  Once merged, these lines can be uncommented. https://github.com/oceanprotocol/algo_dockers/pull/4

setup.py
```diff
@@ -50,6 +50,7 @@
     "pylint",
     "pytest",
     "pytest-watch",
+    "Pillow",
```
calina-c, line 53: I'm not sure we should support Pillow just for one example. I'd rather add `pip install Pillow` just in the relevant flow (in c2d examples, Lena)

tests/readmes/test_c2d_flow.py
```diff
@@ -21,7 +23,35 @@
 
 
 @pytest.mark.integration
-def test_c2d_flow_readme():
+@pytest.mark.parametrize(
+    "dataset_name,dataset_url,algorithm_name,algorithm_url,algorithm_docker_tag",
```
calina-c, line 27: I think just one sample can be tested. The others maybe as part of the "slow" marker (see #734)

tests/resources/algorithms/gpr.py
```diff
@@ -0,0 +1,103 @@
+#
+# Copyright 2022 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+import json
```
calina-c, line 5: I would much rather have the algos in different repos. Or a "resources/examples" repo. That way we don't inject unneeded dependencies into ocean.py, and it is clearer as a user flow that these are independent. They are independently-published algorithms, not part of ocean.py.

calina-c (commented):
I (very much) agree with having the multiple examples integrated, but not files (algos and datasets) related to those examples. I think it is clearer if they stay in a separate repo or repos. I've left comments with more explanations of my POV

tests/readmes/test_c2d_flow.py
DMats: Thanks for reviewing.  Respectfully, I disagree. The primary purpose of this test is to ensure the c2d READMEs work as intended. Therefore, they should mirror the README as closely as possible.

setup.py
DMats: `Pillow` is required by the lena example of `test_c2d_flow_readme`, the integration test that mirrors the c2d examples README. I think it's necessary to adequately test the lena example. I made it a test dependency, so if our PyPi deploy scripts are configured correctly, I don't think normal users will get it when installing ocean.py.

tests/resources/algorithms/gpr.py
DMats: Okay. I'll add them to https://github.com/oceanprotocol/cdt-algorithms

DMats: It appears I don't have permission to submit PRs to https://github.com/oceanprotocol/cdt-algorithms, I think, because it's not part of the Engineering group. https://github.com/orgs/oceanprotocol/teams/engineering.  I've sent a message to the team via Slack to try and get it added.

setup.py
calina-c: I'm just not very happy about adding an image-processing dependency to test an example of an image-processing c2d. It seems out of place and overkill. Let's ask for a deciding vote. @trentmc what do you think?

tests/readmes/test_c2d_flow.py
calina-c: yep :) exactly. Looks good.

calina-c (commented):
I gave this another look. It's shaping up nice but I asked Trent for an opinion regarding the Pillow dependency.

DMats:
Thanks for reviewing, @calina-c!

@trentmc I'm curious to know your thoughts on adding `Pillow` as a test dependency as well.  

I personally think that instantiating an `Image` object using `Pillow` makes the tests more complete and mirror the README more fully. And to do that, we need Pillow as a test dependency.

trentmc:
Hi @DMats and @calina-c , you both asked for my opinion.

It seems both of you agree to have the algorithms & datasets in a different repo (https://github.com/oceanprotocol/c2d-examples). OK. That's what we did circa oceanv3.

So the only question is: should ocean.py have dependency on pillow?

Pros: helps testing. 

Cons: another dependency. 

Some datapoints:
- doing `pip install -r requirements_dev.txt` on a fresh ocean.py (v4 branch) takes 36 s
- doing a `pip install pillow` in a fresh directory takes <0.5 s.

It's *really* useful to have automated tests that mirror our READMEs. I'm glad that we have been doing that, generally. It helps maintainability a lot. Having pillow as a dependency ensures automated testing for this README.

Analysis: testing helps _a lot_, and the extra dependency isn't that heavy, relatively speaking.

Therefore: I recommend to include pillow as a dependency.

calina-c (approved):

PR #740, opened 2022-03-02
Remove named tuples

### LINKED ISSUE #725: V4 DX: common methods have clumsy argument structure
**Background**

Here's an example from the [V4 datatokens README](https://github.com/oceanprotocol/ocean.py/blob/v4main/READMEs/datatokens-flow.md):
```python
# Create ERC20 token related to the above NFT.
erc20_data = CreateErc20Data(
    template_index=1, # default value
    strings=["ERC20DT1", "ERC20DT1Symbol"], # name & symbol for ERC20 token
    addresses=[
        alice_wallet.address, # minter address
        alice_wallet.address, # fee manager for this ERC20 token
        alice_wallet.address, # publishing Market Address
        ZERO_ADDRESS, # publishing Market Fee Token
    ],
    uints=[cap, 0],
    bytess=[b""]
)
nft_factory = ocean.get_nft_factory()
erc20_token = erc721_nft.create_datatoken(erc20_data=erc20_data, from_wallet=alice_wallet)
```

Compare this to [V3 datatokens README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/datatokens-flow.md):
```python
```

**Solution**

Make it support the following. Note how it doesn't fill in all the values; for ones not given it uses defaults. 

```python
erc20_token = erc721_nft.create_datatoken(
  name="DT1", symbol="DT1", cap=ocean.to_wei(100), from_wallet=alice_wallet)
```

Where to do this: 
- Any method that uses lists of parameters (because EVM workaround). Includes: any methods like this in the READMEs, including the "with_erc20" versions like `create_nft_with_erc20()`. 
- Note: ocean.js already does this! [Example for create erc20](https://github.com/oceanprotocol/ocean.js/blob/v4main/src/tokens/NFT.ts#L111)

----

NOTE: Brownie won't work. Brownie does a 1:1 mapping from smart contract interface. But the smart contract interface _itself_ is terrible, and can't be fixed because EVM only allows 4-5 arguments.

mariacarmina:
Fixes #725  .

Changes proposed in this PR:

- remove `CreateERC721Data`, `CreateERC721DataNoDeployer`, `CreateERC20Data`, `PoolData`, `FixedData`, `DispenserData`, `BPoolData`, `BPoolInitialized`, `ProviderFees`, `ConsumerFees`, `OrderParams`;
- fixed tests;
- update READMEs;
- delete unnecessary functions;
- created `.codeclimate.yml` file for check customization;


calina-c (commented):
Ok, so I read through a lot of this and here are my thoughts:
- it is a good first step towards what Trent has proposed, but not enough. Right now, all we have is a 1:1 mapping of what we previously had as tuples, but instead it's function arguments.

What we need to improve if we're going towards this route:
- default parameters: this is a must. If we are hiding complexity, but still asking users to declare things like ZERO ADDRESS, we are not really helping them that much.
- if we are not going for 1:1 mapping with the ABIs, then maybe we can help users with some better namings. E.g. do we ever use "strings" as anything other than name and symbol?

e.g.: `strings=["ERC20DT1", "ERC20DT1Symbol"]`

If so, can we have name and symbol parameters instead? What is "bytess"? What is "uints"? Can we have better parameter naming? Since they are passed "anonymously" (just the values, but no namings) to the contracts, should we keep their names just like they are in the contracts repo?

tagging @trentmc for an opinion.

mariacarmina:
> Ok, so I read through a lot of this and here are my thoughts:
> 
> * it is a good first step towards what Trent has proposed, but not enough. Right now, all we have is a 1:1 mapping of what we previously had as tuples, but instead it's function arguments.
> 
> What we need to improve if we're going towards this route:
> 
> * default parameters: this is a must. If we are hiding complexity, but still asking users to declare things like ZERO ADDRESS, we are not really helping them that much.
> * if we are not going for 1:1 mapping with the ABIs, then maybe we can help users with some better namings. E.g. do we ever use "strings" as anything other than name and symbol?
> 
> e.g.: `strings=["ERC20DT1", "ERC20DT1Symbol"]`
> 
> If so, can we have name and symbol parameters instead? What is "bytess"? What is "uints"? Can we have better parameter naming? Since they are passed "anonymously" (just the values, but no namings) to the contracts, should we keep their names just like they are in the contracts repo?
> 
> tagging @trentmc for an opinion.

I personally agree with this idea to have more descriptive params.

alexcos20:
Agreed

mariacarmina:
The only named tuples left (`Stakes`, `Operations`, `MetadataProof`) are used for `buy_dt_batch` and `stake_batch` where is needed to have a list of "structures". The user can use list of different "compressed structures" such as: `dict`, `tuple` and a specific named tuple. I suggest to keep only those for storing the information easier.

calina-c:
@mariacarmina can you please solve the conflicts?

calina-c (approved):

PR #742, opened 2022-03-04
Issue 741 - SideStaking vesting

### LINKED ISSUE #741: SideStaking vesting needs more testing

If using a pool with 1-sided staking (SideStaking.sol), then there may be vesting. E.g. 10% goes to publisher over the span of one year.

For oceanprotocol/ocean.py: it should have tests for this, for two reasons (a) to verify that it works (b) to show usage.  So far the only tests it has are at time=0, not time >0.

The js tests in oceanprotocol/contracts don't have tests at time>0 either. 

Which means that we have not tested vesting at all (EEK). Vesting working is critical, because if it has bugs then it will really badly mess up publishers' ability to monetize. (And anything we ship, we should test well, of course!) So right now I am worried about this.

We need to test this, not just a bit, but a *lot*, we really need to try to break it.

Have tests at time > 0:
- _during_ vesting 
- after vesting should have completed

And try to break it! Find edge cases, ways that it might go wrong.

Where: ocean.py has two related files: test_side_staking.py and test_vesting.py. One or both could be modified.

Side note: each of those files has just one test, and it's a big long monolith. It will be much more maintainable to modularize it into several smaller tests.

**Additional context 1**

I got really worried about this, so I set up some tests in tokenspice, modifying Trang's code. It passes, that's a sigh of relief. However my tests were still pretty basic, and weren't in ocean.py. We need thorough tests in ocean.py. That's the aim of this issue:)

Example of tests in tokenspice: [`issue29-redux`](https://github.com/tokenspice/tokenspice/tree/issue29-redux) branch, with [this PR](https://github.com/tokenspice/tokenspice/pull/155). Example test:

```console
pytest sol080/contracts/oceanv4/test/test_side_staking.py::test_sideStaking_properties
```

**Additional context 2**

As of Mar 4, @alexcos20 plans to add tests to oceanprotocol/contracts too. Consider that complementary to this.

trentmc:
Fixes #741.

### DIFF
--- a/tests/flows/test_vesting.py
+++ b/tests/flows/test_vesting.py
@@ -15,7 +15,7 @@ from ocean_lib.web3_internal.constants import ZERO_ADDRESS
 from ocean_lib.web3_internal.currency import to_wei
 from tests.resources.helper_functions import get_address_of_type
 
-
+ 
 @pytest.mark.unit
 def test_main(
     web3,


trentmc:
Closing this, someone else will work on this issue

PR #752, opened 2022-03-16
Testing NFT transfer

### LINKED ISSUE #745: Do we thoroughly test when an NFT is transferred? If no, test

V4 introduces erc721 data nfts. I worry that we haven't tested them thoroughly yet. For example :
- When ownership of an nft is transferred to someone else, are all permissions properly updated in erc20s?
- Is the original publisher still on the hook to host the data? If yes, do they still get 1.0 erc20 tokens when hosting it? If no, how does the new owner good host it, and Do they get 1.0 erc20 tokens for hosting it?

- Test nft transfer and related permissions, in happy path, with nonzero values. Including wrt what happens with 3rd party providers: {what if the old owner keeps serving up the data? What if the new owner is to serve up the data?}. What about fixed-rate exchange? What about free / dispenser? Does ERC20 vesting go to the new owner? Does it handle zero ERC20s for the NFT? One? Two?
- Test for unhappy paths. Truly try to break it! Eg  can an original publisher just transfer the nft and disappear, where it appeared that the new nft owner could host, but somehow they're not

mariacarmina:
Fixes #745  .

Changes proposed in this PR:

- added 2 large tests for transfer NFTs: one is related to transferring the NFT before deploying an ERC20, a pool and a FRE and one after;
- checked roles & ownership as well in both cases.

### DIFF
--- a/ocean_lib/models/test/test_erc721_token.py
+++ b/ocean_lib/models/test/test_erc721_token.py
@@ -5,9 +5,14 @@
 import pytest
 from web3 import exceptions
 
-from ocean_lib.models.erc20_token import ERC20Token
+from ocean_lib.models.bpool import BPool
+from ocean_lib.models.erc20_token import ERC20Token, MockERC20
 from ocean_lib.models.erc721_factory import ERC721FactoryContract
 from ocean_lib.models.erc721_nft import ERC721NFT, ERC721Permissions
+from ocean_lib.models.fixed_rate_exchange import (
+    FixedRateExchange,
+    FixedRateExchangeDetails,
+)
 from ocean_lib.web3_internal.constants import BLOB, ZERO_ADDRESS
 from ocean_lib.web3_internal.currency import to_wei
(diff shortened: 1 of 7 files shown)


calina-c (approved):

ocean_lib/models/test/test_erc721_token.py
```diff
@@ -641,3 +644,531 @@ def test_fail_transfer_function(web3, config, publisher_wallet, consumer_wallet)
         == "execution reverted: VM Exception while processing transaction: revert ERC721: transfer caller is not "
         "owner nor approved"
     )
+
+
+def test_transfer_nft(web3, config, publisher_wallet, consumer_wallet, factory_router):
```
trentmc, line 649: This is too long. Break it into 3+ test methods please:)

mariacarmina: I will refactor in #758 and specify the modifications there. Thank you!

PR #754, opened 2022-03-17
Issue744 test v4 fees

### LINKED ISSUE #744: Are v4 fees thoroughly tested? If no, do it

V4 fees are considerably more complex than v3.
 
- so that marketplaces etc have more opportunities to earn, and more flexibility in setting fees. 
- Plus there can now be third party providers,

This is important to get right so that the community can monetize - we need to make sure that we have tested in thoroughly. It can and should be tested in contracts; but since these also affect data science flows, we should also have thorough testing in ocean.py.

I'm pretty sure that the fees are not well tested yet, especially not in ocean.py. Why: there was churn in fees in the last few weeks, yet I saw no signs of any updates to ocean.py to reflect this. And from what I gather, most of the tests so far have mostly zero fees.

- Test all fees, in happy path, with nonzero values. Including for 3rd party providers. 
- Test for extremes of fees
- Test for unhappy paths. Truly try to break it! Eg is there some combination of fees where someone can make an (unfair) killing? Eg are there ways that a malicious marketplace can essentially steal from people?

DMats:
Fixes #744.

Changes proposed in this PR:

- 
- 
-


DMats:
It was decided at the backend refinement today that the v4 fees tests should not be based on #759 so I am closing this PR and starting over on a new branch from `v4main`.

PR #758, opened 2022-03-21
Added vesting tests

### LINKED ISSUE #741: SideStaking vesting needs more testing

mariacarmina:
Fixes #741  .

Changes proposed in this PR:

- added vesting tests;
- refactor NFT transfer tests;
- fix FRE creation test.


tests/flows/test_vesting.py
```diff
@@ -171,3 +172,143 @@ def test_main(
...
     )
     assert side_staking.get_vesting_amount(erc20_token.address) == vesting_amount
+
+    # check if vesting is correct
+    available_vesting_before = side_staking.get_available_vesting(erc20_token.address)
+
+    # advance 100 blocks to see if available vesting increased
+    for _ in range(100):
+        # send 1 wei
+        send_ether(consumer_wallet, ZERO_ADDRESS, 1)
```
calina-c, line 182: do we have to send ether? can we send ocean instead? I think it is easier to mint fake ocean than fake ether (but I might be wrong)

calina-c: I would also update the comment to say something like "send dummy transactions to increase block count (not part of the actual functionality)"

tests/flows/test_vesting.py
```diff
@@ -171,3 +172,143 @@ def test_main(
...
+        SideStaking.EVENT_VESTING,
+        tx_result.blockNumber,
+        web3.eth.block_number,
+        None,
+    )
+    assert vesting_event[0].args.datatokenAddress == erc20_token.address
+    assert vesting_event[0].args.publisherAddress == erc20_token.get_payment_collector()
+    assert vesting_event[0].args.caller == consumer_wallet.address
+
+    # 101 blocks passed, vesting started block = tx_receipt.blockNumber (deploy pool tx)
```
calina-c, line 200: let's make it 100 blocks for easier calcs (meaning the `range` above can be modified).
I'd also assert the vesting amount at each loop/step, not only at the end.

tests/flows/test_vesting.py
```diff
@@ -171,3 +172,143 @@ def test_main(
...
+    assert vesting_event[0].args.amountVested == int(
```
calina-c, line 201: can we also check any wallet balances / withdrawal, for safe measure?

tests/flows/test_vesting.py
```diff
@@ -171,3 +172,143 @@ def test_main(
...
+        == vested_blocks + tx_receipt.blockNumber
+    )
+    assert vesting_created_event[0].args.totalVestingAmount == vesting_amount
+
+    # check if vesting is correct
+    available_vesting_before = side_staking.get_available_vesting(erc20_token.address)
+    assert available_vesting_before == 0
+
+    # advance checkpoint_block blocks to see if available vesting increased
+    for checkpoint_block in checkpoint_blocks:
```
calina-c, line 289: I would suggest a single loop through all the blocks, and the following logic:
- send ether to increase block count
- if block is not a checkpoint block, continue
- (we are implicitly in the case where the block is a checkpoint one, so in the loop), do the assert.

How long do these tests take, btw? If it's not a very expensive test, time-wise, then there is no point in marking it as slow. Can you please benchmark it first, then let's decide whether it is `slow`-worthy or not :)

calina-c (changes requested):
A few cosmetic changes requested, but otherwise this is an excellent job!

mariacarmina:
For vesting progress test - tried with `0.01`, because it takes 30 seconds to run.
Screenshot from 2022-03-22 14-20-40

tests/flows/test_vesting.py
calina-c: as discussed during the stand-up, please address this as well. I referred to it as a withdrawal, but "claim" is indeed a better word.

tests/flows/test_vesting.py
```diff
@@ -171,3 +172,150 @@ def test_main(
...
+    # check if vesting is correct
+    dt_balance_before = erc20_token.balanceOf(erc20_token.get_payment_collector())
+    available_vesting_before = side_staking.get_available_vesting(erc20_token.address)
+    assert available_vesting_before == 0
+
+    # advance checkpoint_block blocks to see if available vesting increased
+    for block in range(checkpoint_blocks[-1]):
+        # send dummy transactions to increase block count (not part of the actual functionality)
+        send_ether(publisher_wallet, ZERO_ADDRESS, 1)
+        if block in checkpoint_blocks:
```
calina-c, line 297: yes, but as I said previously, do not add a layer of imbrication.
if the block is not a checkpoint block, `continue`
then go ahead with the checkpoint block logic (lines 298-321, but unindented by 1 measure)

mariacarmina: Ok, sure. I will change the condition logic. Also I will add the functionality for claiming the datatokens as well.

trentmc:
> [from github issue] And try to break it! Find edge cases, ways that it might go wrong.

Do you feel like you've added tests to truly try to break it? So far what I'm reading in the tests mostly seems to be happy path.

mariacarmina:
> > [from github issue] And try to break it! Find edge cases, ways that it might go wrong.
> 
> Do you feel like you've added tests to truly try to break it? So far what I'm reading in the tests mostly seems to be happy path.

I'm also planning to claim the DTs and test if the functionality works properly. I have discussed regarding my results (my assertion) and these are right and the formulas seem to be right. I am investigating to add more test cases regarding vesting.

mariacarmina:
> > For vesting progress test - tried with `0.01`, because it takes 30 seconds to run. ![Screenshot from
> 
> You should change it so that it doesn't take 30 s to run. It should take max 0.5 s. E.g. make the blockchain jump ahead _many_ blocks. Or if needed, just have it vest in 5 blocks.
Ok, I will decrease the percentages then.

calina-c (approved):

ocean_lib/models/side_staking.py
```diff
@@ -45,26 +58,6 @@ def get_vesting_last_block(self, datatoken: str) -> int:
...
-        self,
-        datatoken: str,
-        dt_amount_in: int,
-        pool_amount_out: int,
-        from_wallet: Wallet,
-    ) -> str:
-        return self.send_transaction(
-            "UnStake", (datatoken, dt_amount_in, pool_amount_out), from_wallet
-        )
-
```
trentmc, line 67: I see why these methods were removed -- they're more of an interface for the Pool contract, and didn't need to be exposed.

However why were they deleted here? It's out of scope of this issue. It would have made more sense to have a separate issue to do that. Keep the scope of each issue - and related PR - tight and well-defined.

That said: rather than adding work, this time just keep it part of this PR. In the future, please keep scope tight.

ocean_lib/models/test/test_erc721_token.py
```diff
@@ -790,8 +791,47 @@ def test_transfer_nft(web3, config, publisher_wallet, consumer_wallet, factory_r
     assert publish_fees[1] == base_token.address
     assert publish_fees[2] == to_wei(1)
 
+
+def test_nft_transfer_with_pool(
+    web3, config, publisher_wallet, consumer_wallet, factory_router
+):
+    """Tests transferring the NFT before deploying an ERC20, a pool."""
+    erc721_nft, erc20_token = deploy_erc721_erc20(
```
trentmc, line 799: 'deploy_erc721_erc20()` is called 14 times in this file, typically at the very beginning of a test function. This is a big DRY violation. Rather than asking you to fix it here, I just created https://github.com/oceanprotocol/ocean.py/issues/765.

ocean_lib/models/test/test_side_staking.py
```diff
@@ -17,6 +17,19 @@
...
+
+    assert side_staking.event_Vesting.abi["name"] == SideStaking.EVENT_VESTING
+    assert (
+        side_staking.event_VestingCreated.abi["name"]
+        == SideStaking.EVENT_VESTING_CREATED
+    )
+
+
 @pytest.mark.unit
 def test_side_staking(
```
trentmc, line 34: This method is now 362 lines long. This is _way_ too long. A good method - testing or implementation - should be max one page long (about 45 lines). Otherwise it's near impossible to inspect, review, or maintain.

This code needs to be refactored as such before I can even review it. 

Be sure to use conftest.py to re-use code and avoid DRY violations.

trentmc (changes requested):

ocean_lib/models/side_staking.py
mariacarmina: I have created #766 for this particular issue. I will revert the changes here.

ocean_lib/models/test/test_side_staking.py
mariacarmina: It was treated different. I focused just on the vesting part which is [here](https://github.com/oceanprotocol/ocean.py/pull/758/files#diff-08860fbdc12cc1930e80f6143c04de0f728006f63cd4a0e12fdd7e97cf3c79b5). @md0x has a PR (#764 ) related to SS. In models we have unit tests for the contract functions themselves and we created a special folder for flows, where we added the functional tests and that's why I implemented the logic in `test_vesting.py`.

calina-c:
I suggest keeping this PR focused on the vesting tests. Refactoring SS tests should be done in that specific PR. And seeing how #766 was added (and also done, in the meantime), #765 as well, then from my point of view we can merge this after fixing the conflicts.

what do you think, is it feasible to close this today? @trentmc @mariacarmina

trentmc:
> I suggest keeping this PR focused on the vesting tests. Refactoring SS tests should be done in that specific PR. And seeing how #766 was added (and also done, in the meantime), #765 as well, then from my point of view we can merge this after fixing the conflicts.
> 
> what do you think, is it feasible to close this today? @trentmc @mariacarmina

Yes, the scope of a PR on testing vesting should be only for testing vesting.

`test_vesting.py` has exactly one method, that is 280 lines long (about 7 pages worth). This is near impossible to review, and un-maintainable. 

Once `test_vesting.py` is refactored into smaller tests, I am happy to re-review. 

I will be expecting to see not only happy-path tests, but tests that are truly trying to _break_ things.

tests/flows/test_vesting.py
```diff
@@ -106,22 +139,127 @@ def test_main(
...
     # Tests consumer deploys pool and check market fee
-    initial_ocean_liq = to_wei("0.02")
+    mint_fake_OCEAN(config)
+    initial_ocean_liq = to_wei(200)
     ocean_contract = ERC20Token(web3=web3, address=get_address_of_type(config, "Ocean"))
-    ocean_contract.approve(
-        get_address_of_type(config, "Router"), to_wei("0.02"), consumer_wallet
+    ocean_contract.approve(factory_router.address, initial_ocean_liq, consumer_wallet)
+
+    erc20_token.deploy_pool(
```
trentmc, line 170: There are 7 calls to `erc20_token.deploy_pool()`, each mostly with the same parameters. Each call takes 17 lines, for 119 lines total. This is a DRY violation. Better: create a new function at the bottom of test_vesting.py called `_deploy_pool()` that takes in a couple parameters, hardcodes the rest, and calls `erc20_token.deploy_pool()` there. Then, then replace all the calls to `erc20_token.deploy_pool()` with the new `_deploy_pool()` method.

trentmc: And, there are 10 calls to `erc721_factory.deploy_erc721_contract()`. DRY violation. Fix it similarly, please.

tests/flows/test_vesting.py
```diff
@@ -147,32 +285,462 @@ def test_main(
...
+        # (it mints another block that change the formula)
+        counter += 1
+        vested_amount = side_staking.get_amount_vested_from_event(
+            tx_hash=tx_hash, erc20_token=erc20_token, from_wallet=publisher_wallet
+        )
+        assert (
+            vested_amount
+            == int(((block + 1 + counter) * vesting_amount) / vested_blocks)
+            - amount_vested_so_far
+        )
```
trentmc, line 425: This is super-hard to read. Please reformat so that it's readable. Eg. `target_vesting_amount = int(((block...`, then `assert vested_amount == target_vesting_amount`.

tests/flows/test_vesting.py
```diff
@@ -147,32 +285,462 @@ def test_main(
...
+        web3.eth.block_number,
+        None,
+    )
+    bpool = BPool(web3, pool_event[0].args.poolAddress)
+
+    # check if the vesting amount is correct
+    assert side_staking.get_vesting_amount(erc20_token.address) == vesting_amount
+
+    # check if vesting is correct
+    dt_balance_before = erc20_token.balanceOf(erc20_token.get_payment_collector())
```
trentmc, line 489: This method has several calls to `erc20_token.get_payment_collector()`, including inside larger expressions that black then forces across multiple lines, which in turn hurts readability. Recommend: create new line `pmt_collector = erc20_token.get_payment_collector()`, and use it throughout the function.

trentmc (changes requested):
It was hard to review "for real" because so many DRY violations and hard-to-read formatting. Please lmk you've made the suggested fixes, and I will re-review.

ocean_lib/models/side_staking.py
```diff
@@ -78,3 +93,47 @@ def set_pool_swap_fee(
         return self.send_transaction(
             "setPoolSwapFee", (datatoken, pool_address, lp_swap_fee_amount), from_wallet
         )
+
+    def check_vesting_created_event(
```
mariacarmina, line 97: @trentmc what do you think about those functions that I use in `test_vesting.py`. Should I add them both as well at the bottom of `test_vesting.py` or keep them here? There were a lot of lines of code for retrieving vesting event.

tests/flows/test_vesting.py
```diff
@@ -72,11 +51,29 @@ def test_main(
 
     permissions = erc721_nft.get_permissions(consumer_wallet.address)
 
-    assert permissions[1] is True
-    assert permissions[2] is True
-    assert permissions[3] is True
+    assert permissions[ERC721Permissions.MANAGER] is False
```
trentmc, line 54: Change to 
```assert not permissions[ERC721Permissions.MANAGER]```

You don't need "is False", it's redundant.

File: tests/flows/test_vesting.py [hunk 14]
```diff
@@ -72,11 +51,29 @@ def test_main(
...
+    assert permissions[ERC721Permissions.DEPLOY_ERC20] is True
```
trentmc, line 55: Change to
```assert permissions[ERC721Permissions.DEPLOY_ERC20]```

You don't need "is True", it's redundant.

File: tests/flows/test_vesting.py [hunk 15]
```diff
@@ -72,11 +51,29 @@ def test_main(
...
+    assert permissions[ERC721Permissions.UPDATE_METADATA] is True
+    assert permissions[ERC721Permissions.STORE] is True
```
trentmc, line 57: Remove "is True", "is False" like above

tests/flows/test_vesting.py
```diff
@@ -6,54 +6,30 @@
 from web3 import exceptions
```
trentmc, line 6: In this file, every test function up to (but not including) `test_pool_creation_fails_for_incorrect_vesting_period()` doesn't have anything to do with testing vesting, and should be moved to the appropriate files. Either in this PR or in a quick follow-up one.

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+        from_wallet=consumer_wallet,
+    )
+    block_deployed_pool = block_before_deploying_pool + 2
+    assert bpool.is_finalized()
+    assert bpool.opc_fee() == to_wei("0.001")
+    assert bpool.get_swap_fee() == to_wei("0.003")
+    assert bpool.community_fee(get_address_of_type(config, "Ocean")) == 0
+    assert bpool.community_fee(erc20_token.address) == 0
+    assert bpool.publish_market_fee(get_address_of_type(config, "Ocean")) == 0
+    assert bpool.publish_market_fee(erc20_token.address) == 0
```
trentmc, line 219: A function should always do one thing and do it well. This is the case for test functions too.

This function is about testing the main flow of vesting. Alright. 

What I've highlighted is a bunch of tests for fees. This is inappropriate for this function's target behavior. And inappropriate for this module. Please move those tests into the appropriate module.

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+    # advance checkpoint_block blocks to see if available vesting increased
+    counter = 0
+    for block in range(checkpoint_blocks[-1]):
+        # send dummy transactions to increase block count (not part of the actual functionality)
+        send_ether(publisher_wallet, ZERO_ADDRESS, 1)
+        if block not in checkpoint_blocks:
+            continue
+        available_vesting_after = side_staking.get_available_vesting(
+            erc20_token.address
+        )
```
trentmc, line 307: This should be one line

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+
+    block_before_deploying_pool = web3.eth.block_number
+    bpool = _deploy_ocean_pool(
+        erc20_token=erc20_token,
+        factory_router=factory_router,
+        config=config,
+        web3=web3,
+        from_wallet=publisher_wallet,
+    )
+    assert bpool.is_finalized()
```
trentmc, line 284: Remove line, out of scope for this function's testing

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+    bpool = _deploy_ocean_pool(
+        erc20_token=erc20_token,
+        factory_router=factory_router,
+        config=config,
+        web3=web3,
+        from_wallet=publisher_wallet,
+    )
+
+    # check if the vesting amount is correct
+    assert side_staking.get_vesting_amount(erc20_token.address) == vesting_amount
```
trentmc, line 366: You don't need to test this in this function, since you've tested this basic functionality already in other functions. Delete line.

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+    # advance 10 blocks to see if available vesting increased
+    for _ in range(9):
+        # send dummy transactions to increase block count (not part of the actual functionality)
+        send_ether(consumer_wallet, ZERO_ADDRESS, 1)
+
+    available_vesting_after = side_staking.get_available_vesting(erc20_token.address)
+    assert (
+        available_vesting_after > available_vesting_before
+    ), "Available vesting was not increased!"
+    side_staking.get_vesting(erc20_token.address, publisher_wallet)
```
trentmc, line 381: Delete line, it does nothing

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+
+    # advance 10 blocks to see if available vesting increased
+    for _ in range(9):
+        # send dummy transactions to increase block count (not part of the actual functionality)
+        send_ether(consumer_wallet, ZERO_ADDRESS, 1)
+
+    available_vesting_after = side_staking.get_available_vesting(erc20_token.address)
+    assert (
+        available_vesting_after > available_vesting_before
+    ), "Available vesting was not increased!"
```
trentmc, line 380: You don't need to test this in this function, since you've tested this basic functionality already in other functions. Delete lines.

Focus on getting to the heart of what _this_ function is trying to test for. Cut the rest out.

tests/flows/test_vesting.py
```diff
@@ -106,73 +103,535 @@ def test_main(
...
+    assert side_staking.get_vesting_amount(erc20_token.address) == vesting_amount
+
+    # check if vesting is correct
+    dt_balance_before = erc20_token.balanceOf(erc20_token.get_payment_collector())
+    available_vesting_before = side_staking.get_available_vesting(erc20_token.address)
+
+    # advance 5 blocks to see if available vesting increased
+    for _ in range(4):
+        # send dummy transactions to increase block count (not part of the actual functionality)
+        send_ether(consumer_wallet, ZERO_ADDRESS, 1)
```
trentmc, line 235: You have 4 or 5 "advance N blocks" chunks of code, each doing the same thing except for N. DRY violation.

To fix: create a new method `_advanceBlocks(num_blocks:int)`, then make these chunks of code call that method instead.

trentmc (changes requested):
This has improved greatly since the last round. I was now able to really look at the functionality of the code, without all the clutter of DRY violations. It's looking to be in decent shape. Great stuff.

Please fix the comments - they're mostly small - then good to go.

tests/flows/test_vesting.py
mariacarmina: I need to call this after I advanced 10 blocks, because it will create the event `Vesting` and will claim the datatokens to the publisher (I verify this in the next assertions)

tests/flows/test_vesting.py
mariacarmina: I have removed those line of code because there were already in `test_swap_fees.py`.

tests/flows/test_vesting.py
mariacarmina: I have created a separate file `test_bpool.py` and put `test_pool_creation_fails_for_incorrect_vesting_period()` there.

tests/flows/test_vesting.py
mariacarmina: I removed all the redundant statements like this for `test_vesting.py` file.

ocean_lib/models/btoken.py
```diff
@@ -73,3 +73,10 @@ def allowance(self, src_address: str, dst_address: str) -> int:
         :return: int
         """
         return self.contract.caller.allowance(src_address, dst_address)
+
+    @enforce_types
+    def total_supply(self) -> int:
+        """
+        :return: int
+        """
```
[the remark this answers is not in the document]
mariacarmina: I have removed those redundant docstrings for each method.

tests/flows/test_vesting.py
mariacarmina: Done

tests/flows/test_vesting.py
mariacarmina: Removed all the occurrences

tests/flows/test_vesting.py
mariacarmina: Kept one of this in `test_vesting_main_flow` & deleted in rest

tests/flows/test_vesting.py
mariacarmina: Kept one of this in `test_vesting_main_flow` & cut the rest out

trentmc (approved):

ocean_lib/models/test/test_erc721_token.py
```diff
@@ -794,8 +795,47 @@ def test_transfer_nft(web3, config, publisher_wallet, consumer_wallet, factory_r
...
+    tx_receipt = web3.eth.wait_for_transaction_receipt(tx)
+    transfer_event = erc721_nft.get_event_log(
+        ERC721FactoryContract.EVENT_TRANSFER,
+        tx_receipt.blockNumber,
+        web3.eth.block_number,
+        None,
+    )
+    assert transfer_event[0].event == "Transfer"
+    assert transfer_event[0].args["from"] == publisher_wallet.address
+    assert transfer_event[0].args.to == consumer_wallet.address
```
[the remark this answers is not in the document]
mariacarmina: It's fixed in #776

PR #760, opened 2022-03-21
Issue759 v4 cleanup

### LINKED ISSUE #759: V4 Cleanup
ocean.py needs a few minor cleanups:

1. ERC721Factory methods are using unnamed argument lists. Should use named arguments instead.
2. Inconsistent argument names, especially for fees and "base_token" which is two words.
3. Large portion of models/ directory is missing type enforcement and as a result, many of the type hints are wrong.
4. There are places where `wait_for_transaction_receipt` is called unnecessarily (ContractBase automatically does this when calling send_transaction)

DMats:
Fixes #759.

Changes proposed in this PR:

- Use named arguments in ERC721Factory
- Fix Inconsistent argument names, especially for fees and "base_token" which is two words.
- Enforce types everywhere
- Fix type hints
- Remove unnecessary calls to `wait_for_transaction_receipt`
- Add `get_wallet(index)` function that returns a wallet with the private key defined at env var `TEST_PRIVATE_KEY{index}`
- Add initial prototype for `test_fees` (Please ignore. currently skipped because it's not complete.  Will be completed in #754 )
- Add `TEST_PRIVATE_KEY4-8` to `pytest.ini`

Notes on fee naming:

1. `consume_market_swap_fee_address` is an "address" because the tokens are sent directly to the address during the swap. `publish_market_swap_fee_collector` is a "collector" because the publish market must collect their fees from the pool/exchange at a later time, after the swap is complete.


tests/fees/test_fees.py
```diff
@@ -0,0 +1,50 @@
+#
```
DMats, line 1: Please ignore this file for now. The `test_fees` test is skipped and will be completed in #754

alexcos20:
you are not sending enough params to fixedrate:  https://github.com/oceanprotocol/contracts/blob/feature/v1.0.0-rc1/contracts/pools/fixedRate/FixedRateExchange.sol#L213

ocean_lib/models/bpool.py
```diff
@@ -403,14 +448,15 @@ def join_swap_extern_amount_in(
...
+    @enforce_types
     def exit_swap_pool_amount_in(
         self,
         pool_amount_in: int,
         min_amount_out: int,
         from_wallet: Wallet,
     ) -> str:
         """
-        Pay `poolAmountIn` pool shares into the pool, getting `tokenAmountOut`
+        Pay `poolAmountIn` pool shares into the poopool_template_addresspool_template_addressl, getting `tokenAmountOut`
```
calina-c, line 459: I think there's a typo here, it's `poopool`

calina-c (commented):
I've noticed a typo. Other than that, let's discuss whether these are the "final" or "final-ish" namings, fix the conflicts and eventually merge.

ocean_lib/models/bpool.py
DMats: Fixed.

calina-c (approved):

PR #772, opened 2022-03-30
Refactor stress tests

### LINKED ISSUE #751: Refactor stress tests arhitecture
This is a follow-up issue for #718 to refactor the tests arhitecture.

mariacarmina:
Fixes #751  .

Changes proposed in this PR:

- used `ThreadPoolExecutor` instead of creating threads manually
- updated the path for stress tests.


mariacarmina:
Here is the result for 20 workers with 50 reps for `pool_creation_with_threads.py` (as an example).
Screenshot from 2022-04-11 19-31-47

tests/stress/compute_flow_with_threads.py
```diff
@@ -261,17 +239,45 @@
...
+                "brainin",
+                "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff",
+                "gpr",
+                "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/gpr.py",
+                "python-brain",
+            )
+
+
+@pytest.mark.slow
+def test_concurrent_c2d():
```
DMats, line 279: I'd like to see this test parameterized so that all three concurrency settings are run (even if one of them fails)

```python
@pytest.mark.slow
@pytest.mark.parameterize(["concurrent_flows", "repetitions"], [(1, 3000), (3, 1000), (20, 50)])
def test_concurrent_c2d(concurrent_flows: int, repetitions: int)
    concurrent_c2d(concurrent_flows, repetitions)
```

tests/stress/pool_creation_with_threads_test.py
```diff
@@ -90,69 +90,18 @@ def asset_displayed_on_sale(ocean: Ocean, wallet: Wallet):
...
-        thread.join()
-        print(f"Main    : thread {index} done")
+    with ThreadPoolExecutor(max_workers=concurrent_flows) as executor:
+        for _ in range(concurrent_flows * duration):
+            executor.submit(asset_displayed_on_sale, ocean)
+
+
+@pytest.mark.slow
+def test_concurrent_pool_creation():
+    concurrent_flows_values = [1, 3, 20]
```
DMats, line 104: Please see previous comments. Please parameterize.

tests/stress/publish_flow_with_threads_test.py
```diff
@@ -0,0 +1,68 @@
...
+    assert asset.nft["address"] == erc721_nft.address
+    assert asset.nft["owner"] == wallet.address
+    assert asset.datatokens[0]["name"] == "ERC20DT1"
+    assert asset.datatokens[0]["symbol"] == "ERC20DT1Symbol"
+    assert asset.datatokens[0]["address"] == erc20_token.address
+
+
+@pytest.mark.slow
+def test_concurrent_pool_creation():
+    concurrent_flows_values = [1, 3, 20]
```
DMats, line 65: Please see previous comments. Please parameterize.

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -0,0 +1,164 @@
...
+        destination=destination,
+        order_tx_id=tx_id,
+    )
+
+    assert (
+        len(os.listdir(os.path.join(destination, os.listdir(destination)[0]))) > 0
+    ), "The asset folder is empty."
+
+
+def concurrent_consume_flow(concurrent_flows: int, duration: int):
```
DMats, line 150: Please see the previous comment. Consider `repetitions` instead of `duration`

tests/stress/pool_creation_with_threads_test.py
```diff
@@ -90,69 +90,18 @@ def asset_displayed_on_sale(ocean: Ocean, wallet: Wallet):
     assert formatted_price
 
 
-@pytest.mark.slow
-def test_pool_creation_with_threads():
+def concurrent_pool_creation(concurrent_flows: int, duration: int):
```
DMats, line 93: Please see previous comments. Consider `repetitions` instead of `duration`

tests/stress/publish_flow_with_threads_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def _get_publishing_requirements(ocean: Ocean, wallet: Wallet, config: Config):
+    erc721_nft, erc20_token = deploy_erc721_erc20(ocean.web3, config, wallet, wallet)
+    data_provider = DataServiceProvider
+    _, metadata, encrypted_files = create_basics(config, ocean.web3, data_provider)
+    return erc721_nft, erc20_token, metadata, encrypted_files
+
+
+def concurrent_publish_flow(concurrent_flows: int, duration: int):
```
DMats, line 26: Please see previous comments. consider `repetitions` instead of `duration`.

DMats:
@mariacarmina Do all the stress tests actually run in GitHub checks successfully? Since they're marked `slow` they should run along with the other slow tests once a week.

I think you could test this by removing the `slow` mark temporarily.

mariacarmina:
> Overall, really great work with this @mariacarmina. I'm excited to see this test pass with 20 concurrent flows.
> 
> I also really love the reduction in code duplication.
> 
> Out of curiosity, did you try a test with even more concurrency, say 3000 concurrent flows for 1 repetition? Is there a point where it breaks? I also wonder if there might be a case for mixing the flows. Like 20x of all flow types at the same time.

I can add 3000 workers if wanted, I remember you suggested maximum 20 workers. I tried only with max 20.

mariacarmina:
> @mariacarmina Do all the stress tests actually run in GitHub checks successfully? Since they're marked `slow` they should run along with the other slow tests once a week.
> 
> I think you could test this by removing the `slow` mark temporarily.

It takes more than 10 minutes for each test.

mariacarmina:
After one run of a test file, the ganache wallet will be out of ETH (the one that transfers funds).
Screenshot from 2022-04-12 20-23-12

DMats (changes requested):

DMats:
> After one run of a test file, the ganache wallet will be out of ETH (the one that transfers funds).

This seems like a problem - it means that when the "slow" tests run, the second test file will fail. Can we replenish the ETH somehow?  

If not, can we reduce the overall number of repetitions?

3000 -> 300
1000 -> 100
50 -> 5

tests/stress/pool_creation_with_threads_test.py
```diff
@@ -90,69 +90,16 @@ def asset_displayed_on_sale(ocean: Ocean, wallet: Wallet):
...
-        print(f"Main    : before joining thread {index}.")
-        thread.join()
-        print(f"Main    : thread {index} done")
+    with ThreadPoolExecutor(max_workers=concurrent_flows) as executor:
+        for _ in range(concurrent_flows * repetitions):
+            executor.submit(asset_displayed_on_sale, ocean)
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize(["concurrent_flows", "repetitions"], [(1, 3000)])
```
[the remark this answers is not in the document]
mariacarmina: Yes, sure, thanks.

tests/stress/compute_flow_with_threads.py
```diff
@@ -200,19 +178,19 @@
     DATA_asset = ocean.assets.resolve(DATA_did)
     ALGO_asset = ocean.assets.resolve(ALGO_did)
 
-    compute_service = get_first_service_by_type(DATA_asset, "compute")
-    algo_service = get_first_service_by_type(ALGO_asset, "access")
+    compute_service = DATA_asset.services[0]
```
calina-c, line 181: I'm not sure about the business logic here. While it is true that currently these indices are correct, we specify in the readmes that this is just a coincidence, and the first compute/access index might be different for different DDOs. We made the decision a few iterations ago that we should keep `get_first_service_by_type` in tests. Should we revisit that decision?

tests/stress/compute_flow_with_threads.py
```diff
@@ -200,19 +178,19 @@
...
     # Pay for dataset
     DATA_order_tx_id = ocean.assets.pay_for_service(
         asset=DATA_asset,
         service=compute_service,
-        consume_market_order_fee_address=bob_wallet.address,
+        consume_market_order_fee_address=wallet.address,
         consume_market_order_fee_token=DATA_datatoken.address,
         consume_market_order_fee_amount=0,
-        wallet=bob_wallet,
+        wallet=wallet,
```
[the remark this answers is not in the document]
calina-c: Actually, even if we double-assign this variable, I'd rather have relevant namings. i.e. consumer_wallet. That is, if for the purposes of this test, the publisher_wallet and the consumer_wallet are the same, I'd rather have `consumer_wallet = publisher_wallet = Wallet(...)` than risk confusion. Or, whatever: `alice_wallet`, `bob_wallet`. `wallet` is not a good variable name because it is vague.

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -0,0 +1,164 @@
...
+    file1 = FilesTypeFactory(file1_dict)
+    file2_url = "https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-abstract10.xml.gz-rss.xml"
+    file2_dict = {"type": "url", "url": file2_url, "method": "GET"}
+    file2 = FilesTypeFactory(file2_dict)
+    file3_url = (
+        "https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-abstract10.xml.gz"
+    )
+    file3_dict = {"type": "url", "url": file3_url, "method": "GET"}
+    file3 = FilesTypeFactory(file3_dict)
+    files = [file1, file2, file3]
```
calina-c, line 54: can we maybe use some fixtures for files? It seems we are repeating this code a lot. Or maybe a function?

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -0,0 +1,164 @@
...
+        account_address=wallet.address,
+        value=ocean.to_wei(20),
+        from_wallet=wallet,
+    )
+    # Start order for consumer
+    tx_id = erc20_token.start_order(
+        consumer=wallet.address,
+        service_index=ddo.get_index_of_service(service),
+        provider_fee_address=response.json()["providerFee"]["providerFeeAddress"],
+        provider_fee_token=response.json()["providerFee"]["providerFeeToken"],
```
calina-c, line 108: oh, please assign response.json()["providerFee"] first. This is a nightmare to read.

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -0,0 +1,164 @@
...
+        s=response.json()["providerFee"]["s"],
+        valid_until=response.json()["providerFee"]["validUntil"],
+        provider_data=response.json()["providerFee"]["providerData"],
+        consume_market_order_fee_address=wallet.address,
+        consume_market_order_fee_token=erc20_token.address,
+        consume_market_order_fee_amount=0,
+        from_wallet=wallet,
+    )
+    # Download file
+    destination = config.downloads_path
```
calina-c, line 121: can we move 121-131 into a function? It seems useful, maybe even for the main functionality. Also, are you sure this works on both mac and linux? AFAIK, there are some tiny differences with os dir utils.

calina-c (changes requested):

tests/stress/consume_flow_with_threads_test.py
mariacarmina: I haven't tested on mac, just linux environment.

tests/stress/compute_flow_with_threads.py
```diff
@@ -261,17 +242,45 @@
...
+                "brainin",
+                "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff",
+                "gpr",
+                "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/gpr.py",
+                "python-brain",
+            )
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize(["concurrent_flows", "repetitions"], [(1, 2), (3, 1), (5, 5)])
```
mariacarmina, line 284: I have decreased the number of repetitions & number of workers because the ether balance is not enough for all the tests from this folder. When these tests will be run, at some point `barge` needs to be pruned (for filling the accounts again with ETH).
After 2 runs of `consume_flow_with_threads.py` with 5 workers and 5 repetitions
Screenshot from 2022-04-19 13-17-46

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -157,8 +134,43 @@ def concurrent_consume_flow(concurrent_flows: int, repetitions: int):
...
 @pytest.mark.slow
-@pytest.mark.parametrize(
-    ["concurrent_flows", "repetitions"], [(1, 300), (3, 100), (20, 5)]
-)
+@pytest.mark.parametrize(["concurrent_flows", "repetitions"], [(1, 2), (3, 1), (5, 5)])
 def test_concurrent_consume_flow(concurrent_flows, repetitions):
     concurrent_consume_flow(concurrent_flows, repetitions)
+
+
+def _create_downloads_path(config):
```
calina-c, line 142: why can't we use tmpdir in these tests?

tests/stress/consume_flow_with_threads_test.py
```diff
@@ -157,8 +134,43 @@ def concurrent_consume_flow(concurrent_flows: int, repetitions: int):
...
+    if not os.path.exists(destination):
+        os.mkdir(destination)
+
+    assert len(os.listdir(destination)) == 0
+
+    return destination
+
+
+def _create_files():
+    file1_url = "https://raw.githubusercontent.com/tbertinmahieux/MSongsDB/master/Tasks_Demos/CoverSongs/shs_dataset_test.txt"
```
calina-c, line 163: let's create a new issue for file fixtures :)
I think to avoid duplication we can keep these definitions inside the readme tests, but I've seen the tbertinmahieux/MSongsDB file so many times in our test suite that I'm pretty sure it warrants a fixture.

calina-c (commented):
I left just one more question and a suggestion for a future issue.

tests/stress/consume_flow_with_threads_test.py
mariacarmina: Sure, I'll create an issue that creates fixtures for these files and paste it here.

tests/stress/compute_flow_with_threads.py
DMats: I think these tests need to demonstrate at least 20 concurrent users. So I think we need to send more ETH to the wallets.

mariacarmina: From where? From other ganache wallets? It works fine one time with 20 workers, but it won't run for all scenarios + plus the other tests.

tests/stress/compute_flow_with_threads.py
DMats: `get_first_service_by_type` is okay in tests where we know the asset only has 1 service of the given type.

If the asset has 2 services of the same type, then we should look up the asset using the index or the id.

```python
asset.get_service_by_index()
asset.get_service_by_id()
```

tests/stress/compute_flow_with_threads.py
DMats: I think we should increase the default wallet balance in barge to 1 million ETH.

```console
ganache-cli -e 1000000
```

DMats: @mariacarmina I created barge PR to fix this.
https://github.com/oceanprotocol/barge/pull/286

tests/stress/consume_flow_with_threads_test.py
DMats: I believe the PR for adding file fixtures can be found here: https://github.com/oceanprotocol/ocean.py/pull/792

DMats (approved):

calina-c (approved):

PR #776, opened 2022-04-04
Remove usages of deploy_erc721_erc20 function

### LINKED ISSUE #765: Fix DRY violations in test_erc721_token.py & related

`ocean_lib/models/test/test_erc721_token.py` has ten calls to `deploy_erc721_erc20()`. There are four more in `test_erc20_token.py`, and five more in other test modules of this directory. Most of these take about 5 lines, and have exactly the same boilerplate code.

This violates DRY (DRY = Don't Repeat Yourself). In turn, this hurts maintainability, extensibility, etc.

Use the directory's `conftest.py` more, as pytest intends it to be. Right now it's only being used to define wallets. Add routine(s) in it to create erc721 & erc20 with a single call to  `deploy_erc721_erc20()`. Then for each test function: (1) add `data_nft` or `datatoken` argument, drawing on the conftest info, and (2) remove its call to `deploy_erc721_erc20`.  

With this, we can remove most calls to `deploy_erc721_erc20()` in the test py files.

Note: there are probably other great opportunities to leverage conftest.py to fix DRY violations. This was simply the first one I noticed, as I hadn't looked at the tests in a while.

mariacarmina:
Fixes #765  .

Changes proposed in this PR:

- add fixtures for ERC721 token, ERC20 token and ERC721 factory;
- remove usages of deploy_erc721_erc20 function (kept some occurrences);
- formed smaller functions when needed;
- removed `is True` & `is False` parts from assertions in `test_erc721_token.py` and from `test_vesting.py`


calina-c (approved):
Excellent work!

trentmc (changes requested):
There are certainly improvements.

Before, there were ten calls to `deploy_erc721_erc20()`. Now, there are still four.  Eg `erc721_token_v2 = deploy_erc721_erc20(..)`. Why?

Looks like still quite a lot of DRY violation happening.

Also, I'm seeing many `assert ( ... ) is False` and `assert (...) is True`. You shouldn't need the `is False` or `is True` parts, it should just be `assert (...)` or `assert not (...)`

mariacarmina:
@trentmc Can I create a separate issue for removing the `is True` & `is False` parts from assertions, because now I only removed for `test_erc721_token.py` and for `test_vesting.py`? Or can I do this work here too?

ocean_lib/models/test/test_erc20_token.py
```diff
@@ -14,35 +14,34 @@
...
+        erc20_token.event_MinterApproved.abi["name"] == ERC20Token.EVENT_MINTER_APPROVED
+    )
+    assert erc20_token.event_OrderReused.abi["name"] == ERC20Token.EVENT_ORDER_REUSED
+    assert (
+        erc20_token.event_OrderExecuted.abi["name"] == ERC20Token.EVENT_ORDER_EXECUTED
+    )
+    assert (
+        erc20_token.event_PublishMarketFeeChanged.abi["name"]
         == ERC20Token.EVENT_PUBLISH_MARKET_FEE_CHANGED
     )
```
trentmc, line 39: Did black do these changes? Yuck, it's so much harder to read

ocean_lib/models/test/test_erc20_token.py
```diff
@@ -404,69 +394,60 @@ def test_main(web3, config, publisher_wallet, consumer_wallet, factory_router):
...
-    erc20.approve(publisher_wallet.address, to_wei("10"), consumer_wallet)
+    erc20_token.approve(publisher_wallet.address, to_wei("10"), consumer_wallet)
 
-    assert erc20.allowance(consumer_wallet.address, publisher_wallet.address) == to_wei(
-        "10"
-    )
-    erc20.burn_from(consumer_wallet.address, to_wei("2"), publisher_wallet)
+    assert erc20_token.allowance(
+        consumer_wallet.address, publisher_wallet.address
+    ) == to_wei("10")
```
trentmc, line 410: Make this easier to read by e.g. : 

```python
allowance = erc20_token.allowance(consumer_wallet.address, publisher_wallet.address)
assert allowance == to_wei("10")
```

ocean_lib/models/test/test_erc721_token.py
```diff
@@ -15,7 +15,7 @@
 )
```
trentmc, line 15: Many lines in this file (and others?) seem to have gone through `black` which converted them into 3-4 line statements, way harder to read. You should always ask yourself: how do I preserve readability?

For this file (and others?), here's one trick. Many of the lines have `consumer_wallet.address`. What you can do is add new function argument `consumer_addr`, which is defined in a method in conftest.py to equal consumer_wallet.address. Do the same for other often-used addresses: `publisher_wallet.address` and `another consumer_wallet.address`. In shortening that text, you should be able to convert back many of the 3-4 line chunks, back into a single line.

trentmc (commented):
Overall this is in pretty decent shape - the big changes are in, they do what they're supposed to, good. My comments were more for a few smaller things, where some changes converted 1 line to 3-4 lines, and I'm trying to find ways to keep things readable.

ocean_lib/models/test/test_erc20_token.py
mariacarmina: Yep, it's due to black formatting.

ocean_lib/models/test/test_erc721_token.py
mariacarmina: It did some improvements. Thanks!

ocean_lib/models/test/test_erc20_token.py
mariacarmina: Done

ocean_lib/models/test/test_erc721_token.py
mariacarmina: I have created fixtures for all 3 used wallets in conftest.py: `publisher_addr`, `consumer_addr` and `anothor_consumer_addr`.

trentmc (approved):

PR #785, opened 2022-04-12
Disconnecting components

mariacarmina:
Fixes oceanprotocol/multi-repo-issue#138.

Changes proposed in this PR:

- create some tests which decouple the components for a bit;
- tests manually with barge the behavior of disconnecting the provider & aquarius.


mariacarmina:
I ran couple of tests with barge and then I have turned it off to see the behavior of pytest.
At the beginning, it froze couple of seconds and then it threw error log (for not finding the connection to the components).

Screenshot from 2022-04-13 21-21-45

After these errors occurred, I started again barge, but nothing seemed to re-establish and the tests were still failing. The problem is that pytest couldn't re-establish the connection with the other components after they were turned off couple of seconds.

bad-aqua-config.ini
```diff
@@ -0,0 +1,14 @@
+[eth-network]
```
calina-c, line 1: please move this somewhere under tests or a samples folder. This file does not belong in the main codebase.

calina-c (changes requested):

calina-c (approved):

PR #790, opened 2022-04-20
Fix issue with large files for download.

calina-c:
Work on https://github.com/oceanprotocol/provider/issues/253. Not marking as closed since maybe it's worth checking ocean.js as well. Here's a sample file: https://test2.fibertelecom.it/5GB.zip


AnaLoznianu:
@calina-c We'll track ocean.js behavior in this [issue](https://github.com/oceanprotocol/ocean.js/issues/1427). If the provider works as expected, we can close this issue.

mariacarmina (approved):

PR #810, opened 2022-05-09
Added additionalInformation key for Service class

### LINKED ISSUE #808: service didn't have additional information attribute
In service.py it didn't have the additional Information attribute 
https://github.com/oceanprotocol/ocean.py/blob/b6707dbae25a2fde04a00d0ebd34de8b5dfb0fc3/ocean_lib/services/service.py#L37-L47

But its available on ocean.js 
https://github.com/oceanprotocol/ocean.js/blob/1d327bf1008917b2b04e03ae43840fa577d397d6/src/%40types/DDO/Service.ts#L105-L109

mariacarmina:
Fixes #808  .

Changes proposed in this PR:

- added `additionalInformation` key for Service class;
- tested the functionality.


ocean_lib/services/test/test_service.py
```diff
@@ -20,13 +20,15 @@ def test_service():
     """Tests that the get_cost function for ServiceAgreement returns the correct value."""
     ddo_dict = get_sample_ddo()
     service_dict = ddo_dict["services"][0]
+    service_dict["additionalInformation"] = {"message": "Sample DDO"}
```
calina-c, line 23: Regarding the "as_dictionary" conversion, I don't think it's necessary. That case, in particular, is for when we have Asset or Service, where we implement specific as_dictionary functions. But here we have a json that comes directly from aquarius, it shouldn't contain any Python objects, just regular json.

Please add a test that also contains lists/nesting inside additionalInformation.

ocean_lib/services/service.py
```diff
@@ -42,6 +43,14 @@ def __init__(
         self.compute_values = compute_values
         self.name = name
         self.description = description
+        self.additional_information = None
+
+        if additional_information:
+            self.additional_information = {}
+            self._reserved_names = ["id", "type", "serviceEndpoint"]
```
calina-c, line 50: I think you can have anything you want in additionalInformation. Previously with "additional values" or however it was called in v3, the values were added directly inside the DDO, at root level. Now, if we assign the dictionary inside "additionalInformation", then there is no point in checking reserved names. Basically, we do not care to police whatever the user adds there.

ocean_lib/services/service.py
```diff
@@ -157,6 +167,19 @@ def as_dictionary(self) -> Dict[str, Any]:
         if self.description is not None:
             values["description"] = self.description
 
+        if self.additional_information is not None:
+            additional_information = {}
+            for key, value in self.additional_information.items():
```
calina-c, line 172: See my other (older) comment from test_service.py regarding this conversion. I think it is unnecessary.

calina-c (changes requested):
I left a few more comments.

calina-c (approved):

PR #825, opened 2022-05-18
Adds slow test notifications.

calina-c:


.github/workflows/slow.yml
```diff
@@ -42,3 +43,10 @@ jobs:
       - name: Test with pytest
         run: |
           py.test -m slow
+          slack_api:
+      - name: Slack notify via api
```
DMats, line 47: This name is incorrect.  We're not using the API, we're using the Incoming Webhook.

.github/workflows/slow.yml
```diff
@@ -42,3 +43,10 @@ jobs:
       - name: Test with pytest
         run: |
           py.test -m slow
+          slack_api:
```
DMats, line 46: What is this line doing?  I think we can safely remove it - I don't see anything like this in the up9cloud/action-notify docs.

.github/workflows/slow.yml
```diff
@@ -42,3 +43,10 @@ jobs:
...
+      - name: Slack notify via api
+        uses: up9cloud/action-notify@master
```
DMats, line 48: When I first read this, I was concerned that we were using a github action that only had 1 star on Github and wondered why we weren't using the official [slack-send](https://github.com/marketplace/actions/slack-send) action.

After reading into it a bit, I realized that `slack-send` is lower level, more flexible, but requires more setup - we'd have to craft the messages ourselves from scratch.

I haven't seen what `up9cloud/action-notify` messages look like, but I'm hoping they're nicely formatted out-of-the-box.

This is just a comment, no change necessary.

DMats (changes requested):
Nice work @calina-c. I left just a couple of change requests.

.github/workflows/slow.yml
calina-c: No big deal on how it looks, it just links to the action and says: success or failure. Check the slack channel for an example, since it has already ran once. I can also change the app image (which I created in Slack) to look prettier but I have no idea in particular as to what it should be :))

DMats (approved):

PR #826, opened 2022-05-18
Changed web3.eth.sign

### LINKED ISSUE #820: Change web3.eth.sign from the production code
web3.eth.sign is not compatible with certain infura RPCs and it needs to be replaced with provider endpoint `initialize` which returns the fees for the provider. It is used [here](https://github.com/oceanprotocol/ocean.py/blob/v4main/ocean_lib/ocean/ocean.py#L359)

mariacarmina:
Fixes #820  .

Changes proposed in this PR:

- replaced `web3.eth.sign` with calling directly the `initialize endpoint` from provider in order to retrieve the provider fees
in `ocean.py` & `erc20-enterprise.md`
- tested in `test_erc20_enterprise.py`


ocean_lib/models/test/test_erc20_enterprise.py
```diff
@@ -245,25 +246,22 @@ def test_buy_from_fre_and_order(
         amount=consume_fee_amount,
         from_wallet=publisher_wallet,
     )
+    data_provider = DataServiceProvider
+    _, metadata, encrypted_files = create_basics(config, web3, data_provider)
```
DMats, line 250: The `create_basics` method is for testing purposes only, and probably should not be included in READMEs.

ocean_lib/ocean/ocean.py
```diff
@@ -335,33 +333,37 @@ def create_pool(
...
-            ["bytes", "address", "address", "uint256", "uint256"],
-            [
-                self.web3.toHex(self.web3.toBytes(text=provider_data)),
-                provider_fee_address,
-                provider_fee_token,
-                provider_fee_amount,
-                valid_until,
-            ],
+        data_provider = DataServiceProvider
+        ddo = self.assets.create(
```
DMats, line 345: The `retrieve_provider_fees` method should probably not call the `assets.create()` method. 

Publishing the asset happens once.  But you'll need to retrieve provider fees potentially many times.

DMats (changes requested):

calina-c:
First of all, why is this tested separately? Won't it work with the automated readme tests?
Second, we have moved away from publishing asset with metadata inside every readme. Please use a structure similar to the marketplace and consume flows, where we refer to the publish-flow readme. This means adding this readme to the list of readme tests that run after the data nft + datatoken and publish flow readmes, see in test_readmes.py.

ocean_lib/ocean/ocean.py
```diff
@@ -335,33 +334,23 @@ def create_pool(
...
-            ["bytes", "address", "address", "uint256", "uint256"],
-            [
-                self.web3.toHex(self.web3.toBytes(text=provider_data)),
-                provider_fee_address,
-                provider_fee_token,
-                provider_fee_amount,
-                valid_until,
-            ],
+    def retrieve_provider_fees(self, asset: Asset, publisher_wallet: Wallet) -> tuple:
+        data_provider = DataServiceProvider
```
calina-c, line 338: I think there is some data provider already set on ocean, isn't there?
Also, regarding the next line, I think should we pass in the service as well, as a parameter to this function. If there is no access service, the error here will be ugly.

mariacarmina: But since we use `ocean.assets.create` function before, in order to create an asset, you have already an access service generated by default.

READMEs/erc20-enterprise.md
```diff
@@ -68,6 +68,19 @@ erc20_enterprise_token = erc721_nft.create_datatoken(
 )
 print(f"ERC20 Enterprise address: {erc20_enterprise_token.address}")
 
+```
+Then, please refer to [publish-flow](publish-flow.md) to generate your metadata and encrypted files.
+Asset creation will be based on the deployment of ERC20 Enterprise token like this:
```
calina-c, line 73: @trentmc I'm tagging you here for approval. I advised Maria to use this structure to avoid repeating herself. The publishing part is described in the publish-flow readme, which we recently introduced. Since the only change here is in the asset creation itself (one instruction differs), while the metadata and encrypted files are identical to the publish flow, I advised against redoing the whole thing and using this language instead.

I think it's the right call, but an OK from you is probably warranted. Thanks!

calina-c (commented):
LGTM. I tagged Trent for an extra approval on the readme changes, so I'm not fully approving yet.

DMats:
This PR looks good to me, as well.  If @trentmc approves the README changes, then I think it can be merged.

calina-c:
> @mariacarmina I think you should also add an Ocean API method that calls initializeCompute, named something like `ocean.retrive_provider_fees_for_compute()` or `ocean.retrieve_provider_fees_compute()`
> 
> CC @calina-c Please let me know if this makes sense.
> 
> I think it would be useful for the following flow:
> 
> // Pseudocode
> 
> 1. Create `dataset` and `algorithm` assets with erc20_enterprise tokens
> 2. `initialize_compute(dataset, algorithm)` <- currently requires importing `DataServiceProvider`
> 3. `dataset_erc20_enterprise.buy_from_fre_and_order()`
> 4. `algorithm_erc20_enterprise.buy_from_fre_and_order()`
> 5. `ocean.compute.start()`

Agreed :)

DMats (changes requested):
Please see my previous comment: https://github.com/oceanprotocol/ocean.py/pull/826#issuecomment-1135096265

calina-c:
@DMats I've created a separate issue to address your comment.

calina-c:
@DMats I'll alsso merge this one since it is the last needed to have a clean slate. I'll handle your comment on provider fees separately.

DMats:
Sounds good to me @calina-c. Thanks for creating the new issue.

PR #835, opened 2022-05-23
Added checks for template index.

### LINKED ISSUE #834: Add cap for ERC20 enterprise creation
ERC20 enterprise token does not override the cap with `0xfff...` value.

**Solution**
Before the contract function call of ERC20 creation, check the template index and throw an error if the cap is None for this specific case.

mariacarmina:
Fixes #834  .

Changes proposed in this PR:

- add cap for ERC20 enterprise token creation


ocean_lib/models/erc721_factory.py
```diff
@@ -167,33 +167,64 @@ def create_nft_with_erc20(
...
+                            datatoken_publish_market_order_fee_address,
+                            datatoken_publish_market_order_fee_token,
+                        ],
+                        [datatoken_cap, datatoken_publish_market_order_fee_amount],
+                        datatoken_bytess,
+                    ),
                 ),
+                from_wallet,
+            )
+        elif datatoken_template == 2 and not datatoken_cap:
```
calina-c, line 199: please treat the exception first, then you can untangle the if statements.

calina-c:
Please add tests to keep coverage the same. I would recommend and expect over 90% differential coverage.

calina-c:
The code looks MUCH better. To quote a classic, "another life".
Please add the exception tests and we're good to go.

PR #847, opened 2022-05-30
Adds function and test for compute provider fees.

### LINKED ISSUE #846: Add an Ocean API method that calls initializeCompute, named something like `ocean.retrive_provider_fees_for_compute()` or `ocean.retrieve_provider_fees_compute()`
@mariacarmina I think you should also add an Ocean API method that calls initializeCompute, named something like `ocean.retrive_provider_fees_for_compute()` or `ocean.retrieve_provider_fees_compute()`

CC @calina-c Please let me know if this makes sense.

I think it would be useful for the following flow:

// Pseudocode
1. Create `dataset` and `algorithm` assets with erc20_enterprise tokens
2. `initialize_compute(dataset, algorithm)` <- currently requires importing `DataServiceProvider`
3. `dataset_erc20_enterprise.buy_from_fre_and_order()`
4. `algorithm_erc20_enterprise.buy_from_fre_and_order()`
5. `ocean.compute.start()`

_Originally posted by @DMats in https://github.com/oceanprotocol/ocean.py/issues/826#issuecomment-1135096265_

calina-c:
Closes #846.

### DIFF
--- a/ocean_lib/ocean/ocean.py
+++ b/ocean_lib/ocean/ocean.py
@@ -15,6 +15,7 @@ from ocean_lib.assets.asset import Asset
 from ocean_lib.config import Config
 from ocean_lib.data_provider.data_service_provider import DataServiceProvider
 from ocean_lib.models.bpool import BPool
+from ocean_lib.models.compute_input import ComputeInput
 from ocean_lib.models.dispenser import Dispenser
 from ocean_lib.models.erc20_token import ERC20Token
 from ocean_lib.models.erc721_factory import ERC721FactoryContract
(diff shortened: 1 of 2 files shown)


ocean_lib/ocean/ocean.py
```diff
@@ -366,3 +368,25 @@ def retrieve_provider_fees(
...
+        initialize_compute_response = DataServiceProvider.initialize_compute(
+            [x.as_dictionary() for x in datasets],
+            algorithm_data.as_dictionary(),
+            datasets[0].service.service_endpoint,
+            consumer_address,
+            compute_environment,
+            valid_until,
+        )
+
+        # TODO: maybe we should adjust this in some way?
```
DMats, line 391: I don't think we should adjust. 

I remember when we removed the response filtering for compute status. It's a similar situation here. I think it's better for the user to get the full response.

And now that I'm glancing at retrieve_provider_fees, I'm thinking maybe we shouldn't be adjusting that response either.

DMats (approved):
Looks good! I commented on the TODO - please remove it before merging. 

I think we need a separate issue to unfilter the initialize response returned by retrieve_provider_fees.

PR #856, opened 2022-06-12
Changes to file structure (schema v4.1.0)

calina-c:


ocean_lib/data_provider/data_service_provider.py
```diff
@@ -519,11 +406,9 @@ def compute_job_result(
...
-        if not response:
-            raise DataProviderException("No response on job result endpoint.")
-
-        if response.status_code != 200:
-            raise DataProviderException(response.content)
+        DataServiceProviderBase.check_response(
+            response, "jobResultEndpoint", compute_job_result_endpoint
+        )
 
         return response.content
```
mariacarmina, line 413: `response.content` returns bytes here, but the signature of this function requires `Dict[str, Any]`.

ocean_lib/data_provider/data_service_provider.py
```diff
@@ -562,185 +447,6 @@ def compute_job_result_logs(
 
         return function_result
```
mariacarmina, line 448: Expected type `Dict[str, Any]` and function_result has `List[Dict[str, Any]]` type.

ocean_lib/data_provider/base.py
```diff
@@ -0,0 +1,310 @@
...
+    @staticmethod
+    @enforce_types
+    def check_response(
+        response: Any,
+        endpoint_name: str,
+        endpoint: str,
+        success_codes: Optional[List] = None,
+        exception_type=DataProviderException,
+    ):
+        if not response or not hasattr(response, "status_code"):
```
mariacarmina, line 284: In `check_response` we can check if the response is None.

calina-c: done, I forgot to paste something from the previous PR

ocean_lib/data_provider/data_service_provider.py
calina-c: changed

ocean_lib/data_provider/base.py
calina-c: changed

alexcos20 (approved):
lgtm

PR #864, opened 2022-06-20
Delete c2d option from the datatokens flow and insert only in c2d README

### LINKED ISSUE #862: Reduce friction: READMEs don't use barge with C2D except C2D README
**Problem**

When people go through the READMEs, it takes about 10 min to get barge going. Most other steps take a few seconds.

This is friction for someone learning ocean.py, which in turn hurts Ocean adoption.

**Opportunity**

Fortunately there's an easy fix!

The bottleneck is in running the C2D part of barge. Barge without C2D only takes 1 min (compared to 10 min with).

Yet the only README that needs C2D is the C2D README, which is far down the list.

**Suggested Solution**

- Remove the C2D-specific stuff in the [first README](https://github.com/oceanprotocol/ocean.py/blob/v4main/READMEs/data-nfts-and-datatokens-flow.md), as shown below
- And reinsert it specifically into the C2D README

barge

mariacarmina:
Fixes #862  .

Changes proposed in this PR:

- delete c2d option from the datatokens flow `Download barge and run services` section and insert only in c2d README

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -28,11 +28,11 @@ Use the `--with-c2d` option when running barge in order to include the Compute-t
 ./start_ocean.sh --with-c2d
 ```
 When running barge with c2d enabled, use the following bash script to wait until
-the `artifacts` directory and `address.json` file are successfully created and c2d is fully deployed:
+c2d is fully deployed:
 ```console
 for i in $(seq 1 50); do
     sleep 5
-    [ -f "$HOME/.ocean/ocean-contracts/artifacts/ready" -a -f "$HOME/.ocean/ocean-c2d/ready" ] && break
+    [ -f "$HOME/.ocean/ocean-c2d/ready" ] && break
     done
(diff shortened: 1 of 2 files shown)


READMEs/c2d-flow.md
```diff
@@ -23,7 +23,18 @@ Let's go through each step.
 ## 1. Setup
 
 ### First steps
-
+Use the `--with-c2d` option when running barge in order to include the Compute-to-Data backend
+```console
+./start_ocean.sh --with-c2d
+```
+When running barge with c2d enabled, use the following bash script to wait until
```
calina-c, line 30: This should be split in 2: there should be a wait script only for ocean-contracts in datatokens flow. And here there should be a wait script for both ocean-contracts and ocean-c2d. It is explained right here, and the explanation should also be updated.

calina-c (changes requested):

calina-c (approved):

PR #868, opened 2022-06-24
Docker checksums

calina-c:
This PR just includes the workflow change and the unrelated ExampleConfig fix. If tests pass and no other changes (they should, I tested a huge chunk locally) are needed we can close this without changes.


calina-c:
Closing, this functionality change does not impact the current ocean.py code.

PR #872, opened 2022-07-05
Filecoin Integration

richardblythman:
Filecoin integration from core tech earmarks.
Complements PR to provider https://github.com/oceanprotocol/provider/pull/513

Changes proposed in this PR:

- Create StorageProvider class for uploading to decentralized storage
- Integrate HuggingFace Hub library and create HuggingFaceHub class for uploading to data science platforms
- Modify encrypt function to encrypt file content (as well as urls and ddos)
- Create an environmental variable for Filecoin urls
- Add class for Filecoin file structures
- Readme tutorial flow for uploading to storage and HuggingFace
- Add tests for uploading to and downloading from decentralized storage and HuggingFace


.github/workflows/libcheck.yml
```diff
@@ -51,9 +51,9 @@ jobs:
         run: |
           mkcodes --github --output tests/generated-readmes/test_{name}.{ext} READMEs
           py.test tests/test_readmes.py
-      - name: Slack notify via webhook
-        uses: up9cloud/action-notify@master
-        if: cancelled() == false
-        env:
-          GITHUB_JOB_STATUS: ${{ job.status }}
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      # - name: Slack notify via webhook
```
calina-c, line 54: why are these commented out in tests?

READMEs/store-ipfs-huggingface-flow.md
```diff
@@ -0,0 +1,113 @@
...
+export HF_KEY=<ADD HUGGINGFACE KEY HERE>
+
+```
+
+## 2. Alice encrypts a dataset and uploads to Filecoin and HuggingFace
+
+Then in the same python console:
+```python
+
+object_path = "ocean_lib/storage_provider/test/hello.txt"
```
calina-c, line 44: could you make it clear through text that this is just a convenient sample and the user can use any path they want? I am not concerned with the user being stupid, I am just concerned with bulk copying from the readme being a possible problem.

READMEs/store-ipfs-huggingface-flow.md
```diff
@@ -0,0 +1,113 @@
...
+with open(object_path, "rb") as f:
+    files = f.read()
+
+# Encrypt file(s) using provider
+from ocean_lib.data_provider.data_encryptor import DataEncryptor
```
calina-c, line 49: isn't there some helper for encrypting directly from the ocean object or assets object? If there isn't... there should be :) I know Trent likes keeping readmes very short, so that's what I would suggest.

READMEs/store-ipfs-huggingface-flow.md
```diff
@@ -0,0 +1,113 @@
...
+encrypt_response = DataEncryptor.encrypt(
+    files,
+    ocean.config.provider_url,
+)
+encrypted_files = encrypt_response.content.decode("utf-8")
+
+# Upload encrypted file to decentralized storage 
+from ocean_lib.storage_provider.storage_provider import StorageProvider
```
calina-c, line 57: same here, to avoid the import, please add a shorthand function in the Ocean class. It should be return a StorageProvider without the user needing to do the import and initialisation here. From my POV, it could also do the upload directly. Also, why are we decoding (in line 54) and then re-encoding encrypted_files (in line 60)?

READMEs/store-ipfs-huggingface-flow.md
```diff
@@ -0,0 +1,113 @@
...
+storage_provider = StorageProvider()
+storage_response = storage_provider.upload(encrypted_files.encode("utf-8"))
+cid = storage_response.json()['cid']
+
+from ocean_lib.structures.file_objects import IpfsFile
+ipfs_file = [IpfsFile(hash=cid)]
```
calina-c, line 63: we can also create a wrapper and combine all these instructions, e.g. a FilecoinFile type

READMEs/store-ipfs-huggingface-flow.md
```diff
@@ -0,0 +1,113 @@
...
+from ocean_lib.storage_provider.huggingface_hub import HuggingFaceHub
+huggingface = HuggingFaceHub()
+repo_url = huggingface.upload(encrypted_files.encode("utf-8"), object_name='hello', object_type='dataset')
+```
+
+## 3. Alice publishes an asset
+
+```python
+# Specify metadata and services, using the Branin test dataset
+metadata = {
```
calina-c, line 75: oh, so this PR is pretty old. We have added a lot of helpers to avoid these many imports and statements. Please look at the other readmes for the one-liner and use that instead.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -28,8 +28,9 @@
 from ocean_lib.models.datatoken import Datatoken
 from ocean_lib.ocean.util import get_address_of_type
 from ocean_lib.services.service import Service
+from ocean_lib.storage_provider.storage_provider import StorageProvider
 from ocean_lib.structures.algorithm_metadata import AlgorithmMetadata
-from ocean_lib.structures.file_objects import FilesType
+from ocean_lib.structures.file_objects import FilesType, IpfsFile
```
calina-c, line 33: it seems here you have added the imports, but not any functionality

ocean_lib/storage_provider/__init__.py
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright 2022 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+from .storage_provider import StorageProvider
```
calina-c, line 5: I don't think we need this init file with Python3.

ocean_lib/structures/file_objects.py
```diff
@@ -43,10 +44,31 @@ class IpfsFile(FilesType):
...
+            "hash": self.hash,
+            "url": self.url,
+            "gateway": self.gateway,
+        }
+
+    def get_download_url(self):
+        if not self.gateway:
+            raise Exception("No IPFS_GATEWAY defined, can not resolve ipfs hash.")
+
+        if self.gateway in [
```
calina-c, line 63: what is the purpose of this?

pytest.ini
```diff
@@ -26,4 +26,5 @@ env =
     D:PROVIDER_PRIVATE_KEY=0xfd5c1ccea015b6d663618850824154a3b3fb2882c46cefb05b9a93fea8c3d215
     D:ADDRESS_FILE=~/.ocean/ocean-contracts/artifacts/address.json
     D:OCEAN_NETWORK_URL=http://127.0.0.1:8545
+    D:IPFS_GATEWAY=https://api.web3.storage/upload
```
calina-c, line 29: this shouldn't be set for all tests, please set it with monkeypatch inside the needed tests.

setup.py
```diff
@@ -34,6 +34,7 @@
     "enforce-typing==1.0.0.post1",
     "json-sempai==0.4.0",
     "eciespy",
+    "huggingface_hub"
```
calina-c, line 37: please set the dependency version. Whatever is non-grounded now is probably due to dependency resolving issues, but if possible let's have the dependency in clear. (Or a comment explaining why it's not)

tests/test_readmes.py
```diff
@@ -20,10 +20,12 @@ def test_script_execution(script, monkeypatch):
         or "publish-flow" in script.name
         or "data-nfts-and-datatokens-flow" in script.name
         or "c2d-flow-more-examples" in script.name
+        or "store-ipfs-huggingface-flow" in script.name
     ):
         # developers.md skipped because it does not have end-to-end Python snippets, just console
         # data-nfts-and-datatokens-flow.md and publish-flow skipped because it they run as prerequisites for the others, so they are tested implicitly
         # c2d-flow-more-examples skipped because it can not be parsed separately from c2d-flow
+        # store-ipfs-huggingface-flow skipped because we do not want to make real API keys publicly available
```
calina-c, line 28: actually we can set up a secret in Github Actions, and we won't make any real API keys public. Simply use an env var in the python code that gets picked up by the readme tests (i.e. in the .md file).

calina-c (changes requested):

calina-c:
Nope, we can remove it. I wasn't aware about the DBS progress. @richardblythman disregard my comments, this can be closed

PR #890, opened 2022-08-08
Cleaning up content disposition header.

mariacarmina:
Changes proposed in this PR:

- sanitizing content disposition header while downloading files
- adding test for it.

### DIFF
--- a/ocean_lib/data_provider/base.py
+++ b/ocean_lib/data_provider/base.py
@@ -250,11 +250,28 @@ class DataServiceProviderBase:
 
     @staticmethod
     @enforce_types
+    def _sanitize_content(header: str) -> bool:
+        pattern = re.compile(r"\\|\.\.|/")
+        if pattern.findall(header):
+            return False
+        return True
+
+    @staticmethod
+    @enforce_types
     def _get_file_name(response: Response) -> Optional[str]:
         try:
-            return re.match(
-                r"attachment;filename=(.+)", response.headers.get("content-disposition")
-            )[1]
+            if DataServiceProviderBase._sanitize_content(
+                response.headers.get("content-disposition")
+            ):
+                return re.match(
+                    r"attachment;filename=(.+)",
+                    response.headers.get("content-disposition"),
+                )[1]
+            else:
+                logger.error(
+                    "Invalid content disposition format. It was not possible to get the file name."
+                )
+                return None
         except Exception as e:
             logger.warning(f"It was not possible to get the file name. {e}")
             return None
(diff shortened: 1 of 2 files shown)


ocean_lib/data_provider/base.py
```diff
@@ -248,13 +248,30 @@ def write_file(
         else:
             logger.warning(f"consume failed: {response.reason}")
 
+    @staticmethod
+    @enforce_types
+    def _sanitize_content(header: str) -> bool:
+        pattern = re.compile(r"\\|\.\.|/")
+        if pattern.findall(header):
```
calina-c, line 255: I think you can return not bool(pattern.findall(header))

ocean_lib/data_provider/base.py
```diff
@@ -248,13 +248,30 @@ def write_file(
...
+        return True
+
     @staticmethod
     @enforce_types
     def _get_file_name(response: Response) -> Optional[str]:
         try:
-            return re.match(
-                r"attachment;filename=(.+)", response.headers.get("content-disposition")
-            )[1]
+            if DataServiceProviderBase._sanitize_content(
```
calina-c, line 263: I don't understand. Are we validating or sanitizing? It seems we are validating, so please rename this function _validate_content_disposition and return True if valid, False otherwise.

Even more, we can have some generic header validation, e.g. _validate_headers(response), which for now only validates the content-disposition header.

I also suggest treating the denied case first, e.g.
```
if not _validate_content_disposition(response.headers.get("content-disposition"):
     logger.error(...)
     return None
return re.match(r"attachment ...)
```

ocean_lib/data_provider/test/test_base.py
```diff
@@ -0,0 +1,15 @@
...
+# Copyright 2022 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+from ocean_lib.data_provider.base import DataServiceProviderBase
+
+
+def test_sanitize_content_disposition():
+    header = "./../../my/relative/path"
+    res = DataServiceProviderBase._sanitize_content(header)
+    assert res is False
```
calina-c, line 11: you can `assert` res and `assert not res`, "assert ... is True" is superfluous. I would also cover the _get_file_name statements.

calina-c (changes requested):

ocean_lib/data_provider/test/test_base.py
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright 2022 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+from ocean_lib.data_provider.base import DataServiceProviderBase
+
+
+def test_sanitize_content_disposition():
```
calina-c, line 8: please add tests for _get_file_name as well, not only for _validate_content_disposition

calina-c (changes requested):

calina-c (approved):

PR #892, opened 2022-08-09
Integrate EIP 1559.

### LINKED ISSUE #888: Add EIP-1559 support
- [ ] detect if the blockchain supports the EIP-1559
- [ ] if it is supported, use the new fee structure

mariacarmina:
Fixes #888  .

Changes proposed in this PR:

- modified transactions in order to accept `maxPriorityFeePerGas` & `maxFeePerGas` instead of `gasPrice` for chains that support EIP 1559.
- updated helper functions to accept different networks, not only `ganache`
- tested the modifications with rinkeby & ganache.

### DIFF
--- a/ocean_lib/web3_internal/contract_base.py
+++ b/ocean_lib/web3_internal/contract_base.py
@@ -9,7 +9,6 @@ import os
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import requests
-import web3.gas_strategies.rpc
 from enforce_typing import enforce_types
 from eth_typing import ChecksumAddress
 from hexbytes import HexBytes
@@ -25,6 +24,7 @@ from ocean_lib.web3_internal.contract_utils import (
     get_contracts_addresses,
     load_contract,
 )
+from ocean_lib.web3_internal.utils import get_max_fee_per_gas, get_gas_price
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_overrides.contract import CustomContractFunction
 
@@ -169,16 +169,6 @@ class ContractBase(object):
             callback, timeout_callback=timeout_callback, timeout=timeout, blocking=wait
         )
 
-    @staticmethod
-    @enforce_types
-    def get_gas_price(web3_object: Web3) -> int:
-        if os.getenv("GAS_SCALING_FACTOR"):
-            return int(
-                web3_object.eth.gas_price * float(os.getenv("GAS_SCALING_FACTOR"))
-            )
-
-        return web3.gas_strategies.rpc.rpc_gas_price_strategy(web3_object)
-
     @enforce_types
     def send_transaction(
         self,
@@ -204,12 +194,12 @@ class ContractBase(object):
             "chainId": self.web3.eth.chain_id,
         }
 
-        fee_tx = ContractBase.get_max_fee_per_gas(web3=self.web3, tx=_transact)
+        fee_tx = get_max_fee_per_gas(web3=self.web3, tx=_transact)
         if fee_tx:
             _transact.update(fee_tx)
             _transact["gas"] = self.web3.eth.estimate_gas(_transact)
         else:
-            _transact["gasPrice"] = self.get_gas_price(self.web3)
+            _transact["gasPrice"] = get_gas_price(self.web3)
 
             gas_price = os.environ.get(ENV_GAS_PRICE, None)
             if gas_price:
@@ -223,18 +213,6 @@ class ContractBase(object):
             from_wallet.block_confirmations.value,
             from_wallet.transaction_timeout.value,
         ).hex()
-
-    @staticmethod
-    @enforce_types
-    def get_max_fee_per_gas(web3: Web3, tx: dict) -> Optional[dict]:
-        try:
-            history = web3.eth.fee_history(block_count=1, newest_block="latest")
-        except ValueError:
-            return None
-        tx["maxPriorityFeePerGas"] = web3.eth.max_priority_fee
-        tx["maxFeePerGas"] = web3.eth.max_priority_fee + 2 * history["baseFeePerGas"][0]
-
-        return tx
 
     @enforce_types
     def get_event_argument_names(self, event_name: str) -> Tuple:
@@ -265,7 +243,7 @@ class ContractBase(object):
         built_tx = _contract.constructor(*args).buildTransaction(
             {
                 "from": ContractBase.to_checksum_address(deployer_wallet.address),
-                "gasPrice": cls.get_gas_price(web3),
+                "gasPrice": get_gas_price(web3),
             }
         )
         if "chainId" not in built_tx:
--- a/ocean_lib/web3_internal/contract_base.py
+++ b/ocean_lib/web3_internal/contract_base.py
@@ -5,7 +5,6 @@
 
 """All contracts inherit from `ContractBase` class."""
 import logging
-import os
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import requests
@@ -18,13 +17,12 @@ from web3._utils.filters import construct_event_filter_params
 from web3.contract import ContractEvent, ContractEvents
 from web3.datastructures import AttributeDict
 from web3.exceptions import MismatchedABI, ValidationError
-from ocean_lib.web3_internal.constants import ENV_GAS_PRICE
 from ocean_lib.web3_internal.contract_utils import (
     get_contract_definition,
     get_contracts_addresses,
     load_contract,
 )
-from ocean_lib.web3_internal.utils import get_max_fee_per_gas, get_gas_price
+from ocean_lib.web3_internal.utils import get_gas_price
 from ocean_lib.web3_internal.wallet import Wallet
 from ocean_lib.web3_internal.web3_overrides.contract import CustomContractFunction
 
@@ -194,16 +192,8 @@ class ContractBase(object):
             "chainId": self.web3.eth.chain_id,
         }
 
-        fee_tx = get_max_fee_per_gas(web3=self.web3, tx=_transact)
-        if fee_tx:
-            _transact.update(fee_tx)
-            _transact["gas"] = self.web3.eth.estimate_gas(_transact)
-        else:
-            _transact["gasPrice"] = get_gas_price(self.web3)
-
-            gas_price = os.environ.get(ENV_GAS_PRICE, None)
-            if gas_price:
-                _transact["gasPrice"] = int(gas_price)
+        gas_tx = get_gas_price(web3_object=self.web3, tx=_transact)
+        _transact.update(gas_tx)
 
         if transact:
             _transact.update(transact)
@@ -241,16 +231,15 @@ class ContractBase(object):
 
         _contract = web3.eth.contract(abi=_json["abi"], bytecode=_json["bytecode"])
         built_tx = _contract.constructor(*args).buildTransaction(
-            {
-                "from": ContractBase.to_checksum_address(deployer_wallet.address),
-                "gasPrice": get_gas_price(web3),
-            }
+            {"from": ContractBase.to_checksum_address(deployer_wallet.address)}
         )
+
         if "chainId" not in built_tx:
             built_tx["chainId"] = web3.eth.chain_id
 
-        if "gas" not in built_tx:
-            built_tx["gas"] = web3.eth.estimate_gas(built_tx)
+        if "gas" not in built_tx or "gasPrice" not in built_tx:
+            gas_tx = get_gas_price(web3_object=web3, tx=built_tx)
+            built_tx.update(gas_tx)
 
         raw_tx = deployer_wallet.sign_tx(built_tx)
         logging.debug(
--- a/ocean_lib/web3_internal/contract_base.py
+++ b/ocean_lib/web3_internal/contract_base.py
@@ -204,19 +204,11 @@ class ContractBase(object):
             "chainId": self.web3.eth.chain_id,
         }
 
-        try:
-            history = self.web3.eth.fee_history(block_count=1, newest_block="latest")
-
-            _transact["gas"] = self.web3.eth.estimate_gas(_transact)
-            fee_tx = ContractBase.get_max_fee_per_gas(
-                web3=self.web3, tx=_transact, history=history
-            )
+        fee_tx = ContractBase.get_max_fee_per_gas(web3=self.web3, tx=_transact)
+        if fee_tx:
             _transact.update(fee_tx)
-        except ValueError as e:
-            assert (
-                e.args[0]["message"] == "Method eth_feeHistory not supported."
-            ), f"Another error occurred: {e.args[0]}"
-
+            _transact["gas"] = self.web3.eth.estimate_gas(_transact)
+        else:
             _transact["gasPrice"] = self.get_gas_price(self.web3)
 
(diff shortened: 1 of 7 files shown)


mariacarmina:
Rinkeby success logs
```
INFO     ocean:ocean_assets.py:449 Successfully created datatoken with address 0x53c7Ce78A26861e744E620a22376b091f6cCA188.
INFO     ocean_lib.data_provider.data_encryptor:data_encryptor.py:58 Asset urls encrypted successfully, encrypted urls str: 0x04c400f33a71b5632bedb88fec2e6c48bf03d510be4ce8a859c760b3ccdec77667699095052a09f760d520d2b25ddf643777d0d9137d85fa2f22289c748281d00f5ba6f1d63a90c5952e9333029fffda52e82eeed492e687bf71c7febda428da1d03c29d96a463e40eac7e2012c8924a33d53ad893c94f886763d4498fa6051ba03379b31e77a29b983e1dd051f879682145176bd86f92ca460d7f0349f95037c336d69e7b5923068d776dbace811459dabc459abefae1b510397feacc651a2883e3b426063c4b17e5b5a847e90690943daec1ca0ed0a8b7aec4432dccbef467a87da3f4ac67034b971c659da674240dba1d0ede24361861f086de4fa20b3008690f69bdd2dbfefd290251915f0f029e469efc71e59934736de9a81d9b7f942de333d24e5a2bd47850a86308cbd89a9da489e7e9cd939ec16b64a650f7ce76829fc4c1a8860da28bc433ad33530784a29456ed0bf991bd8130b8cdda4d5b6d59b414e4fa967f1de79ae06def51cd7907b3922c7718eef0b9fd, encryptedEndpoint https://v4.provider.rinkeby.oceanprotocol.com/api/services/encrypt
INFO     ocean_lib.data_provider.data_encryptor:data_encryptor.py:58 Asset urls encrypted successfully, encrypted urls str: 0x0414333e078f0b60cb82c1b47ce8af6bb9883b6923ee729e1c9f4c8c29f4d38d103637305c9368e76d6b93fead74f94a332e7544677fd449ea1578205577d6e091e54920d7a6e7f43fd3aec3db7786011002e5ae2f30f06e302a7ba1447bd18c5149f57f48494b124f461fafbe8fb9df15a763fa204f6fb7fb13bb60f03e6caa69fcb8ac41f2d7eda3f0eed4b98ed7dd2e07e5878371f9c66fed3c70b66623c7ea8be2da295d69946a7e1b122908aa5fbb6f1dda17fce1bf60ac1c6f3da3b13e8e69d514633029300b036f5f36819635ed3b62c93498761d4c7585f7bf655ab3b2fa9e300b86d1a26c8e6cc321225bc285b6d5a37c5bfcd9e914a910fce24d16930611d03fcba2bd036349f1b686965b63e3949c6fe44b5a44b2e6ffe2da8e59f466f13c8a337bf015d1d9871c4d810cf5e147b07819a4705b8996d5a3a6828188aa7a9eb6d357ad3ab36e5688d2fe7de3067f044ff9afffad6dcc5055ff410e92a042fec7cde43981852ca9eb43dbee438f0b3bbb1416202ca04d3b627ddfcc12e4d97a88bfd2bebfac9b5c590a512f84ec7970e8a1a00cf3aaf18c64ec38f27e63601030ba41ef6ad3279edba438b35807a00f81a9447b15be65a12c510e4724a4d3703026f938a8b32c6ccf7482b65d6303e9b975c2f07f725ed75652ceeef390f2f789b442584f22fa606def76684dbd2776032c137c3b1af98f5e631048d8e6960b7f879b26b07d79a1c041d4323ce3fb83041e940e4146bdf80386304831b840786e076cae839f0fe30903d6570a6dd20e79d80fdeb6d625a82227f98e980f8dee2a6f7561d4e65827e71089b240be9974441fc49a30551c623bc7e033e3c37d43ba7342572266b61f1cbe76c64b1910e7f7ea6141ca53676ddfc5c4e753338af8f0ad7c877a19238a944a76f5c9d171bc4fe995d668f30aeb9811aeee579718b8b30d2678dce5ebc4144979d60c5d54722689fc560efba8f940ee30fafb78e88b226c86da654c761e1ee3b652fd92f3a318426da73edec496a3eec06eeb708ca7330a80ea7e4fb192a4b331be3e8eb587ea092924a1b5bb77adb70c039be049cc4250fb6f6cc11f78ce154eb98fd14ff2e55bae28e46f531718dcf9eec9f569a7ba0a15fd6595e47785d4a74a3c14b7e83183940430b41a89d0965df23342b85293103b189df38c45bd315a3019947eef87ce04053861b9fb9b94e0f960518cdac4bf725b69ebb46310c6233bdbc4899131aa6461f07f06d68502c20c989e948c3d625d95e3e25f1fc87c9628941efa08df209fc60ac73f8eed8d86baef1bfc1ec27695e0e499337b547b1ff305fa9e6f38228c4e3606b87d58523b2ba98779a19c3b0fab7b18e8bc7a2fc0484381572fff9ff67c94094701d1113f4a695e82e89e328d2a91f859b6237c48d56f95020f2e086685ca73245514660075682b4a3aea6cda5872fe2b3176a44b089ea621442e4ad7d22407f4f2d0c8712019bbbacce162d625e17459fe6eed1f5e1b0e65c3b2cfe6eaa55dc80026ca44601aa2a37d67c25e4dd7d787c8a2, encryptedEndpoint https://v4.provider.rinkeby.oceanprotocol.com/api/services/encrypt
INFO     ocean_lib.data_provider.data_service_provider:data_service_provider.py:82 Service initialized successfully initializeEndpoint https://v4.provider.rinkeby.oceanprotocol.com/api/services/initialize
```

ocean_lib/web3_internal/contract_base.py
```diff
@@ -174,12 +171,11 @@ def subscribe_to_event(
 
     @staticmethod
     @enforce_types
-    def get_gas_price(web3) -> int:
+    def get_gas_price(w3: Web3) -> int:
```
calina-c, line 174: please use the web3 naming, not w3, we sometimes search through the codebase and it helps to have the same variable name

ocean_lib/web3_internal/contract_base.py
```diff
@@ -199,16 +195,32 @@ def send_transaction(
...
+                "from": ContractBase.to_checksum_address(from_wallet.address),
+                "account_key": from_wallet.key,
+                "chainId": self.web3.eth.chain_id,
+            }
+            _transact["gas"] = self.web3.eth.estimate_gas(_transact)
+            _transact["maxPriorityFeePerGas"] = self.web3.eth.max_priority_fee
+            _transact["maxFeePerGas"] = (
+                self.web3.eth.max_priority_fee + 2 * history["baseFeePerGas"][0]
+            )
+        except ValueError as e:
```
calina-c, line 211: will the ValueError only be encountered in fee_history? Then I suggest limiting the try to that line only. And since "account_key", "from", and "chain_id" have the same value, no matter the method chosen, please assign those first, outside the choice.

It seems to me there are in fact 3 options: when the env gas_price is set, when we are using get_gas_price and when we use the pricing history. Please make sure that all these options are clear and readable and there's a clear cause->consequence relation when reading the code. This implies better naming ("get_gas_price" is too generic now), extraction of the common keys and other small readability tweaks.

ocean_lib/web3_internal/wallet.py
```diff
@@ -123,20 +123,35 @@ def sign_tx(
             )
         else:
             nonce = Wallet._get_nonce(self.web3, account.address)
+            if not gas_price:
+                try:
+                    history = self.web3.eth.fee_history(
+                        block_count=1, newest_block="latest"
+                    )
+                    tx["type"] = "0x2"
+                    tx["maxPriorityFeePerGas"] = self.web3.eth.max_priority_fee
```
calina-c, line 132: this code seems duplicated. Could we extract this part into a different function and create the transaction keys separately? We could reuse it in contract_base as well, and help on the readability issues I just mentioned over there.

ocean_lib/web3_internal/wallet.py
```diff
@@ -123,20 +123,35 @@ def sign_tx(
...
+                    tx["maxFeePerGas"] = (
+                        self.web3.eth.max_priority_fee + history["baseFeePerGas"][0] * 2
+                    )
+                    tx["gas"] = GAS_LIMIT_DEFAULT
+                except ValueError as e:
+                    assert (
+                        e.args[0]["message"] == "Method eth_feeHistory not supported."
+                    )
+
+                    from ocean_lib.web3_internal.contract_base import ContractBase
```
calina-c, line 142: why is thihs import here and not at the top of the file?

calina-c (changes requested):

ocean_lib/web3_internal/wallet.py
mariacarmina: It will generate a circular import error if I put at the top of the file.

ocean_lib/web3_internal/contract_base.py
mariacarmina: I let `web3` in the first place, but it couldn't be distinguished between the web3 package from the import, so it creates a confusion locally. Can I rename to `web3_object`?

ocean_lib/web3_internal/wallet.py
mariacarmina: I can create a separate function for EIP-1559 transactions.

ocean_lib/web3_internal/contract_base.py
```diff
@@ -199,16 +197,31 @@ def send_transaction(
...
+            fee_tx = ContractBase.get_max_fee_per_gas(
+                web3=self.web3, tx=_transact, history=history
+            )
+            _transact.update(fee_tx)
+        except ValueError as e:
+            assert (
+                e.args[0]["message"] == "Method eth_feeHistory not supported."
+            ), f"Another error occurred: {e.args[0]}"
+
+            _transact["gasPrice"] = self.get_gas_price(self.web3)
```
calina-c, line 220: why are we calculating the gas price even if we have an env defined?

ocean_lib/web3_internal/contract_base.py
calina-c: we do not assert error messages like this, please wrap the value error only where you need it. It is generally bad practice in the code to assert an exception's text. Also, why is history passed in as a parameter? Could it be moved inside the function? Then you could treat exceptions there, and simply return None if the calc could not be calculated.

please also look into the naming suggestions I gave previously. We now have a few generic "get gas price" options, but they should reflect their meaning. One is based on history, one is based on the env, one is what? Legacy? We should clearly state this.

ocean_lib/web3_internal/wallet.py
```diff
@@ -123,20 +123,35 @@ def sign_tx(
...
+                    history = self.web3.eth.fee_history(
+                        block_count=1, newest_block="latest"
+                    )
+                    tx["type"] = "0x2"
+                    fee_tx = ContractBase.get_max_fee_per_gas(
+                        web3=self.web3, tx=tx, history=history
+                    )
+                    tx.update(fee_tx)
+                    tx["gas"] = GAS_LIMIT_DEFAULT
+                except ValueError as e:
```
calina-c, line 139: the duplication here will be fixed when you move history inside the get_max_fee_per_gas function.

calina-c (changes requested):
Still a lot of duplication and the exception handling needs much improvement.

ocean_lib/web3_internal/contract_base.py
mariacarmina: I call the function get_gas_price so it includes the env defined if exists.

ocean_lib/web3_internal/contract_base.py
mariacarmina: `get_gas_price` is clearly for transactions that do not support EIP 1559. I wanted to include the logic of retrieving an envvar if exists for gas price calculation and also a fallback to a web3 method that evaluates the gas price for that network. I didn't see a naming suggestion for it and I think it fits well, maybe `calculate_gas_price` sounds better? 
`get_max_fee_per_gas` includes just the specific fields for EIP 1559 transactions which I think is clear what it does.
A good idea is to also provide docstrings for these functions.
I modified `get_max_fee_per_gas` the function to treat the exception inside of it and to avoid code duplication.

ocean_lib/web3_internal/wallet.py
```diff
@@ -123,20 +123,26 @@ def sign_tx(
             )
         else:
             nonce = Wallet._get_nonce(self.web3, account.address)
+            from ocean_lib.web3_internal.contract_base import ContractBase
```
calina-c, line 126: you can move the needed function somewhere in utils to avoid the import error

ocean_lib/web3_internal/wallet.py
```diff
@@ -123,20 +123,26 @@ def sign_tx(
...
+
+            if not gas_price:
+                fee_tx = ContractBase.get_max_fee_per_gas(web3=self.web3, tx=tx)
+                if fee_tx:
```
calina-c, line 130: this still seems duplicated to what I'm seeing in contract_base. Could we improve it even more?

ocean_lib/web3_internal/contract_base.py
```diff
@@ -219,6 +224,17 @@ def send_transaction(
...
+    @staticmethod
+    @enforce_types
+    def get_max_fee_per_gas(web3: Web3, tx: dict) -> Optional[dict]:
+        try:
+            history = web3.eth.fee_history(block_count=1, newest_block="latest")
+        except ValueError:
+            return None
+        tx["maxPriorityFeePerGas"] = web3.eth.max_priority_fee
+        tx["maxFeePerGas"] = web3.eth.max_priority_fee + 2 * history["baseFeePerGas"][0]
+        return tx
```
calina-c, line 236: please add an empty line above the return statement

tests/resources/helper_functions.py
```diff
@@ -48,7 +48,7 @@ def get_example_config():
 def get_address_of_type(
     config: Config, address_type: str, key: Optional[str] = None
 ) -> str:
-    addresses = get_contracts_addresses(config.address_file, _NETWORK)
+    addresses = get_contracts_addresses(config.address_file, config.network_name)
```
calina-c, line 51: other than this change, are there any tests we can come up with?

calina-c (changes requested):
Looks good! I left a few small comments left and asked for tests

ocean_lib/web3_internal/wallet.py
mariacarmina: Here I just call the function, but there are different fields to fill or different conditions.

ocean_lib/web3_internal/wallet.py
mariacarmina: I have tried right now to move into web3_internal/utils and also in web3_internal/transactions (because it fits better in this folder since it's related to web3) and it's still complaining about the conftest (circular import errors).

calina-c: I'm sure we can find a way. It is a basic function not using any higher level components, so we should be able to move it somewhere safe.

ocean_lib/web3_internal/wallet.py
calina-c: I think it will become more clear after you refactor the handling in contract base and moving the function separately. You can handle this comment last.

calina-c (approved):
Amazing work! Clean and uniform, just what we need.

PR #898, opened 2022-09-06
Upgrade ocean contracts

calina-c:
Closes #905 as well.

PR #903, opened 2022-09-06
Towards #900 - UX tweaks for READMEs

### LINKED ISSUE #900: UX tweaks for READMEs
Changes for data-nfts-and-datatokens-flow.md:
- [x] Avoid turning on non-essential `start-ocean.sh` services. https://github.com/oceanprotocol/ocean.py/pull/903/commits/8aa0aad2cb5b4965c15665c91d4cbbd10e6d780f
  - Why: faster loading = better UX. Essential = all READMEs still work (can ignore C2D, it has its own setup). 
  - [df-py example of start-ocean.sh with fewer services](https://github.com/oceanprotocol/df-py#prerequisites): `./start_ocean.sh --no-aquarius --no-elasticsearch --no-provider --no-dashboard --with-thegraph`
- [x] If start-ocean is fast enough from above, then remove "Wait until the artifacts directory.." line. If not fast enough, then (a) specify which console to run it in. It's not barge's console, since that's already busy with barge; (b) let the user know that "-f" means it's a bash-style query for the existence of a file [link](https://stackoverflow.com/questions/30825762/what-does-f-mean-in-bash).  Result: it was fast enough; the commit reflects it: https://github.com/oceanprotocol/ocean.py/pull/903/commits/c063e2232f7e99d8868490f099c3f0876ac7930f
- [x] Change "Publish an NFT token" --> "Publish NFT". Why: "token" is redundant https://github.com/oceanprotocol/ocean.py/pull/903/commits/9c5d09dc841abe17a1482bb41b4660e6d433194a
- [x] Change `NFTToken1` --> `NFT1`. Why: "token" is redundant. Also: change in other READMEs as needed - do check. https://github.com/oceanprotocol/ocean.py/pull/903/commits/25ed4f011341095affc369f371283c44fc6db497
- [x] Move (envvar setting, py code) wrt OCEAN from this README (introduced [here](https://github.com/oceanprotocol/ocean.py/pull/908)), to the _marketplace_ README where it's first used. [Details here](https://github.com/oceanprotocol/ocean.py/issues/901#issuecomment-1239436609). https://github.com/oceanprotocol/ocean.py/pull/903/commits/2c3c9c2cfd0787509d6d5167659577d02494d97b
- [x] Move setup of Alice's wallet into a special "Python setup" section. https://github.com/oceanprotocol/ocean.py/pull/903/commits/2c3c9c2cfd0787509d6d5167659577d02494d97b

Changes for publish-flow.md:
- [x] Change reference to "2.1 create data NFT" --> "2.1 get Alice wallet". Also: change in other READMEs as needed - do check. https://github.com/oceanprotocol/ocean.py/pull/903/commits/ec9f9092e4f797d810130759eb1b2131c0c2e64f
- [x] Change "automatically deploys a data NFT token using the ERC20Template" --> "automatically deploys a data NFT and datatoken using the ERC721Template and ERC20Template, respectively". Also: reconcile the rest of the paragraph too. https://github.com/oceanprotocol/ocean.py/pull/903/commits/a2d399b4434dd2d5fbec891d4926f8483150c2f0
- [ ] Call to `asset = ocean.assets.create()` passes in _many_ parameters that don't need passing in. To fix: remove passing in parameters that already get appropriate values by default
- [ ] The README ends kind of abruptly, it's jarring for the user. We should give the user some idea what was created. To do: print out data NFT name & symbol, and datatoken name & symbol, similar to in the [first README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/data-nfts-and-datatokens-flow.md#3-tips-and-tricks)
- [ ] We need to give the user a feel for what else an the Ocean `asset` can do. A basic way is `dir(asset)`. Better yet: give a few more specific calls; and point to the [code](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/ocean/ocean_assets.py) that shows the whole interface.

Changes for consume-flow.md:
- [x] Streamline refs to creating data NFT etc. https://github.com/oceanprotocol/ocean.py/pull/903/commits/1412fe82b9a02270d06c1abf62e9e0e95a8f0bf7
-- [x] In "tips and tricks" section, "encryption" subsection: it's redundant with the encryption section in the previous README. So, delete that whole subsection here. https://github.com/oceanprotocol/ocean.py/pull/903/commits/ae447e3f4e3957ce49fa84859199927b358dee9e
- [x] Remove lines `from ocean_lib.web3_internal.constants import ZERO_ADDRESS` and `fee_receiver = ZERO_ADDRESS  # could also be market address', they're unused. https://github.com/oceanprotocol/ocean.py/pull/903/commits/164871b8eb3869d95ab7697b8a81f902e71bfa49
- [ ] In this README, the `asset` object was created by Alice previously. (Of class `ocean_lib.assets.asset.Asset`.) But Bob wouldn't normally have access to that. We need to show how Bob would normally get the `asset` object: `asset = ocean.assets.resolve(did)` (like how C2D README does it)

Changes for fixed_rate_exchange.py: 
- [x] Section 3 has Alice minting datatokens. Section 4 is labeled "Bob buys at fixed rate datatokens"; but it does (a) Alice creating the market in addition to (b) Bob buying. To fix, move Alice creating market to Section 3, and relabel Section 3 to "Alice creates the market". https://github.com/oceanprotocol/ocean.py/pull/903/commits/2c3c9c2cfd0787509d6d5167659577d02494d97b 
- [x] This is the first time Bob needs to spend OCEAN. If he doesn't have it (due to config issues etc), it will be hard for the user to tell. We already have a check if Bob doesn't have ETH: `assert ocean.web3.eth.get_balance(bob_wallet.address) > 0, "need ganache ETH"`. To fix, add a similar check for OCEAN: `OCEAN_token.balanceOf(bob_wallet.address) > 0`. https://github.com/oceanprotocol/ocean.py/pull/903/commits/2c3c9c2cfd0787509d6d5167659577d02494d97b
- [x] A few more tweaks in https://github.com/oceanprotocol/ocean.py/pull/903/commits/2c3c9c2cfd0787509d6d5167659577d02494d97b
- [ ] Share the interface for the object resulting from call to `ocean.get_fixed_rate()`. It's a `FixedRateExchange` object; [here's the interface code](https://github.com/oceanprotocol/ocean.py/blob/9806caf5dc53ee4d48c3b60ece503f146e93db43/ocean_lib/models/fixed_rate_exchange.py#L44). 

In dispenser-flow.md:
- [x] Streamline refs to creating data NFT etc. https://github.com/oceanprotocol/ocean.py/pull/903/commits/6e9026b48213f80d91e694ed0fbfd638ed246360
- [ ] it kinda looks like we grab a dispenser object, then we create it? Weird. What's actually going on? Explain it.
- [ ] tiny tweak: `## Create dispenser` should be above `max_amount = `
- [ ] section header `3. Dispenser creation & activation` needs to be larger font (h1 not h2?)
- [ ] there aren't enough `# ___` comments inside the code to let the reader know what's happening. Add 2-3, it will help a lot
- [ ] in call to `datatoken.create_dispenser()`, the user won't know what the input parameters mean: `with_mint`, `allowed_swapper`, the ZERO_ADDRESS value for `allowed_swapper`
- [ ] there's no real info on what `dispenser_status` list holds. It asserts that entry 0 and 2 are True, but what does that mean?
- [ ] in calling `dispenser.dispense_tokens()`, can anyone dispense? 
- [ ] it asserts that Alice's initial balance is zero, and that it's `max_amount` after dispense. But if someone runs this README more than once, it will complain. To fix: remove the first assert, and change the second assert to `initial_balance + max_amount`
- [ ] so that user can know what else dispenser can do: maybe give one or two more queries; and point to its implementation in code

In c2d-flow.md:
- [x] Re "When running barge with c2d enabled, use the following bash script to wait until c2d is fully deployed", tell the user which console to run in. The barge console is already busy, so it's likely the other one. https://github.com/oceanprotocol/ocean.py/pull/903/commits/02c991835f0a7427ebee0872a680c7b2cd29c412
- [x] Re bash script, let the user know that "-f" means it's a bash-style query for the existence of a file [link](https://stackoverflow.com/questions/30825762/what-does-f-mean-in-bash) https://github.com/oceanprotocol/ocean.py/pull/903/commits/02c991835f0a7427ebee0872a680c7b2cd29c412
- [x] Move `pip install numpy matplotlib` to right before we plot. Its motivation will be more obvious then. https://github.com/oceanprotocol/ocean.py/pull/903/commits/02c991835f0a7427ebee0872a680c7b2cd29c412
- [x] It has "Download barge and run services". But that's incorrect, since we are running different services here, as it specifies above. Fix accordingly. https://github.com/oceanprotocol/ocean.py/pull/903/commits/02c991835f0a7427ebee0872a680c7b2cd29c412

DEFER TO LATER:
- In consume-flow.md: There are calls to `ocean.assets` such as `ocean.assets.pay_for_access_service(asset=_, ..)`. With proper OO, the call would be to `asset` object itself, not some global Ocean interface. Then the flow would be: pass in the id, get the `asset` object (instance of `Asset`). Then interact directly with object. (The cleanest way: use Brownie!)
- In fixed_rate_exchange.md: The call to `ocean.fixed_rate_exchange.buy_dt()` needs to know `consume_market_swap_fee_address` and `consume_market_swap_fee_amount`. The README knows because of god-level information. But in general how does a user possibly get this? Also: with proper OO, the call would be to the consume market itself, not some global Ocean interface. Then the flow would be: pass in the id, get the `consume_market` object (instance of `FixedRateExchange`). Then interact directly with the object. (The cleanest way: use Brownie!)

trentmc:
Towards #900.

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -10,11 +10,11 @@ This quickstart describes a C2D flow.
 Here are the steps:
 
 1. Setup
-2. Alice publishes data asset
+2. Alice publishes dataset
 3. Alice publishes algorithm
 4. Alice allows the algorithm for C2D for that data asset
 5. Bob acquires datatokens for data and algorithm
-6. Bob starts a compute job using a free C2d environment (no provider fees)
+6. Bob starts a compute job using a free C2D environment (no provider fees)
 7. Bob monitors logs / algorithm output
 8. Tips and tricks
 
@@ -22,47 +22,74 @@ Let's go through each step.
 
 ## 1. Setup
 
-### First steps
+### Prerequisites
+
+From [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md), do:
+- [x] Setup : Prerequisites
+
+
+### Download barge and run services
+
+Here, we use barge's "Compute-to-Data" backend of barge via its `--with-c2d` option. The steps below detail how.
+
+If you have a console with barge running, turn off barge by pressing `Ctrl-C` and waiting for it to shut down.
+
+Then, in the barge console:
+```console
+# Grab repo (if needed)
+git clone https://github.com/oceanprotocol/barge
+cd barge
+
+# Clean up old containers (to be sure)
+docker system prune -a --volumes
+
+# Run barge: start Ganache, Provider, Aquarius; deploy contracts; update ~/.ocean
+# (For speed, we turn off components we don't need for quickstarts, like IPFS)
+./start_ocean.sh --with-c2d-option --no-ipfs --no-dashboard --skip-subgraph-deploy
+```
+
 Use the `--with-c2d` option when running barge in order to include the Compute-to-Data backend
 ```console
 ./start_ocean.sh --with-c2d
 ```
-When running barge with c2d enabled, use the following bash script to wait until
-c2d is fully deployed:
-```console
-for i in $(seq 1 50); do
-    sleep 5
-    [ -f "$HOME/.ocean/ocean-c2d/ready" ] && break
-    done
-```
-To get started with this guide, please refer to [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md) and complete the following steps :
-- [x] Setup : Prerequisites
-- [x] Setup : Download barge and run services
+
+### Install the library
+
+From [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md), do:
 - [x] Setup : Install the library from v4 sources
 
 ### Install extra libraries
 
-This example uses c2d to create a regression model. In order to visualise it or manipulate it, you also need some dependencies.
-
-In your project folder, in this case my_project from `Setup : Install the library` in First Steps, run the following command:
-
+This example uses C2D to create a regression model. We'll use the library `matplotlib` to visualize it. So, in the same console:
 ```console
 pip install numpy matplotlib
 ```
 
 ### Set envvars
 
-Set the required enviroment variables as described in [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md):
+From [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md), do:
 - [x] Setup : Set envvars
 
-## 2. Alice publishes a Data NFT
+### Wait for barge C2D to be ready.
 
-In your project folder (i.e. my_project from `Install the library` step) and in the work console where you set envvars, run the following:
+It can take >10 min for barge with C2D to be ready. We can tell it's ready when the file `$HOME/.ocean/ocean-c2d/ready` exists. Here's a convenient way to wait until it's ready. In the console:
 
-Please refer to [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md) and complete the following steps :
-- [x] 2.1 Create a data NFT
+```console
+#loop until target file exists
+for i in $(seq 1 50); do
+    sleep 5
+    [ -f "$HOME/.ocean/ocean-c2d/ready" ] && break
+    done
+```
 
-## 3. Alice publishes a dataset
+When it's done, we can proceed to the next step.
+
+### Setup in Python
+
+From [data-nfts-and-datatokens-flow](data-nfts-and-datatokens-flow.md), do:
+- [x] Setup : Setup in Python
+
+## 2. Alice publishes dataset
 
 In the same python console:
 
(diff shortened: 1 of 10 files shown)


mariacarmina:
`tect_data_service_provider.py::test_get_root_uri` is outdated and will be fxed in @calina-c 's PR #898. The second one, `test_compute_flow.py::test_compute_update_trusted_algorithm` fails when the compute job is taking too long to run, when this happens, we usually re-run the tests on CI, but it doesn't happen everytime.

trentmc:
Please put these comments in the appropriate issue: #905.

trentmc:
> `tect_data_service_provider.py::test_get_root_uri` is outdated and will be fxed in @calina-c 's PR #898 ([440a8d5](https://github.com/oceanprotocol/ocean.py/commit/440a8d5478e756069face84da005e3fa70e17790)). The second one, `test_compute_flow.py::test_compute_update_trusted_algorithm` fails when the compute job is taking too long to run, when this happens, we usually re-run the tests on CI, but it doesn't happen everytime.

Please put these comments in the appropriate issue: #905.

PR #913, opened 2022-09-09
Fix #909 - C2D README use remote

### LINKED ISSUE #909: In ocean.py C2D README, barge*C2D has issues. Solve by: README uses remote not local networks
[ocean.py C2D README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/c2d-flow.md) uses barge which has Ocean Provider running in barge C2D mode.

At the beginning of that README, it asks us to barge running in C2D mode: `./start_ocean.sh --with-c2d`

It fails with the error of: "kindcluster_1      | ERROR: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?"

**TODO to solve this issue: Switch ocean.py C2D from local to remote**

Why: more practical solution for C2D. [Details here](https://github.com/oceanprotocol/ocean.py/issues/909).

trentmc:
Fixes #909


calina-c:
@trentmc Shouldn't we keep a local c2d readme as well? All readmes should have a local version due to dependency issues, and I think we should discuss a solution to mitigate the differences between barge versions and remotely deployed versions. E.g. we could have a perfectly working Provider version on rinkeby, but we could have updates on barge that we are completely new and different. This renders the readme tests useless, since they will run on a "frozen" version set in the remote branch (Provider, Aqua), but will not trigger when updating the remote networks with new versions. This is opposed to the current flow where we always test the newest changes from barge and then adapt the remote networks, after making sure everything works.

Not sure if I'm explaining this right, but we can discuss more when available.

trentmc:
> @trentmc Shouldn't we keep a local c2d readme as well? 

No, not in this case. It means docker-inside-docker, which gives a Pandora's box of errors. (E.g. it doesn't work on my relatively vanilla ubuntu 22 setup.)

> All readmes should have a local version due to dependency issues

I see that this will change over time: we will be having more remote-only readmes. That's where the practical usage lies. We'll keep simple local readmes and some remote counterparts, but let's not force 1:1.

PR #934, opened 2022-09-18
Fix #933 - Need a simple remote README

### LINKED ISSUE #933: Need a simple remote README

All of our READMEs are on ganache. The C2D one is in progress to change to remote, via #909, good. But remote _and_ C2D at once is a big jump for users. It will be better to walk them down the garden path, first with remote, then with C2D (with remote).

A README that does something simple, on a remote network.

**TODOs:**
- [ ] In the [PR](https://github.com/oceanprotocol/ocean.py/pull/913) for remote C2D README, there's also a simple README too. [Here](https://github.com/oceanprotocol/ocean.py/pull/913/files#diff-72972fbc5633f22a62d471e94a0912b091703f9bef40fb580c1d430410204ee1). Move it from that PR into here. 

Maybe also helpful: new remote-network unit test: [tests/integration/remote/test_mumbai.py](https://github.com/oceanprotocol/ocean.py/blob/main/tests/integration/remote/test_mumbai.py), from #923.

trentmc:
Fixes #933

### DIFF
--- a/READMEs/simple-remote.md
+++ b/READMEs/simple-remote.md
@@ -1,0 +1,100 @@
+<!--
+Copyright 2022 Ocean Protocol Foundation
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Quickstart: _Remotely_ Publish Data NFT & Datatoken
+
+This quickstart is like the [simple-flow](READMEs/data-nfts-and-datatokens-flow.md), but uses [remote services and Mumbai testnet](https://docs.oceanprotocol.com/core-concepts/networks#mumbai).
+
+Let's go through each step.
+
+## 1. Setup
+
+### Prerequisites & installation
+
+From [simple-flow](data-nfts-and-datatokens-flow.md), do:
+- [x] Setup : Prerequisites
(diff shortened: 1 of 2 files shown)


READMEs/simple-remote.md
```diff
@@ -0,0 +1,100 @@
...
+import os
+from ocean_lib.web3_internal.wallet import Wallet
+alice_private_key = os.getenv('TEST_PRIVATE_KEY1')
+alice_wallet = Wallet(ocean.web3, alice_private_key, config.block_confirmations, config.transaction_timeout)
+```
+
+
+## 2. Publish Data NFT & Datatoken
+
+From here on, the code's is the same as local.
```
calina-c, line 86: can we add a sample here for CI purposes, if possible? I know it would be a duplicate and a DRY violation, but it would help with automated testing. Alternately let's have remote-based tests. We should explore that option as well: a remote-based test suite running periodically, like we run our lib viability/compatibility test.

What do you think, @trentmc?

trentmc:
Fix #933 - Need a simple remote README

PR #953, opened 2022-09-21
Remove config

### LINKED ISSUE #938: Streamline Config: no config.ini file, no dict, just envvars. **Will help on CI failures**
Instantiating a `Config` object is a mess by having to support (a) `config.ini` file (b) input dict and (c) envvars just to set some config values. The config code itself is pretty messy, and error prone to maintain. We keep running into issues, as listed below, e.g. _many_ random CI failures, which are mostly due to envvar <> config issues

It will be way cleaner to support _just_ envvars. We can test + debug far more efficiently.

TODO for this issue: change Config (and related) to be _just_ envvars. Deprecate config.ini and input dict, and related cases

Once we solve this issue, we can go back and solve the following issues very quickly:
- [#927](https://github.com/oceanprotocol/ocean.py/issues/927): Config needs both network and network_name; it shouldn't
- [#931](https://github.com/oceanprotocol/ocean.py/issues/931): Config needs address_file specified for remote networks; it shouldn't (except for custom network)
- [#930](https://github.com/oceanprotocol/ocean.py/issues/930): Why does ocean.py source code hold a config.ini? It's error-prone
- [#905](https://github.com/oceanprotocol/ocean.py/issues/905): CI test randomly fails: test_compute_flow.py::test_compute_update_trusted_algorithm
- [#943](https://github.com/oceanprotocol/ocean.py/issues/943): CI test randomly fails: test_mumbai::test_ocean_tx
- [#942](https://github.com/oceanprotocol/ocean.py/issues/942): CI test randomly fails: test_utils.py::test_gas_scaling_factor
- [#939](https://github.com/oceanprotocol/ocean.py/issues/939): CI tests randomly fail: many tests saying "Exception: Address not found for ganache (alias development). Please check your address file."

calina-c:
Closes #938.

TODO:
- [x] remove fallbacks on env vars, as discussed with Trent
- [x] run a validation/check on the config dict provided

GAS_LIMIT only used from env, should we keep it in the config dict?


calina-c:
@trentmc @mariacarmina Here's the gist of what I did here:
- removed the Config class and object-based configuring, including config.ini
- removed fallbacks to env variables when configuring Ocean. The reason for this is simple: users will be able to instantiate multiple Ocean objects i.e. ganache_ocean, mumbai_ocean, etc. and compare between them
- Ocean class validates the config given to it for presence and types of all configuration options, no fallback, no guessing. Every Ocean class needs to be configured with all the options.
- updated Example Config: I kept most of it in place but flattened the result into a simple dict. I think this logic it is valuable and we should not remove it. If we do, however, decide to remove it, we should do that in a separate PR. ExampleConfig now handles the "guessing and defaulting" part for users, if they wish to use it, while the Ocean class does not care about that. It requires everything fully and properly configured. I believe this is a good balance between UX and thoroughness.
- fixed a bunch of tests, refactored all places where config was involved etc.

The readme tests were skipped when I started this issue. I believe we should have a follow-up issue to fix all the necessary readme tests and other skipped tests as needed.

Pending discussion:
I have noticed that the GAS_LIMIT option inside the config was not actually used in gas calculations. I suggest we either work it into the calculation some way, or drop it from the config dict.

trentmc (approved):

PR #974, opened 2022-09-29
Deprecate chain id from config dict function.

### LINKED ISSUE #968: Cleanup: Auto-compute CHAIN_ID from RPC_URL (then deprecate it from config dict)
Pre-requisite: #927

The nicely updated config (dicts) still need CHAIN_ID to be specified. We can simplify, and auto-compute it in instead.

How:

1. Given the RPC_URL string,
2. create a web3 object
3. From the web3 object retrieve the CHAIN_ID.

With this, we can deprecate CHAIN_ID from the config dict:)

mariacarmina:
Fixes #968  .

Changes proposed in this PR:

- removed chain_id parameter from config dict generation function
- removed checks for invalid chain_id because we have fallback to ganache network url.

### DIFF
--- a/ocean_lib/example_config.py
+++ b/ocean_lib/example_config.py
@@ -10,6 +10,7 @@ from enforce_typing import enforce_types
 
 from ocean_lib.ocean.util import get_web3
 from ocean_lib.web3_internal.constants import GAS_LIMIT_DEFAULT
+from ocean_lib.web3_internal.utils import get_chain_id_from_url
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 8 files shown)


ocean_lib/example_config.py
```diff
@@ -83,9 +83,9 @@
 
 
 @enforce_types
-def get_config_dict(chain_id: int, network_url: str) -> dict:
-    if chain_id not in CONFIG_NETWORK_HELPER:
-        raise ValueError("The chain id for the specific RPC could not be fetched!")
+def get_config_dict(network_url: str) -> dict:
```
calina-c, line 86: aren't there any usages of this function, or usages of the "CHAIN_ID" key anywhere? I remember at least in example config I left an ocurrence. Those should be removed as well.

calina-c (changes requested):

ocean_lib/example_config.py
mariacarmina: Yes, I saw the CHAIN_ID key in the config_helper dict. I will delete from there as well.

mariacarmina:
Close this due to #979.

PR #975, opened 2022-09-30
Addresses imported from the venv

### LINKED ISSUE #971: Cleanup: import ocean-contracts library for Ocean addresses
High priority because doing this issue will solve other issues:  
- #943
- #931 
- and likely more
----

The fix for [#927](https://github.com/oceanprotocol/ocean.py/issues/927), via [PR 963](https://github.com/oceanprotocol/ocean.py/pull/963), added a new function `ocean_lib/web3_internal/contract_utils.py` called `get_addresses_with_fallback(config)`. In uses the `config` dict to retrieve Ocean contract addresses.

Conversation about it:

Trent:

> I was expecting this to use the ocean-contracts [PyPi library](https://pypi.org/project/ocean-contracts/) that the [oceanprotocol/contracts repo publishes](https://github.com/oceanprotocol/contracts#python).
> 
> Is it somewhere else, and perhaps I missed it?
> 
> And: if we aren't using it yet, let's create a new github issue to switch things over so that it does use it. All in the name of shipping this PR.

Calina:

> for the address fallback, I’ll switch to contracts import 
> because right now we are importing from ocean itself, from addresses, I think
> but importing the addresses from ocean-contracts makes much more sense

Aim of this issue: switch to importing addresses from `ocean-contracts` library

mariacarmina:
Fixes #971  .

Changes proposed in this PR:

- removed `--pre` flag from ocean-lib installation step
- fixed addresses import, updated address file path
- tested on rinkeby.

### DIFF
--- a/tests/integration/remote/test_mumbai.py
+++ b/tests/integration/remote/test_mumbai.py
@@ -7,7 +7,6 @@ import random
 import string
 import warnings
 
-import pytest
 import requests
 
 from ocean_lib.ocean.ocean import Ocean
@@ -60,7 +59,6 @@ def test_nonocean_tx(tmp_path):
     assert bob_eth_after > bob_eth_before
 
 
-@pytest.mark.skip(reason="Don't skip, once fixed #943")
 def test_ocean_tx(tmp_path):
     """Do a (simple) Ocean tx on Mumbai"""
 


mariacarmina:
Test output:
Screenshot from 2022-09-30 16-54-42

It retrieved the addresses successfully.

trentmc:
This should work by default, not just if you use ExampleConfig. Putting it into ExampleConfig is too late.

Examples:
1. Anyone using ocean_lib for some remote network should be able to just specify the network RPC. From that, it should auto-find the chainid. From that, it should auto-find the addresses. If the user has to specific chainid or addresses, then we're still not done.
2.  You can capture the essence of (1) in the test_mumbai.py::test_ocean_tx. It should work, without having to specify the Ocean addresses anywhere. (And not by forcing it in some example config).

calina-c:
@mariacarmina please use the latest version of main.Then, keeping your `--pre` changes, I think adding the changes in get_contracts_addresses, like done for the CHAIN_ID, is a good way to go.

mariacarmina:
> 

Can we have some test keys on mumbai as we have exposed for ganache? In order to make `test_mumbai::test_ocean_tx` pass, we need to have remote test keys exported and I thought that maybe we can set them in `pytest.ini` file as well.

trentmc:
> @trentmc, can we have some test keys on mumbai as we have exposed for ganache? In order to make `test_mumbai::test_ocean_tx` pass, we need to have remote test keys exported and I thought that maybe we can set them in `pytest.ini` file as well.

We have test keys, but they cannot be publicly visible because bots will eat the fake MATIC, and you can only get a little fake MATIC at a time from faucets.

Our test keys are visible to CI as Github Actions Secrets. The CI can see them: the other Mumbai test needs fake MATIC and the tests pass.

You will need the keys locally too. Instructions are at :
https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/developers.md

trentmc (changes requested):
You can *not* put those private keys into the code, because bots will see them and grab them.

I've already explained to you elsewhere today how they can access these private keys: using github Actions Secrets. The other Mumbai test uses these already. 

Even doing this commit to this branch likely compromises the keys. Now we will likely have to generate new ones :(

ocean_lib/example_config.py
```diff
@@ -86,6 +90,11 @@ def get_config_dict(chain_id: int, network_url: str) -> dict:
 
     if chain_id != 8996:
         config_helper["METADATA_CACHE_URI"] = METADATA_CACHE_URI
+        config_helper["ADDRESS_FILE"] = (
```
calina-c, line 93: this shouldn't be in Example Config, it should be in `get_contracts_addresses`, similar to how we configure chain_id automatically when undefined.

PR #976, opened 2022-09-30
Removed pass-through function for contracts' addresses retrieval

### LINKED ISSUE #970: Cleanup: re pass-through function "get_contracts_addresses_web3()"
The fix for [#927](https://github.com/oceanprotocol/ocean.py/issues/927), via [PR 963](https://github.com/oceanprotocol/ocean.py/pull/963), included the following in `ocean_lib/ocean/util.py`:

```python
def get_contracts_addresses(config) -> Dict[str, str]:
    return get_contracts_addresses_web3(config)
```

Conversation about it:
> [Trent] Why the pass-through function? Can we instead just have callers call get_contracts_addresses_web3() directly?

> [Calina]: this is an older thing actually, we need the function in two modules and we always had trouble with the imports. I can look into moving this

The aim of this issue is to move it, how Calina envisioned

mariacarmina:
Fixes #970  .

Changes proposed in this PR:

- removed pass-through function
- renamed it into `get_contracts_addresses_web3`

### DIFF
--- a/ocean_lib/ocean/test/test_util.py
+++ b/ocean_lib/ocean/test/test_util.py
@@ -35,7 +35,7 @@ def test_get_web3_connection_provider(monkeypatch):
 
 @pytest.mark.unit
 def test_get_ocean_token_address(config):
-    addresses = util.get_contracts_addresses(config)
+    addresses = util.get_contracts_addresses_web3(config)
     assert addresses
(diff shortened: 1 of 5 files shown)


calina-c (approved):

calina-c:
Superseded by #980

PR #979, opened 2022-10-03
Remove chain_id from config and simplify some complexities.

### LINKED ISSUE #968: Cleanup: Auto-compute CHAIN_ID from RPC_URL (then deprecate it from config dict)

calina-c:
Fixes #968 (another option)


ocean_lib/web3_internal/contract_utils.py
```diff
@@ -92,3 +102,59 @@ def _checksum_contract_addresses(
...
+    Return a web3 instance connected via the given network_url.
+    Adds POA middleware if needed.
+    """
+    if not network_url:
+        network_url = GANACHE_URL
+
+    provider = get_web3_connection_provider(network_url)
+    web3 = Web3(provider)
+
+    # TODO: fix this...
```
mariacarmina, line 119: What should we do here exactly?

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -59,6 +63,12 @@ def get_addresses_with_fallback(config):
 @enforce_types
 def get_contracts_addresses(config) -> Optional[Dict[str, str]]:
     """Get addresses for all contract names, per network and address_file given."""
+    if "CHAIN_ID" not in config:
+        w3 = get_web3(config["RPC_URL"])
+        # cache it to prevent further calls
+        # TODO: maybe we shouldn't?
```
mariacarmina, line 69: Now I saw that you have kept this key. In my PR I have removed all the occurrences.

calina-c: yes, because Alex asked us to save RPC calls, and therefore I'm caching the CHAIN_ID key. I think it is a reasonable compromise.

PR #1006, opened 2022-10-10
Updated block confirmations map for Goerli network.

### LINKED ISSUE #1005: Block confirmations dict not updated
Block confirmations dict is not updated for Goerli network.

mariacarmina:
Fixes #1005  .

Changes proposed in this PR:

- updated block confirmations map for Goerli.

### DIFF
--- a/ocean_lib/web3_internal/constants.py
+++ b/ocean_lib/web3_internal/constants.py
@@ -26,8 +26,7 @@ MIN_INT256 = 2**255 * -1
 block_number poll interval = 1/2 average block time for a given chain"""
 BLOCK_NUMBER_POLL_INTERVAL = {
     1: 6.5,
-    3: 6.0,
-    4: 7.5,
+    5: 7.5,
     56: 1.5,
     137: 1.0,
     246: 2.6,


calina-c:
@mariacarmina new configurations for networks will be done using a brownie config file. Right now, it is minimal and left to the user to configure, but if you want to help improve it, any help is appreciated.

calina-c:
@trentmc should this still be done in main? The brownie version will automatically pick up on the `required_confs` key in the network-config file. Also, with the current config we have in main, the user can define any block confirmation number. So I assume this issue is only for the default in Example Config? Not sure what we're trying to do here.

@mariacarmina this seems to update the poll interval and not the config itself, is that the goal?

trentmc:
> Not sure what we're trying to do here.

Me neither 

We should be focusing on getting Brownie branch as main branch 

It could be that this pr becomes fully obsolete

mariacarmina:
I did the modifications before brownie was integrated and I noticed that some maps weren't updated for goerli. I thought this PR was closed since we are focusing on a different config.

calina-c:
I'll leave this still open in case of emergency. For now, let's focus on the brownie branch.

Issue #1007, opened 2022-10-11
(a) consume flow does not work with GraphQL assets on Mumbai (b) asset returns as None


MBadea17:
[Update Oct 14 trentmc]: at first we had both issues (a) and (b). It wasn't clear if they were the same. But, they were different. I could have split the issues, but we have branches & PRs with "1007" so I'm keeping it all here.

- (a) was reported in detail right below, in this description. It was solved on Oct 13 [see this comment](https://github.com/oceanprotocol/ocean.py/issues/1007#issuecomment-1277611428)
- (b) was reported in detail in [this comment](https://github.com/oceanprotocol/ocean.py/issues/1007#issuecomment-1276286245). Solution is now WIP, via [this PR](https://github.com/oceanprotocol/ocean.py/pull/1009)
- We'll close this issue when both (a) and (b) are solved.

----
**Describe the bug - issue (a)**
Consume flow does not work with an asset of GraphQL type, on Mumbai. 
Running the following command
`order_tx_id = ocean.assets.pay_for_access_service(asset, bob_wallet)`
throws this error
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mihai/my_project/venv/lib/python3.8/site-packages/enforce_typing/decorator.py", line 29, in wrapper
    return func(*args, **kwargs)
  File "/home/mihai/my_project/venv/lib/python3.8/site-packages/ocean_lib/ocean/ocean_assets.py", line 808, in pay_for_access_service
    raise AssetNotConsumable(consumable_result)
ocean_lib.agreements.consumable.AssetNotConsumable: 2
```

**To Reproduce**
1. Perform Set envvars from Quickstart: Remotely Publish Data NFT & Datatoken (https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/simple-remote.md#set-envvars)
```
export REMOTE_TEST_PRIVATE_KEY1=<your REMOTE_TEST_PRIVATE_KEY1>
export REMOTE_TEST_PRIVATE_KEY2=<your REMOTE_TEST_PRIVATE_KEY2>
```
2. Perform Setup in Python from Quickstart: Remotely Publish Data NFT & Datatoken (https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/simple-remote.md#setup-in-python)

```
# Create Ocean instance
import os
from ocean_lib.example_config import ExampleConfig
from ocean_lib.ocean.ocean import Ocean
[...]

Issue #998, opened 2022-10-07
"replacement transaction underpriced" error in test_mumbai::test_nonocean_tx


trentmc:
`test_mumbai::test_nonocean_tx` randomly fails in CI, such as [this run](https://github.com/oceanprotocol/ocean.py/actions/runs/3203531109/jobs/5233736814).

I first ran into this error [during work on #996](https://github.com/oceanprotocol/ocean.py/issues/966#issuecomment-1260310606). I had put what I thought was a fix. But I guess it didn't fully fix it!

Hypothesis: random.random() isn't enough. Perhaps it picked the same random seed?

The aim here: fix it better, so that the test doesn't randomly fail.

Traceback:
```text
=================================== FAILURES ===================================
_______________________________ test_nonocean_tx _______________________________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/test_nonocean_tx0')

    def test_nonocean_tx(tmp_path):
        """Do a simple non-Ocean tx on Mumbai. Only use Ocean config"""
    
        # setup
        config = _remote_config(tmp_path)
        ocean = Ocean(config)
        (alice_wallet, bob_wallet) = _get_wallets(ocean)
    
        # Get gas price (in Gwei) from Polygon gas station
        gas_station_url = "https://gasstation-mumbai.matic.today/v2"
        gas_price = requests.get(gas_station_url).json()["fast"]["maxFee"]
    
        # Simplest possible tx: Alice send Bob some fake MATIC
        web3 = ocean.web3
        bob_eth_before = web3.eth.get_balance(bob_wallet.address)
        nonce = web3.eth.getTransactionCount(alice_wallet.address)
    
        # avoid "replacement transaction underpriced" error: make each tx diff't
        amt_send = 1e-8 * random.random()
        tx = {
            "nonce": nonce,
            "gasPrice": web3.toWei(gas_price, "gwei"),
            "gas": 21000,  # a standard ETH transfer needs 21K gas
            "chainId": web3.eth.chain_id,
            "to": bob_wallet.address,
            "from": alice_wallet.address,
            "value": web3.toWei(amt_send, "ether"),
        }
        signed_tx = web3.eth.account.sign_transaction(tx, alice_wallet.private_key)
    
        print("Do a send-Ether tx...")
        try:  # it can get away with "insufficient funds" errors, but not others
            tx_hash = web3.eth.send_raw_transaction(signed_tx.rawTransaction)
        except ValueError as error:
[...]

PR #1033, opened 2022-10-21
Ocean brownie 2b

calina-c:
Remove web3 parameter from contracts, ocean assets.
Move block confirmations and transaction timeouts from custom wallets into the config dict.

### DIFF
--- a/tests/integration/remote/test_mumbai.py
+++ b/tests/integration/remote/test_mumbai.py
@@ -26,6 +26,7 @@ def test_nonocean_tx(tmp_path):
     gas_price = requests.get(gas_station_url).json()["fast"]["maxFee"]
 
     # Simplest possible tx: Alice send Bob some fake MATIC
+    # TODO: replace with send_ether?
     web3 = ocean.web3
     bob_eth_before = web3.eth.get_balance(bob_wallet.address)
     nonce = web3.eth.getTransactionCount(alice_wallet.address)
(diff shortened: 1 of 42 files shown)


tests/integration/remote/test_mumbai.py
```diff
@@ -21,41 +18,19 @@ def test_nonocean_tx(tmp_path):
...
-    gas_price = requests.get(gas_station_url).json()["fast"]["maxFee"]
-
     # Simplest possible tx: Alice send Bob some fake MATIC
     web3 = ocean.web3
     bob_eth_before = web3.eth.get_balance(bob_wallet.address)
-    nonce = web3.eth.getTransactionCount(alice_wallet.address)
-
-    # avoid "replacement transaction underpriced" error: make each tx diff't
-    normalized_unixtime = time.time() / 1e9
-    amt_send = 1e-8 * (random.random() + normalized_unixtime)
```
trentmc, line 35: We need to make it such that each time the test is run, it's a unique tx. Otherwise the EVM will think you're trying to send the same tx, and you will have issues with nonces, and we will see issues #998 and #1007(b) again.

The easiest way to guarantee a unique tx is to send a unique amount each time. At first I just used random.random(), but that was not enough because it may have the same random seed. So I also added the unixtime. That's how I solved those issues above.

How to fix it here: First, re-insert:
```python
    normalized_unixtime = time.time() / 1e9
    amt_send = 1e-8 * (random.random() + normalized_unixtime) #in ether
```

Then use `amt_send` in `send_ether()`. E.g.:
```python
  send_ether(config, alice_wallet, bob_wallet.address, f"{amt_send:.15f} ether")
```

trentmc (changes requested):
All's well! Except one small changed needed.

trentmc (approved):

Issue #1034, opened 2022-10-21
Installing ocean.py in Mac with M1 processor fail due to issues with Cryptography


jfdelgad:
**Describe the bug**

**To Reproduce**
1. Installing the library on Mac with M1 processors fail due to an issue with Cryptography library

**Expected behavior**
The library installation should succeed in M1 based computers.

**Desktop (please complete the following information):**
 - OS: MacOS
 - python 3.8+

Installing Cryptography succeeds in Mac with M1 but the version used in ocean.py since to have problems.

trentmc:
Related previous issue: https://github.com/oceanprotocol/ocean.py/issues/887

trentmc:
Note: full credit to this fix goes to @jfdelgad . I just did it to keep its change clean from other work, e.g. in https://github.com/oceanprotocol/ocean.py/pull/1046/files

PR #1046, opened 2022-10-27
Fixes #1041

### LINKED ISSUE #1041: Creating onchain datasets fail if the view function called has i put parameters
**Describe the bug**
Creating function call datasets fail if the function has input parameters. This is because the connectivity_check under `pay_for_access_service` fails with error AssetNonConsumable with code 2. This error can be traced to the call to the provider that responds with "missing parameters".  This error does not happens if the function call is made to a function without parameters. Please note that the error is not at the moment of consuming but when an attempt to pay for access is done

**To Reproduce**

Using: GORLI
Create two wallets `alice_wallet` and `bob_wallet`

```python
 name = "Uniswap v2 function call"
contract_address = "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D"
contract_abi = {
 		"inputs": [
 			{
 				"internalType": "uint256",
 				"name": "amountIn",
 				"type": "uint256"
 			},
 			{
 				"internalType": "address[]",
 				"name": "path",
 				"type": "address[]"
 			}
 		],
 		"name": "getAmountsOut",
 		"outputs": [
 			{
 				"internalType": "uint256[]",
 				"name": "amounts",
 				"type": "uint256[]"
 			}
 		],
 		"stateMutability": "view",
 		"type": "function"
 	}

# #create asset
(data_nft, datatoken, asset) = ocean.assets.create_onchain_asset(name, contract_address, contract_abi, alice_wallet)
print(f"Just published asset, with did={asset.did}")


to_address = bob_wallet.address
amt_tokens = ocean.to_wei(10)  # just need 1, send more for spare
datatoken.mint(to_address, amt_tokens, alice_wallet)

order_tx_id = ocean.assets.pay_for_access_service(
    asset,
    bob_wallet,
    service,
    consume_market_order_fee_address=bob_wallet.address,
    consume_market_order_fee_token=datatoken.address,
    consume_market_order_fee_amount=0,
)
```

This will produce the following error:
```
raise AssetNotConsumable(consumable_result)
ocean_lib.agreements.consumable.AssetNotConsumable: 2

```
**Expected behavior**
Access to the asset should be granted, the call should not fail

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 1.6.1

**Additional context**

jfdelgad:
Fixes #1041

- handles missing file_name in the header response form provider


setup.py
```diff
@@ -48,7 +48,7 @@
     "docker",
     "mccabe",
     "pylint",
-    "pytest",
+    "pytest=7.1.3",
```
trentmc, line 51: Why?

trentmc (changes requested):

setup.py
jfdelgad: Had issues running the jobs and this was suggested by Alex who was having the same issue.

alexcos20: pytest 7.2.0 (latest) is broken.
for now, only solution is to downgrade to 7.1.3

setup.py
```diff
@@ -29,7 +29,7 @@
     "tqdm",
     "pytz",
     "web3==5.28.0",
-    "cryptography==3.3.2",
+    "cryptography",
```
alexcos20, line 32: there were some discussions about cryptography , especially on MacOS
maybe it's better if we force it to a specific version

setup.py
alexcos20: @trentmc

setup.py
trentmc: The issues before were because it had been forced to a specific (older) version. If not forcing solves the issue, keep like that:)

setup.py
trentmc: I see. I thought it was solvable by renaming "py.test" file. OK then, no problem.

setup.py
trentmc: More detail; The issue in question is : https://github.com/oceanprotocol/ocean.py/issues/1034

And @jfdelgad had found that removing the constraint.

TBH this really should be its own PR, to make it easy for people to track.

Give me a minute, I'll do that...

trentmc: OK, done. @jfdelgad please merge main into your branch, and then this change will go away:)

alexcos20: I'm not a big fan of wild dependency  (ie: with no fixed version). That can lead to different scenarios and bugs, very hard to reproduce.

Therefore I recommend "  cryptography==38.0.1"  (https://github.com/oceanprotocol/ocean.py/actions/runs/3339736289/jobs/5529041238#step:6:30)

PR #1065, opened 2022-11-08
Increase waiting time for address file to 5 minutes.

### LINKED ISSUE #939: CI tests randomly fail: many tests saying "Exception: Address not found for ganache.  Please check your address file."
Pre-requisite to this issue: [#938](https://github.com/oceanprotocol/ocean.py/issues/938) "Streamline Config: no config.ini file, no dict, just envvars." Then, doing this issue will be much, much simpler.

Note: brownie doesn't fix this.

----
**Describe the bug**

CI randomly fails across all tests that need ganache. 

[Here's an example fail](https://github.com/oceanprotocol/ocean.py/actions/runs/3085128851/jobs/4988063967#step:7:8279) from [this PR](https://github.com/oceanprotocol/ocean.py/pull/913) (for remote C2D). Summary from the example: [all_fails.txt](https://github.com/oceanprotocol/ocean.py/files/9601974/all_fails.txt), [example_error.txt](https://github.com/oceanprotocol/ocean.py/files/9601975/example_error.txt).

[Here's another fail](https://github.com/oceanprotocol/ocean.py/actions/runs/3085347784), this time from a direct commit to main - tweaking a README. 

[Here's another, from Oct 27](https://github.com/oceanprotocol/ocean.py/actions/runs/3329714624/jobs/5507277845).

Importantly, they don't always fail, it seems to be randomly. [Here's an example pass](https://github.com/oceanprotocol/ocean.py/actions/runs/3085166191) from a direct commit to main, from the commit from another README tweak done around same time as the tweak above.

**The core problem**

Ganache hasn't had enough time to get going yet. If we give it more time, it would solve it.

There are currently scripts that wait for barge to be ready. But they might be broken, since they don't actually wait. Likely:
- they wait for a certain file to be created, but that's probably not a strong enough gate, they likely have to wait for more. 
- OR we're looking at the wrong file.

Note: while Brownie is AWESOME... it doesn't solve this:(

**To Reproduce**

Any push will trigger CI (from main, or a branch/PR). Things may fail, or not.

**Notes**

This is recent. Just started happening after the commit for #923, to start having remote tests. That was introduced in this commit: https://github.com/oceanprotocol/ocean.py/commit/a4fb91153891e2b68be90d4a5a7558bbb44e14d4

But we've been seeing other random failures of CI, e.g. #905. Though, that one may be more due to random timeout of C2D, as Calina mentions in the issue.

This issue is perhaps another symptom of a super-messy Config setup. #938 proposes streamlining Config. 

However maybe in the near term there's a quick fix for _this_ issue.

That said, it's almost certainly better to simply do #938 ASAP. It will remove a lot of pain for us in developing on ocean.py.

calina-c:
Closes #939.

I can not entirely verify this, but here is what I found: with 50 loops (as previously), the time was 250s (a little over 4 minutes). For the test failure mentioned in the original ticket, the barge run time was 4min:15sec and I remember empirically that the entire test suite was under 8mins for cases where the address file was missing. Based on these calculations, I increased the waiting time to 5 minutes (60 loops of 5 seconds), which I think is a reasonable increase.

### DIFF
--- a/.github/workflows/libcheck.yml
+++ b/.github/workflows/libcheck.yml
@@ -29,7 +29,7 @@ jobs:
         working-directory: ${{ github.workspace }}/barge
         run: |
           bash -x start_ocean.sh --no-dashboard 2>&1 --with-provider2 --with-rbac --with-c2d > start_ocean.log &
-          for i in $(seq 1 50); do
+          for i in $(seq 1 60); do
             sleep 5
             [ -f "$HOME/.ocean/ocean-contracts/artifacts/ready" -a -f "$HOME/.ocean/ocean-c2d/ready" ] && break
             done
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -28,7 +28,7 @@ jobs:
         working-directory: ${{ github.workspace }}/barge
         run: |
           bash -x start_ocean.sh --no-dashboard 2>&1 --with-provider2 --with-rbac --with-c2d --with-thegraph > start_ocean.log &
-          for i in $(seq 1 50); do
+          for i in $(seq 1 60); do
             sleep 5
             [ -f "$HOME/.ocean/ocean-contracts/artifacts/ready" -a -f "$HOME/.ocean/ocean-c2d/ready" ] && break
             done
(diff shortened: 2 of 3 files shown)


trentmc (approved):
Reasonable + practical fix. Worksforme

PR #1084, opened 2022-11-22
Document fix for failed txs in README

### LINKED ISSUE #1026: Document fix for txs that sometimes fail with "transaction underpriced"
**Describe the bug**

In using the eth-predict readme (on Polygon), I sometimes encounter an error "transaction underpriced". What's happening: network rejects the tx because it doesn't think you're paying a high enough gas fee. 

Workaround in predict-eth: if I retry 1-2 times, it usually goes through.

But we need to fix this properly because:
- the workaround is not suitable for fully automated flows
- the workaround is poor UX for manual flows like READMEs
- and what if it doesn't go through, and we need to pay a higher gas fee? It's not so easy within ocean.py right now

**Overall goal**

Get txs through more reliably, more often.

**Proposed solution**

- If the error comes back, try again 1-2 times. 
- If there are still issues, then multiply gas price by e.g. 1.2 and try again. 
- Repeat (multiply by 1.2, try again) until it works, or until e.g. 50 tries done.

mariacarmina:
Fixes #1026 .

Changes proposed in this PR:

- updated the README with the fix for `transaction underpriced` issue.

### DIFF
--- a/READMEs/data-nfts-and-datatokens-flow.md
+++ b/READMEs/data-nfts-and-datatokens-flow.md
@@ -39,6 +39,15 @@ pip3 install --pre ocean-lib
 
 - Mac users: if you encounter an "Unsupported Architecture" issue, then install including ARCHFLAGS: `ARCHFLAGS="-arch x86_64" pip install ocean-lib`. [[Details](https://github.com/oceanprotocol/ocean.py/issues/486).]
 
+- in order to avoid errors from `brownie`, such as `transaction underpriced`,
+we recommend setting the priority fee before interacting with the smart contracts
+in the python console like this:
+    ```python
+    import brownie.network
+    from brownie.network import priority_fee
+    
+    priority_fee(brownie.network.chain.priority_fee)
+    ```
 #### Create your network configuration file
 
 ocean.py uses brownie to connect to deployed smart contracts. To configure the RPC URLs, gas prices and other settings,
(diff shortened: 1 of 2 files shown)


READMEs/data-nfts-and-datatokens-flow.md
```diff
@@ -39,6 +39,15 @@ pip3 install --pre ocean-lib
...
 
+- in order to avoid errors from `brownie`, such as `transaction underpriced`,
+we recommend setting the priority fee before interacting with the smart contracts
+in the python console like this:
+    ```python
+    import brownie.network
+    from brownie.network import priority_fee
+    
+    priority_fee(brownie.network.chain.priority_fee)
+    ```
```
trentmc, line 50: This is a start. However it has issues:
- It's not a problem in ganache, only the remote flow. So whatever you put in, it should be in simple-remote.md README, not the simple ganache one.
- doesn't the current fix simply just set the value to what it was already set to? We would need to set it to something different. And, I'm looking at the brownie docs, there are _many_ potential different cases to cover.
- it's too verbose
- it's too early in the README - we haven't even asked people to open the console yet. 

But I have a simple fix: just point to the brownie docs, in the right place!

What to do, specifically: 
- remove the current change
- in simple-remote.md, in section "Setup in Python", at the very end, put in something like the following.

"""
If you get a gas-related error like "transaction underpriced", you'll need to change the `priority_fee` or `max_fee`. See details in [brownie docs](https://eth-brownie.readthedocs.io/en/stable/core-gas.html).
"""

That's it! :)

trentmc (changes requested):

READMEs/data-nfts-and-datatokens-flow.md
mariacarmina: That fix helped for our polygon test, because apparently I have set before with different values for the priority fee and it failed randomly. After I have used the recommended fee for polygon which will change automatically. 
I will change it as you suggested above.

Issue #1087, opened 2022-11-23
ocean_assets::create returns a did, when it was supposed to return a ddo or a tuple


trentmc:
### Background

The fix for #1007 was supposed to be: the user can choose whether to wait for aquarius via the `wait_for_aqua` boolean argument. All higher-level methods pass through to the `create()` method, which also accepts `wait_for_aqua`. Here's the commit: https://github.com/oceanprotocol/ocean.py/commit/e35dd36f33c7772e7a69c94c9b43088435034130. That commit was about a month ago.

Working on that, I had realized that sometimes the caller just wanted the `asset` (ddo), and other times it was super-useful to have the whole tuple `(data_nft, datatoken, asset)`. This dovetails with wait_for_aqua: if wait_for_aqua=False then often the user only really cares about the `data_nft` and `datatoken`, and not the `asset`. Upon that realization, I added a complementary argument to `create()` called `return_asset` which leads to either just `asset` being returned, vs the tuple.

### The actual bug

I discovered this via code inspection. `ocean_assets::create()` has the following code:

```python
        if wait_for_aqua:
            asset = self._aquarius.wait_for_asset(did)
        else:
            return did

        # Return
        if return_asset:
            return asset
        else:
            datatokens = [
                Datatoken(self._config_dict, d["address"]) for d in datatokens
            ]
            return (data_nft, datatokens, asset)
```

Breakdown:
- The first two lines were added by commit e35dd above.
- The next two lines were added by a new commit 10 days ago https://github.com/oceanprotocol/ocean.py/commit/286e2c1a58effa4653a063316ff264993e7d11c6 . It was a large commit, lots going on
- The remaining lines were also part of e35dd above. They're the logic to decide what to return - asset or tuple. They also have to condition the datatokens into a nice list.

The two lines added by commit 286e2c1 were unnecessary, and have two errors. 
- Unnecessary: they don't need to handle returning things, since the lines right below handle that. 
- Error 1: it returned the did. create() is supposed to return either the ddo ("asset") or the tuple.
- Error 2: it didn't account for whether to return asset vs the tuple

### How to fix

Simple! Delete the two extra lines

PR #1088, opened 2022-11-23
Fix #1086 #1087 - create() issues

### LINKED ISSUE #1086: create_url_asset() ignores wait_for_aqua argument
### Describe the bug

The fix for #1007 was supposed to be: the user can choose whether to wait for aquarius via the `wait_for_aqua` boolean argument. The new-ish methods `ocean_assets::create_url_asset()`, `create_graphql_asset()`, and `create_onchain_assets()` all have that argument. They pass through to `create1()` which also accepts `wait_for_aqua`. Here's the commit: https://github.com/oceanprotocol/ocean.py/commit/e35dd36f33c7772e7a69c94c9b43088435034130

The problem: looking at the code that calls `create1`, the calling methods neglect to pass in `wait_for_aqua`. So it reverts to the default of `True`. 

I also encountered this in practice in work on READMEs.

### Describe the fix

Make each call to `create1` to include `wait_for_aqua` argument. Example:
```python
self._create1(name, files, publisher_wallet, wait_for_aqua)
```

trentmc:
Fixes #1086 and #1087

### DIFF
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -266,7 +266,7 @@ class OceanAssets:
     ) -> tuple:
         """Create an asset of type "UrlFile", with good defaults"""
         files = [UrlFile(url)]
-        return self._create1(name, files, publisher_wallet)
+        return self._create1(name, files, publisher_wallet, wait_for_aqua)
 
     @enforce_types
     def create_graphql_asset(
@@ -279,7 +279,7 @@ class OceanAssets:
     ) -> tuple:
         """Create an asset of type "GraphqlQuery", with good defaults"""
         files = [GraphqlQuery(url, query)]
-        return self._create1(name, files, publisher_wallet)
+        return self._create1(name, files, publisher_wallet, wait_for_aqua)
 
     @enforce_types
     def create_onchain_asset(
@@ -294,7 +294,7 @@ class OceanAssets:
         chain_id = self._chain_id
         onchain_data = SmartContractCall(contract_address, chain_id, contract_abi)
         files = [onchain_data]
-        return self._create1(name, files, publisher_wallet)
+        return self._create1(name, files, publisher_wallet, wait_for_aqua)
 
     @enforce_types
     def _create1(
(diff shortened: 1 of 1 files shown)


trentmc:
@calina-c please review. FYI the only failing tests are from 'replacement transaction underpriced' in test_mumbai. That's outside the scope of this issue; it's for #1089.

ocean_lib/ocean/ocean_assets.py
```diff
@@ -598,8 +598,6 @@ def create(
         # Fetch the asset on chain
         if wait_for_aqua:
             asset = self._aquarius.wait_for_asset(did)
-        else:
```
calina-c, line 601: I would keep the did return here, because otherwise we "loose" it. When not waiting for aqua, if a did is generated, then the user won't know what they created. They can use the did later to wait for aqua or handle in any way, but if we delete this they have no reference to what was created.

calina-c (changes requested):
I left a small comment. From experience, I remember it is difficult to debug without that did.

ocean_lib/ocean/ocean_assets.py
trentmc: This change is wrt issue #1087.  The bugs that it is solving are:

- Error 1: it returned the did. create() is supposed to return either the ddo ("asset") or the tuple
- Error 2: it didn't account for whether to return asset vs the tuple

That's why right below it does the "return"

However, I agree, it would be useful to also get the did as part of the tuple. I would consider that a separate issue, because it's changing the interface of create(). I can create that issue and put it as part of this PR. OK?

trentmc: > However, I agree, it would be useful to also get the did as part of the tuple. I would consider that a separate issue, because it's changing the interface of create(). I can create that issue and put it as part of this PR. OK?

I dug into the code more. `create()` already tacks the did onto the ddo. Though it's unclear whether it retains that after the call to aquarius.

Note: I just created #1094 "Improve ocean_assets::create() interface & implementation". It will make create() much better. I plan to do it soon.

Rather than changing did stuff in this PR, let's defer it to that new issue, as part of a bigger cleanup of create().

Then, are you ok with this PR as-is?

calina-c (approved):

PR #1099, opened 2022-11-23
Towards #1094: Fix confusion between "asset" and "ddo"

### LINKED ISSUE #1094: Improve ocean_assets::create() interface & implementation
### Issues

- Confusion between "asset" and "ddo". Most of the code that uses an "asset" object is really just using a "ddo". We should call it as such, and let "asset" be a more general idea of the _union_ of (data nft, datatokens, ddo).
- Until recently, ocean_assets.create() didn't have good defaults for many of its args. It was super-hard to use. So, I built _create1(), which filled in good defaults, and on top of that, three even simpler methods: `create_url_asset()`, `create_graphql_asset()`, and `create_onchain_asset()`. Those have an even simpler interface and allowed us to nicely clean up the READMEs. Now, create() _does_ have more good defaults. And it is still often used directly. 
- However in create(), the developer still has to do a lot of work to set the needed values (metadata, services, etc).
- And now we have four methods for asset creation: the lower-level create() plus the three new higher-level ones.
- create()'s implementation is quite hard to read, partly because it's so long
- create()'s unit tests are kinda hard to navigate, and (it looks like) not thorough
- if someone wants to create an asset, it's usually just one datatoken, and they already know the pricing: free, or fixed-price with a price value. But right now, the work to set the price must be done after, and it's fairly involved too
- there are many unit tests that pass in 10+ args to create(). Most are boilerplate. This makes it hard to maintain tests.

### TODOs

- [x] clearer labeling: code uses "ddo" to mean ddo, and "asset" to mean union of (data nft, datatokens, ddo)
- [x] create() becomes as simple to use as the create_url_asset() etc, via even better use of good defaults, easy-to-read implementation, and easy-to-read unit tests
- [x] include: create() always returns a tuple (data nft, datatokens, ddo). Don't ever return just ddo
- [x] include: in tests for create(), usually just use good defaults, vs 10+ args
- [ ] create() lets the user optionally set a pricing scheme (free or fixed-price) with associated parameters
- [ ] deprecate create_url_asset(), because create() has becomes so good

### TODOs (added by Călina)
- [ ] throughout the tests, there are a lot of assets created that use the same defaults. Comb through all the usages and use a single test helper to publish. Remove duplicated usages.
- [ ] same goes for metadata in tests -> use a fixture instead; also check `url =` and other common links

trentmc:
Solving the issue in the first bullet of #1094: 
- Confusion between "asset" and "ddo". Most of the code that uses an "asset" object is really just using a "ddo". We should call it as such, and let "asset" be a more general idea of the union of (data nft, datatokens, ddo).

What I did: 
- Rename "Asset (class)" --> "DDO (class)", "asset (object)" --> "ddo (object)", and related. While a lot of code is changed, it's all basically renaming. All tests pass, except for test_ve_ocean due to the gas error (being worked on elsewhere)

UPDATE 20221128: 
- this PR has all the asset -> ddo changes. There are many! Alas, there was a bug on "chain not updating state". I tried various tactics but none would solve it. So my final tactic is: introduce a subset of changes at once, ensure all tests pass, merge, then repeat. [Here's the first subset-PR](https://github.com/oceanprotocol/ocean.py/pull/1116)
- This PR will remain only for reference. But we will never merge it.


trentmc:
Status:

I'm seeing test fails in test_exchange_swap_fees.py like:
```text
FAILED tests/flows/test_exchange_swap_fees.py::test_exchange_swap_fees[Ocean-0-0-1-1]
FAILED tests/flows/test_exchange_swap_fees.py::test_exchange_swap_fees[MockUSDC-0-0-1-1]
```

With errors like:
```python
        # Get balances after swap
        consumer_bt_balance_after = bt.balanceOf(consumer_wallet.address)
        consumer_dt_balance_after = dt.balanceOf(consumer_wallet.address)
        exchange_bt_balance_after = exchange_info[FixedRateExchangeDetails.BT_BALANCE]
        exchange_dt_balance_after = exchange_info[FixedRateExchangeDetails.DT_BALANCE]
    
        # Get Swapped event
        swapped_event = receipt.events["Swapped"]
    
        # Assign "in" token and "out" token
        if swapped_event["tokenOutAddress"] == dt.address:
            in_token_amount = swapped_event["baseTokenSwappedAmount"]
            out_token_amount = swapped_event["datatokenSwappedAmount"]
            consumer_in_token_balance_before = consumer_bt_balance_before
            consumer_out_token_balance_before = consumer_dt_balance_before
            consumer_in_token_balance_after = consumer_bt_balance_after
            consumer_out_token_balance_after = consumer_dt_balance_after
        else:
            in_token_amount = swapped_event["datatokenSwappedAmount"]
            out_token_amount = swapped_event["baseTokenSwappedAmount"]
            consumer_in_token_balance_before = consumer_dt_balance_before
            consumer_out_token_balance_before = consumer_bt_balance_before
            consumer_in_token_balance_after = consumer_dt_balance_after
            consumer_out_token_balance_after = consumer_bt_balance_after

        # Check consumer balances
>       assert (
            consumer_in_token_balance_before - in_token_amount
            == consumer_in_token_balance_after
        )
E       assert (3110798000000000000000 - 1091000000000000000) == 3110798000000000000000
```

However, it's sporadic:
- the tests never fail if I run just one parameterized tests, e.g. `tests/flows/test_exchange_swap_fees.py::test_exchange_swap_fees[Ocean-0-0-1-1]` never fails
- the tests only fail if I run all of `tests/flows/test_exchange_swap_fees.py::test_exchange_swap_fees`
- and they only fail _sometimes_ (!)

The shows that the updated values have somehow not registered in the smart contracts. When you query them, you get the same value as before the change. 

This is just like what I saw with two other failing tests. 
- I'd used a workaround trick to stop the fails: loop until the change has actually been seen. 
- The workarounds are in [data_nft.py](https://github.com/oceanprotocol/ocean.py/pull/1099/files#diff-dedd91f8f6004744ecaaae99465668936078aa0533317de6170cd92b7358c912) and [test_nft_creation.py](https://github.com/oceanprotocol/ocean.py/pull/1099/files#diff-805796545e6ed7776f7726d3f220ab7d3e6aaae581fdbdbdf5aac1230127475f); search for `"time.sleep"`
- The workarounds topped the fail from occurring

I could do a similar loop-based workaround for test_exchange_swap_fees.py.

However these workarounds are not a great fix, because they don't get to the root of the issue. And the root of the issue is worrisome: the chain isn't supposed to work this way. Each tx was supposed to change the smart contract's state on-chain; and then when you query the smart contract, it should give you updated values. Yet it's not.

More thoughts:
- Is this specific to ganache? Or just ganache * brownie? Or might it happen on remote networks too?
- Why do I see these fails in this PR? My intent was simply a change in variable name from "asset" --> "ddo". Did I change something else? 

TODO: get to the bottom of this... (vs do a higher-level workaround for test_exchange_swap_fees.py).
- By the time I'm done, I should be able to undo the loop-based workarounds referenced above.

trentmc:
In https://github.com/oceanprotocol/ocean.py/pull/1099/commits/beac81028da4ecc8dee14cc836d0c01c8b354d38, I just removed the other two loop-based workarounds.

As a quick check, I ran:
```console
pytest tests/flows/test_nft_creation.py::test_nft_count
```

And it failed, as expected:
```text
        count2 = data_nft_factory.getCurrentNFTCount()
>       assert count2 == count1 + 1
E       assert 0 == (0 + 1)
...
FAILED tests/flows/test_nft_creation.py::test_nft_count - assert 0 == (0 + 1)
```

trentmc:
In https://github.com/oceanprotocol/ocean.py/pull/1099/commits/beac81028da4ecc8dee14cc836d0c01c8b354d38, I just removed the other two loop-based workarounds.

As a quick check, I ran:
```console
pytest tests/flows/test_nft_creation.py::test_nft_count
```

And it failed, as expected:
```text
        count2 = data_nft_factory.getCurrentNFTCount()
>       assert count2 == count1 + 1
E       assert 0 == (0 + 1)
...
FAILED tests/flows/test_nft_creation.py::test_nft_count - assert 0 == (0 + 1)
```

This is a great test to iterate against since (a) it takes <1 s to run (b) it fails consistently.

trentmc:
>  what's special about the PR's diff: what changes are there besides "asset" and "ddo" that might have caused issues?
> If needed, I can brute-force this by just introducing subsets of changes at once, and see which issues emerge. But that's tedious, maybe I can do it faster.

I went through all the other options first. Alas, I still have the issue.

Therefore I am doing this task - just introducing subsets of changes at once.

_This_ PR will remain only for reference. But we will never merge it.

First of the new subset PRs is: https://github.com/oceanprotocol/ocean.py/pull/1116

trentmc:
This PR was my first crack at doing this. It had basically all the changes. But several tests didn't pass. There were too many changes to debug properly. So the overall PR felt unsafe. Therefore I changed my strategy, and instead did about five separate PRs, each one much smaller and self-contained. I used this PR as a guide. All the PRs are done now! :) So we can close this.

PR #1101, opened 2022-11-24
New purpose based readmes - from descriptive to automatic and mission driven

### LINKED ISSUE #1095: Streamline main README flows; harmonize main local <> remote versions
### Summary

Replace most existing READMEs with (a) a _single_ local README that does publish, consume, marketplace/fixed-price, free, search, and (b) a _single_ remote README that does similar.

Pre-requisites: let's do the following issues first. It will help this one, a lot
- [DONE] Streamline dispenser-flow README [ocean.py#910](https://github.com/oceanprotocol/ocean.py/issues/910) 
- Streamline marketplace-flow README [ocean.py#1149](https://github.com/oceanprotocol/ocean.py/issues/1149)

### Background / Motivation

Right now we have 1 installation README (brand new).

Then we have 6 local READMEs: 
1. create data nft & datatoken
2. publish flow
3. consume flow
4. marketplace flow
5. dispenser flow
6. search & filter data assets

And we have 3 remote READMEs
1. get test MATIC
2. create data nft & datatoken
3. C2D flow

(Plus there are "advanced flows" like graphql or user profiles. We'll ignore those here.)

Until recently, all these READMEs was kinda the only way to do it, because there were 30-50 lines of code (lotsa boilerplate) to do a simple publish or consume. However, work of the last 8 weeks has cleaned that up, so now a publish or consume is only a few lines max. That helped, a lot:) And it leads to opportunity; read on...

Key remaining problems:
1. "Create data nft & datatoken" isn't really that useful on its own, because it doesn't directly lead to a consumable data asset. "Publish asset" is really where the rubber hits the road.
2. In each README, users have to jump around among several READMEs to go through it. (And if we didn't do it, we'd be violating DRY like crazy)
3. remote READMEs are not thorough, and therefore don't handle some gotchas. E.g. in marketplace flow, Bob needs to buy an asset with OCEAN. In local flow, he's given free OCEAN in barge. But for remote, where does he get that OCEAN? 

### Proposed solution

Pre-requisite: clean up ocean_assets::create() more [#1094](https://github.com/oceanprotocol/ocean.py/issues/1094).

- [ ] Push "create data nft & datatoken" flow later down, into where it _is_ useful: key/value & user-profile READMEs.
- [ ] Have a single local README that combines publish flow, consume flow, marketplace flow, dispenser flow, search & replace 
- [ ] Have a single remote README that combines get test MATIC, publish flow, consume flow, marketplace flow, dispenser flow, search & replace.
- [ ] Move "further flexibility" appendix content that's in the current READMEs to >=1 READMEs that are linked lower down in the main README.

Note: C2D flow stays as its own README. (It's getting lots of work on its own.)

This proposal meets all three problems above:)

juanpablovs:
Fixes #1095  .

Changes proposed in this PR:

- This PR is not for merge. It is an example of my approach for both C2D and what I call a "mission-based readme"
- In particular in: "Creator C2D: Put an Asset and its Algorithm for sale in one step" I created a readme that simplifies the process. It has a clear goal and the publisher only needs to know what they are supposed to know: their own data and algorithm. From us there are only 2 lines of code, one import and one function call
- I added a layer on top of ocean.py basic functions that is in a folder called "processes" the purpose of these processes is to manage and automate the flows of the "missions" that the users want to accomplish with us
- Again: this is just for you to review and let me know how you like this and if its worth pursuing as an approach for redoing all the readmes
- Btw, lots of opportunities for improvements with enums and structs for automation that will ensure us and the users that the data is correctly mapped
- Thanks


trentmc:
It's kinda cool that you took a swing at this.

Just throwing everything into one big command that the user calls does not help the user gain an understanding.

ocean.py itself strives to offer interfaces where in a single line of code, it does a lot (e.g. publish; consume; ..). This is achieved when we fill in good defaults instead of needing lots of boilerplate. Users should be able to understand how to combine these single lines of code to get useful complex behavior. 

The aim of #1095 is to show an end-to-end flow of using lots of these. Therefore this PR does not solve the problem that #1095 aims to solve. As we discussed on Monday, and previously: I have already committed to solving #1095, and first I need to do #1094. Therefore please hold off from further work on #1095.

As discussed on Monday, your main target task is #40 (along with @jfdelgad). 
- First step of that is "Change existing ocean.py C2D README, to show how to submit an algorithm for use by C2D easily. (The current C2D README doesn't do this right now, rather.) IIRC, docker containers are involved." Please stay focused on that. 
- Once you're done the first step, the next step of that issue is "Improve ocean stack (ocean.py & otherwise) such that its C2D README is much simpler and lower friction to use." Here's where you'll be chopping down 30 lines of boilerplate into (ideally) 1-3 lines of code.

juanpablovs:
noted. thank you for letting me know

trentmc:
I should mention: I do like the idea of purpose based readmes! And we should consider them. 

To get there:
- let's walk before we run: the tasks I mentioned in my previous comment, for me and you and Jaime 
- then, let's consider it. We'd bypass the negative in your PR that had the giant copy-paste. Rather, it would look a bit more like our existing readmes, except focus on more on outcome rather than steps.

juanpablovs:
understood. Can I delete this? As mentioned in the initial description, I just made it show you the approach

trentmc:
> understood. Can I delete this? As mentioned in the initial description, I just made it show you the approach

Sure, go ahead.

Issue #923, opened 2022-09-17
ocean.py has no remote tests. To fix: add a simple one for Mumbai


trentmc:
### Motivation

Practical usage of ocean.py means working with remote networks. (And production.)

So we need READMEs to show how to do this. In #909 I was working on this for a simple README, and a C2D one. On Mumbai.

I hit bugs #920, #921. Trying to fix these was slow going because I didn't have any automated tests. 

So I then I dug into ocean.py tests. First, I found that there were none for remote networks.  And, I found that the root conftest.py forces the use of ganache, making it harder to write any such tests.

### TODO

- [ ] Write *a* test on a remote network (eg Mumbai). Ensure all other tests pass. This is harder than it looks because we have to fix the root conftest.py issue and ensure all the other tests run too.

PR #1107, opened 2022-11-25
Fixes #1106

### LINKED ISSUE #1106: Change existing ocean.py C2D README, to show how to submit an algorithm for use by C2D easily. (The current C2D README doesn't do this right now, rather.) IIRC, docker containers are involved.
**Describe the bug**
Simplify C2D flow Readme, motivated by [predict-eth](https://github.com/oceanprotocol/predict-eth/issues/54)

jfdelgad:
Fixes #1106 
Changes proposed in this PR:

- Reduce to one line creating nft, datatoken and asset for dataset
- Reduce to one line creating nft, datatoken and asset for algorithm
-

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -54,32 +54,14 @@ From [simple-remote](simple-remote.md), do:
 In the same python console:
 
 ```python
-# Publish data NFT & datatoken for dataset
-DATASET_data_nft = ocean.create_data_nft('DATASET_DN1', 'DATASET_DN1', alice_wallet)
-print(f"Created data NFT. Its address is {DATASET_data_nft.address}")
+# Publish data NFT, datatoken, and aset for dataset based on url
+DATASET_url = "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff"
 
-DATASET_datatoken = DATASET_data_nft.create_datatoken("DATASET_DT1", "DATASET_DT1", from_wallet=alice_wallet)
+name = "Branin dataset"
+(DATASET_data_nft, DATASET_datatoken, DATASET_asset) = ocean.assets.create_url_asset(name, DATASET_url, alice_wallet, wait_for_aqua=True)
+print(f"DATASET_data_nft address = '{DATASET_data_nft.address}'")
 print(f"DATASET_datatoken address = '{DATASET_datatoken.address}'")
 
-# Specify metadata and services for the Branin test dataset
-DATASET_date_created = "2022-09-19T10:55:11Z"
-DATASET_metadata = {
-    "created": DATASET_date_created,
-    "updated": DATASET_date_created,
-    "description": "Branin dataset",
-    "name": "Branin dataset",
-    "type": "dataset",
-    "author": "Trent",
-    "license": "CC0: PublicDomain",
-}
-
-# ocean.py offers multiple file object types. A simple url file is enough for here
-from ocean_lib.structures.file_objects import UrlFile
-DATASET_url_file = UrlFile(
-    url="https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/branin.arff"
-)
-
-DATASET_files = [DATASET_url_file]
 
 # Set the compute values for compute service
 DATASET_compute_values = {
@@ -90,26 +72,20 @@ DATASET_compute_values = {
 }
 
 # Create the Service
-from ocean_lib.services.service import Service
+from ocean_lib.structures.file_objects import UrlFile
 DATASET_compute_service = Service(
     service_id="2",
     service_type="compute",
     service_endpoint=ocean.config_dict["PROVIDER_URL"],
     datatoken=DATASET_datatoken.address,
-    files=DATASET_files,
+    files=[UrlFile(url=DATASET_url)],
     timeout=3600,
     compute_values=DATASET_compute_values,
 )
 
-# Publish asset with compute service on-chain
-DATASET_asset = ocean.assets.create(
-    metadata=DATASET_metadata,
-    publisher_wallet=alice_wallet,
-    files=DATASET_files,
-    services=[DATASET_compute_service],
-    data_nft_address=DATASET_data_nft.address,
-    deployed_datatokens=[DATASET_datatoken],
-)
+# Add service and update asset
+DATASET_asset.add_service(DATASET_compute_service)
+DATASET_asset = ocean.assets.update(DATASET_asset, alice_wallet)
 
 print(f"DATASET_asset did = '{DATASET_asset.did}'")
 ```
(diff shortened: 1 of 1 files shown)


READMEs/c2d-flow.md
```diff
@@ -90,26 +72,20 @@ DATASET_compute_values = {
 }
 
 # Create the Service
-from ocean_lib.services.service import Service
```
trentmc, line 93: Don't delete this line - you still need access to the `Service` class below.

READMEs/c2d-flow.md
```diff
@@ -90,26 +72,20 @@ DATASET_compute_values = {
...
+from ocean_lib.structures.file_objects import UrlFile
 DATASET_compute_service = Service(
     service_id="2",
     service_type="compute",
     service_endpoint=ocean.config_dict["PROVIDER_URL"],
     datatoken=DATASET_datatoken.address,
-    files=DATASET_files,
+    files=[UrlFile(url=DATASET_url)],
```
trentmc, line 81: Cleaner: rather than rebuilding a UrlFile object from the raw DATASET_url, you should be able to retrieve the existing one from the asset's list of files. (And less risky: what if the asset object itself relies on referencing the in-memory objects?)

READMEs/c2d-flow.md
```diff
@@ -90,26 +72,20 @@ DATASET_compute_values = {
...
-    metadata=DATASET_metadata,
-    publisher_wallet=alice_wallet,
-    files=DATASET_files,
-    services=[DATASET_compute_service],
-    data_nft_address=DATASET_data_nft.address,
-    deployed_datatokens=[DATASET_datatoken],
-)
+# Add service and update asset
+DATASET_asset.add_service(DATASET_compute_service)
+DATASET_asset = ocean.assets.update(DATASET_asset, alice_wallet)
```
trentmc, line 88: This in an excellent way to add this service. Nice! :)

READMEs/c2d-flow.md
```diff
@@ -175,9 +140,10 @@ print(f"ALGO_asset did = '{ALGO_asset.did}'")
 
 In the same Python console:
 ```python
-compute_service = DATASET_asset.services[0]
+compute_service = DATASET_asset.services[1]
```
trentmc, line 143: I'm not sure if this is correct. Does the DATASET_asset have two services now, instead of one? (And I'm pretty sure it still has one.)

READMEs/c2d-flow.md
```diff
@@ -206,7 +172,7 @@ ALGO_did = ALGO_asset.did
 DATASET_asset = ocean.assets.resolve(DATASET_did)
 ALGO_asset = ocean.assets.resolve(ALGO_did)
 
-compute_service = DATASET_asset.services[0]
+compute_service = DATASET_asset.services[1]
```
trentmc, line 175: Like above: I'm not sure if this is correct. Does the DATASET_asset have two services now, instead of one? (And I'm pretty sure it still has one.)

trentmc (changes requested):
Overall, great to see this. I'm glad you were able to take advantage of create_url_asset() to chop complexity down. As well as metadata updates.

Before the next request for review, please make sure that the end-to-end flow has been tested to work. The current one doesn't for sure, since there would have been an issue due to the missing Service (as I pointed out in a comment). And there will likely be an issue due to referencing services[1] not services[0] in two places (as I also pointed out).

trentmc:
Note: this PR's title and description were pointing to a github issue that was specified confusingly (title was different than description; and the issue itself was duplicate). So: I closed that github issue, and have the changed the title and description of this PR to point to the correct github issue.

READMEs/c2d-flow.md
jfdelgad: Yes, there are two services because `create_url_asset` adds an access service by default. On top of that we are adding the compute service.

READMEs/c2d-flow.md
jfdelgad: Yes, there are two services because create_url_asset adds an access service by default. On top of that we are adding the compute service.

READMEs/c2d-flow.md
jfdelgad: Done

READMEs/c2d-flow.md
trentmc: OK. 

Then why in the original one is it accessing service 0? Perhaps the original one has two services as well, with the order of services different?

READMEs/c2d-flow.md
jfdelgad: Done

READMEs/c2d-flow.md
```diff
@@ -54,32 +54,20 @@ From [simple-remote](simple-remote.md), do:
...
-DATASET_metadata = {
-    "created": DATASET_date_created,
-    "updated": DATASET_date_created,
-    "description": "Branin dataset",
-    "name": "Branin dataset",
-    "type": "dataset",
-    "author": "Trent",
-    "license": "CC0: PublicDomain",
-}
+# Publish data NFT, datatoken, and aset for dataset based on url
```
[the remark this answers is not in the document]
jfdelgad: Fixed

jfdelgad:
> Overall, great to see this. I'm glad you were able to take advantage of create_url_asset() to chop complexity down. As well as metadata updates.
> 
> Before the next request for review, please make sure that the end-to-end flow has been tested to work. The current one doesn't for sure, since there would have been an issue due to the missing Service (as I pointed out in a comment). And there will likely be an issue due to referencing services[1] not services[0] in two places (as I also pointed out).

The code that is already in the repo runs well, I did this several times a couple of days ago.
I did the same with the new one.

There are issue running this in mumbai because of the time that it takes to get things from Aquarius. 

The comment about the services was addressed in a previous comment, there are two services.

trentmc:
> The comment about the services was addressed in a previous comment, there are two services.

Sure, there are two services now. And there were two before, right? So why is a different service being pointed to? (`services[1]` vs `services[0]`). Perhaps the order was reversed?

trentmc:
> There are issue running this in mumbai because of the time that it takes to get things from Aquarius.

To be clear: you can run with your local repo, and with a remote network (Mumbai). And that's what you should be doing.

C2D README explicitly runs with remote components, not local, because local causes many issues, and it isn't a practical flow anyway.
- Details: up until about 6 weeks ago, we did have C2D README on local. But we had issue after issue because via barge, we end up running docker-in-docker which is error-prone and tedious to debug. Then we realized that it wasn't a practical flow anyway, and so the obvious fix was to go remote. And we did, via [ocean.py#909](https://github.com/oceanprotocol/ocean.py/issues/909). 
- From that work, we also found that we didn't have any other simpler remote READMEs, so we did that too, via [ocean.py#933](https://github.com/oceanprotocol/ocean.py/issues/933).
- From both those issues, we also realized that we didn't have any unit / CI testing for remote networks. So we did that too, via [ocean.py#923](https://github.com/oceanprotocol/ocean.py/issues/923).
- In short: there's history on this! And the punchline is that we C2D README needs to be remote. And therefore working well for remote

jfdelgad:
> > There are issue running this in mumbai because of the time that it takes to get things from Aquarius.
> 
> To be clear: you can run with your local repo, and with a remote network (Mumbai). And that's what you should be doing.
> 
> C2D README explicitly runs with remote components, not local, because local causes many issues, and it isn't a practical flow anyway.
> 
> * Details: up until about 6 weeks ago, we did have C2D README on local. But we had issue after issue because via barge, we end up running docker-in-docker which is error-prone and tedious to debug. Then we realized that it wasn't a practical flow anyway, and so the obvious fix was to go remote. And we did, via [ocean.py#909](https://github.com/oceanprotocol/ocean.py/issues/909).
> * From that work, we also found that we didn't have any other simpler remote READMEs, so we did that too, via [ocean.py#933](https://github.com/oceanprotocol/ocean.py/issues/933).
> * From both those issues, we also realized that we didn't have any unit / CI testing for remote networks. So we did that too, via [ocean.py#923](https://github.com/oceanprotocol/ocean.py/issues/923).
> * In short: there's history on this! And the punchline is that we C2D README needs to be remote. And therefore working well for remote

Yes, I am running this with my local repo and in mumbai as I stated in my last comment. It seems, for your comment, that you have missunderstood or that I was not able to explain myself correctly. 

The comment about mumbai is just to make everyone aware (I also shared this yesterday in slack) that you should await for at least 5 minutes to get things from aquiarious.

Let me know if there is something else that you think should be done. Otherwise as soon as you approve the PR I will merge it.

trentmc (approved):

PR #1121, opened 2022-11-28
Publish algos with default metadata including docker

### LINKED ISSUE #1108: In C2D README, show how to submit an algorithm for use by C2D *including docker container*
While the README shows how to submit a url to the algorithm, it says basically nothing about docker containers. Yet under the hood it uses a pre-prepared docker image for gpr.py. 

This means that if someone wants to use any other algorithm than gpr.py, they will need to prepare their own docker image, yet we don't provide any guidance on how to do so. This is a super-high point of friction and needs to be resolved. Fixing this is the aim of this issue.

Related: [predict-eth#40](https://github.com/oceanprotocol/predict-eth/issues/40) needs this issue done.

juanpablovs:
Fixes #1108 .

Changes proposed in this PR:

- There are different types of algos and different types of containers - leave a structure that can be expanded later on
- The start of a new group of functions. The first one of these is "create_url_algo". It is similar to create_url_asset but includes an argument named "container" that in this first case the parameter is set to "python". Other param values can be added later in the future
- the "create_url_algo" parameter is subsequently added to the create1 wrapper to account for additional metadata defaults (this also can be easily expanded later)

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -107,40 +107,11 @@ In the same Python console:
 ALGO_url = "https://raw.githubusercontent.com/oceanprotocol/c2d-examples/main/branin_and_gpr/gpr.py"
 
 name = "grp"
-(ALGO_data_nft, ALGO_datatoken, ALGO_ddo) = ocean.assets.create_url_asset(name, ALGO_url, alice_wallet, wait_for_aqua=True)
+ALGO_container = "python"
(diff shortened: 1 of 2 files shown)


trentmc (approved):

trentmc:
This was done via https://github.com/oceanprotocol/ocean.py/commit/28c724a0e03415e026ae468926618c71d211c1b3

PR #1135, opened 2022-11-30
Fix #1136 - Remove stress tests

### LINKED ISSUE #1110: Get libcheck & slow-test up to par, now that we're post-brownie
### Background

_From Trent, to Calina in slack:_

Hello, I have some quick Qs related to testing READMEs. (With the aim of a clean fix for [ocean.py#1109](https://github.com/oceanprotocol/ocean.py/issues/1109))

In `.github/workflows/libcheck.yml`, it has a line `py.test tests/test_readmes.py`. I can't figure out how this works.

- What is the aim of `libcheck.yml` ? I couldn't find context via google searches, or via the [original commit](https://github.com/oceanprotocol/ocean.py/commit/4863b42005966bff7af18585989c8752dde3b7d1) to create the file.
- `py.test` is an old thing, it was replaced by `pytest` [[ref](https://stackoverflow.com/questions/39495429/py-test-vs-pytest-command)], so why are we using it?
- the path to `test_readmes.py` in the file (`tests/test_readmes.py`) is not the path that's actually in the repo (`tests/readmes/test_readmes.py`).

_Response from Calina:_

- libcheck.yaml [sic] is a periodic github check configured in Actions to run on every Tuesday to check whether the released library version is compatible with the readmes
- the regular readme checks take the readmes and a local ocean.py clone to run tests
- libcheck installs from pypi and checks the readmes
- pytest could be fine
- but I’m not sure, we should give it some runs
- related to the path, it is probably obsolete, I haven’t checked the runs in a while since we were concentrating on brownie
- which left the slow tests and libcheck tests in a less refined state
- we could create a ticket to bring the slow tests and libcheck up to date with the new releases

### TODO for this issue

Resolve the issues that Calina points out.

### LINKED ISSUE #1136: Remove stress tests
Costs of the tests: take time + energy to maintain.

Benefit of the tests: marginal, since they're really only just about higher throughput of ocean.py and otherwise already covered by integration tests etc.

Therefore: remove

calina-c:
Fix #1136. Closes #1110


trentmc (approved):

calina-c:
cherry picked here: https://github.com/oceanprotocol/ocean.py/pull/1140

PR #1167, opened 2022-12-07
Adapt to brownie default config file

### LINKED ISSUE #1156: Add all supported networks configs in the network-config.sample
Add all supported networks in: https://github.com/oceanprotocol/ocean.py/blob/main/network-config.sample
We need to point to the network config from the venv if the config file is not defined. 
Otherwise, we need to create a custom network config for the user
Add this information in the READ.ME

mariacarmina:
Fixes #1156  .

Changes proposed in this PR:

- document brownie network names in `install.md`;
- changed `polygon` to `polygon-main` and `mumbai` to `polygon-test` all over the place
- updated tests for config;
- updated `contract_utils` to match the keys from `address.json` file;
- removed `network-config.sample`, added `network-config.copy` for CI usage.


trentmc (approved):

network-config.copy
```diff
@@ -0,0 +1,318 @@
+live:
+  - name: Ethereum
```
trentmc, line 2: This appears to be a copy of brownie's current version of "network-config.yaml".

Which means it won't evolve with future versions of "network-config.yaml".

There's gotta be a more lightweight way.

Remember: when brownie gets installed, it should also include  "network-config.yaml".

This PR should be a "light touch" that lightly edits the file, automatically, using e.g. ">", "|" and "sed".

trentmc:
>  have tried to search the path for brownie and network-config.yaml, also to printed the PYTHONPATH which is null, in the previous commits, but no success. 

Yet it must exist, if brownie is installed, no?

To flip it around: let's say that Alice has never used Brownie before. If she installs brownie, does network-config.yaml get auto-added as part of that, or does Alice need to create her own separately?

mariacarmina:
> > have tried to search the path for brownie and network-config.yaml, also to printed the PYTHONPATH which is null, in the previous commits, but no success.
> 
> Yet it must exist, if brownie is installed, no?
> 
> To flip it around: let's say that Alice has never used Brownie before. If she installs brownie, does network-config.yaml get auto-added as part of that, or does Alice need to create her own separately?
Alice will get locally the `network-config` auto-added in the correct path, yes. But I was surprised that on CI I couldn't find it.

mariacarmina:
@trentmc can I merge this PR after the tests for `merge main` commit will pass?

mariacarmina:
@trentmc I will present more detalied the process that I went through for discovering the right path for .brownie folder in CI runner.

> Yet it must exist, if brownie is installed, no?

`.brownie/network-config.yaml` should appear in the home directory after `eth-brownie` is installed. That happened to me and also to other machines. The problem appears on the CI.

> If she installs brownie, does network-config.yaml get auto-added as part of that, or does Alice need to create her own separately?

`network-config.yaml` gets auto-added as part of that. I didn't see the use case where the user should add it by itself if brownie was already installed (that worked in my case and also for @AnaLoznianu when testing the predict-eth.md README).

Methods that I have used inside the workflow .yml file:

- print the `PYTHONPATH` for searching the installed packages [here](https://github.com/oceanprotocol/ocean.py/pull/1167/commits/37b2421850dd5b73b65207e83f8d94d60e09049b) -> result:  `PYTHONPATH` is [not](https://github.com/oceanprotocol/ocean.py/actions/runs/3657890141/jobs/6181989925#step:6:473) set;
- venv doesn't exist, so impossible to search there for brownie package;
- run commands such as: 

1. `find / -type d -name '.brownie'` -> the output was not helpful in this case as you can see [here](https://github.com/oceanprotocol/ocean.py/actions/runs/3656859638/jobs/6179785160#step:6:473)

2. `find / -type f -name 'network-config.yaml'` -> the [output ](https://github.com/oceanprotocol/ocean.py/actions/runs/3657735573/jobs/6181661190#step:6:5696) was the same as at point 1.

Locally I didn't have any trouble finding `.brownie` folder, but CI acts strange when installing packages. If you have any suggestions for this, I am happy to modify as you request.

trentmc:
> CI acts strange when installing packages. If you have any suggestions for this, I am happy to modify as you request.

I did a [google search for: `"github actions" "network-config.yaml" "brownie"`](
https://www.google.com/search?q=%22github+actions%22+%22network-config.yaml%22+%22brownie%22&sxsrf=ALiCzsbk2npu_UQfO8qmzN7A3GHMch7F6A%3A1670870752933&ei=4HaXY_DCOO3c1sQPu--RuAs&ved=0ahUKEwjwssbN3vT7AhVtrpUCHbt3BLcQ4dUDCA8&uact=5&oq=%22github+actions%22+%22network-config.yaml%22+%22brownie%22&gs_lcp=Cgxnd3Mtd2l6LXNlcnAQAzIFCCEQoAE6CggAEEcQ1gQQsAM6BQgAEKIEOggIIRDDBBCgAToECCEQFUoECEEYAEoECEYYAFCqBFjeHGCXHmgBcAF4AIABb4gB4gKSAQMyLjKYAQCgAQHIAQjAAQE&sclient=gws-wiz-serp). The most obvious thing I could think of. I got two hits, both interesting.

### Hit One

Url: https://github.com/yearn/yearn-lens

Text of interest: 
"""
This would require changes in your `~/.brownie/network-config.yaml` file, affecting global Brownie configuration.
"""

They said my concern even better than me! This is the thing to avoid.

### Hit Two

Url: https://eth-brownie.readthedocs.io/_/downloads/en/stable/pdf/

"""
Section 4.2.2 4.2.2 Continuous Integration Template

Github-Actions-Mix is a template preconfigured for use with Github Actions continuous integration, as well as other
useful tools.
To initialize from this mix:
```console
$ brownie bake github-actions
```

See the Github-Actions-Mix repo for a detailed explanation of how to configure and use the tools within this template.
"""

Looks like we might have a winner! So I'd pursue that path.

network-config.copy
```diff
@@ -0,0 +1,318 @@
+live:
```
trentmc, line 1: You shouldn't need this new file anymore

trentmc (changes requested):
One thing left: delete that extra file that was added.

READMEs/install.md
```diff
@@ -44,12 +44,20 @@ pip3 install --pre ocean-lib
 #### Create your network configuration file
 
 ocean.py uses brownie to connect to deployed smart contracts. To configure the RPC URLs, gas prices and other settings,
-create and fill in a network-config.yml in your `~/.brownie` folder, or copy our (very basic) sample from ocean.py:
-
-```console
-mkdir ~/.brownie
-cp network-config.sample ~/.brownie/network-config.yaml
+create and fill in a network-config.yaml in your `~/.brownie` folder.
```
AnaLoznianu, line 47: This is automatically created when installing brownie. I think we should remove the 'create' statement

READMEs/install.md
```diff
@@ -44,12 +44,20 @@ pip3 install --pre ocean-lib
 #### Create your network configuration file
```
AnaLoznianu, line 44: This is not for the creation of the network configuration file, cause that comes out of the box with the brownie installation.
My take on this is that we should move this section way below as an information addition. Cause ocean.py should work out of the box with the default config. The only case for people to edit that file is to add the energy web chain or to configure the gas prices..

READMEs/install.md
```diff
@@ -44,12 +44,20 @@ pip3 install --pre ocean-lib
...
 
 ocean.py uses brownie to connect to deployed smart contracts. To configure the RPC URLs, gas prices and other settings,
```
AnaLoznianu, line 46: Comment from Veronica: 
Intimidating language for the lay user: “To configure the RPC URLs, gas prices and other settings,... ------> 
'Please check that you have configured all networks before proceeding.'

READMEs/install.md
mariacarmina: @trentmc should we rename the title and move the section as @AnaLoznianu suggested?

READMEs/simple-remote.md
```diff
@@ -34,6 +34,11 @@ export REMOTE_TEST_PRIVATE_KEY1=<your REMOTE_TEST_PRIVATE_KEY1>
 export REMOTE_TEST_PRIVATE_KEY2=<your REMOTE_TEST_PRIVATE_KEY2>
 ```
 
+To avoid using Infura keys, especially for Polygon networks, use this command in the console:
```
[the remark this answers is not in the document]
mariacarmina: For polygon-main, I don't use the `maticcivil` RPC because it changed to `https://polygon-rpc.com`. I changed into
"""
The brownie default RPCs require you to have your own infura account, and corresponding token WEB3_INFURA_PROJECT_ID.

    If you have an infura account: set the envvar WEB3_INFURA_PROJECT_ID
    If not: one way is to get an infura account. Simpler yet is you can bypass the need for it, by changing to RPCs that don't need tokens. The command below replaces infura RPCs with public RPCs:

"""

trentmc (approved):

PR #1179, opened 2022-12-12
Updated brownie

mariacarmina:
Changes proposed in this PR:

- updated brownie version


mariacarmina:
Closing this due to unwanted other commits on this branch. Created #1180 instead.

trentmc (changes requested):
When a user installs ocean.py, they likely also have other py libraries. If ocean.py forces to a specific version, it may make it impossible to install because of version conflicts.

This is why's it's super-important to _not_ force a specific version on users.

PR #1186, opened 2022-12-12
Fixes #1185 test_data_nft.py call to create_exchange() has wrongly-named argument "publish_market_fee_amount" 

### LINKED ISSUE #1185: test_data_nft.py call to create_exchange() has wrongly-named argument "publish_market_fee_amount"
### Background

Argument should be "publish_market_fee"

See here: https://github.com/oceanprotocol/ocean.py/blob/26cd96a8bb0b9bf5a9d734eff29ebdfbfd1442d7/ocean_lib/models/test/test_data_nft.py

Yet, weirdly, tests are passing.

### TODO

- [ ] Give it the proper argument label
- [ ] Bonus: understand why tests were passing in the first place

trentmc:
Fixes #1185


trentmc:
Note: I merged, despite error, because the error was not related at all to the change

PR #1195, opened 2022-12-13
Fix #1194 Confusion on the 'Download barge and run services section' from install.md

### LINKED ISSUE #1194: Confusion on the "Download barge and run services" section from install.md
From Veronica:

"Regarding this instruction:

“Ocean barge runs ganache (local blockchain), Provider (data service), and Aquarius (metadata cache).
In a new console:”

Problem: Am confused if I should open a new Terminal window with venv still active or not. Assumed not needed. But in the same directory???? Assumed yes. Must specify this."

trentmc:
Fixes #1194

Changes proposed in this PR:
- Fix the core issue, by:
  - moving barge section to be first. Why: once we have barge going in a separate console, we don't have to deal with that console again. (This is how it was pre-brownie too.) 
  - in the "install library" section, emphasize that it's a new console
- Tune some other language in the README, to make it easier to follow 

Note: this PR does _not_ currently fix Veronica's feedback about brownie installation. It focuses on the core barge issue


trentmc:
Note: the one failing test has nothing to do with this README.

READMEs/install.md
```diff
@@ -25,23 +25,40 @@ python3 -m venv venv
 source venv/bin/activate
 ```
 
-##### Install the library
+Now that we have barge running, we can mostly ignore its console while it runs.
```
AnaLoznianu, line 28: I'm confused. So in this section called "Download barge and run services" you are giving the commands to create a project directory and to activate the venv. But, we say that we have barge running. 
I don't see the download barge and run services commands

AnaLoznianu (changes requested):

READMEs/install.md
trentmc: Good catch! Looks like my commit that merged from main inadvertently caused this issue.

I just pushed a fix.

AnaLoznianu (approved):

Issue #1190, opened 2022-12-12
Clean up datatoken.py::create_exchange() workaround for BT_decimals != 18


trentmc:
### TODOs

1. [x] Triage: fix datatoken. Done: [PR](https://github.com/oceanprotocol/ocean.py/pull/1200), [commit](https://github.com/oceanprotocol/ocean.py/commit/65f1483a9bfa15dea8a78caa0eb0f1419c380e9a)
2. [ ] Properly fix the tests: (a) test_datatoken_enterprise.py::test_buy_from_fre_and_order, (b) test_exchange_fees.py

### Background / motivation

When working on [ocean_lib/models/test/test_exchange_fees.py](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/models/test/test_exchange_fees.py), I found that it would fail unless I gave it the _wrong_ number of decimals for the basetoken. 

Example: USDC has 16 decimals. The test uses USDC and 16 decimals. 
- If in `datatoken.py::create_exchange()`, if I called `self.contract.createFixedRate()` with `BT_decimals = BT.decimals()` (ie value of 16) then the test would fail. It's like it wouldn't register that the fees were gathered
- Yet if I put BT_decimals = 18, the test would pass. So inside `datatoken.py::create_exchange()`, I made the following workaround:

```python
        # test_datatoken_enterprise only passes when BT_decimals == 18,
        # even if BT.decimals() != 18. Accomodate with this workaround:
        if "DatatokenEnterprise" in str(self.__class__):
            BT_decimals = 18
        else:
            BT_decimals = BT.decimals()
```

### Goal of this issue

Obviously, the workaround above is ugly, and shouldn't be there. And maybe it causes more bugs! We need a clean solution. In doing the solution, maybe we'll find other bugs.

### Towards solving

Datapoint: test_exchange_fees.py was previously /tests/flows/test_exchange_swap_fees.py. I found that the old test file also erroneously had 18 decimals.

[`helper_functions.py::convert_bt_amt_to_dt()`](https://github.com/oceanprotocol/ocean.py/blob/4da2ade35660d92b39d75710387f57cceb8c3c87/tests/resources/helper_functions.py#L304) smells a bit off. Below is its code. (Note that I'd inserted in the comment '"`## FIXME: SHOULDN'T 18 BE bt_decimals ??`"

```python
def convert_bt_amt_to_dt(
    bt_amount: int,
    bt_decimals: int,
[...]

PR #1200, opened 2022-12-13
#1190 part one: fix Datatoken

trentmc:
Part 1/2 of #1190.

This part just fixes Datatoken.

Part 2/2 will revive test_exchange, properly


trentmc:
Observed: test_buy_from_fre_and_order is failing. I'll need to fix that, or skip the test then fix later.

PR #1214, opened 2022-12-15
Simplify DatatokenEnterprise interface

### LINKED ISSUE #1207: Unwieldy interfaces for DatatokenEnterprise.buy_from_dispenser_and_order() and buy_from_fre_and_order()
### Background / motivation

We have an opportunity to improve the interfaces to DatatokenEnterprise.buy_from_dispenser_and_order() and buy_from_fre_and_order(). As well as setting up the messages for ordering. 

Using them is crazy complex - see [test_datatoken_enterprise.py](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/models/test/test_datatoken_enterprise.py). 

It takes 50 lines to do what should take 1-2 lines with good defaults.

### TODO

Add new methods to datatoken_enterprise.py::DatatokenEnterprise.py that provide a cleaner interface to buy_from_dispenser_and_order(), buy_from_fre_and_order(); and account for cleaner setting-up of messages to do the order.

calina-c:
Closes #1207.


ocean_lib/models/test/test_datatoken_enterprise.py
```diff
@@ -114,18 +90,10 @@ def test_buy_from_dispenser_and_order(
     _ = DT.buy_from_dispenser_and_order(
```
trentmc, line 90: Right now, the call to this method is returning a "_". But we should show what it returns. Is it order_tx_id? More?

Also, after this call, I think there should probably be some basic sanity test that the order_tx_id is valid.

ocean_lib/models/datatoken_enterprise.py
```diff
@@ -18,37 +21,37 @@ def buy_from_fre_and_order(
...
-        provider_fee_amount: Union[int, str],
-        v: int,
-        r: Union[str, bytes],
-        s: Union[str, bytes],
-        valid_until: int,
-        provider_data: Union[str, bytes],
+        provider_fees: dict,
         consume_market_order_fee_address: str,
         consume_market_order_fee_token: str,
         consume_market_order_fee_amount: int,
```
trentmc, line 27: This is best as some sort of "fees" class. It's a recurring pattern that holds a 3-tuple of fees info: address, token, amount. Not just for consume_market_order_fee, but other things. Call it e.g. "FeeGroup".

I recommend: create such a fees class, and use it liberally to replace input arguments (like here), and to replace raw "getter" calls to smart contracts such as DT.getPublishingMarketFee().

You can look to fixed_rate_exchange.py::ExchangeDetails class, and fixed_rate_exchange.py::FeesInfo class for inspiration. E.g. copy & paste one of those, then chop it down into holding 3 attributes ( address, token, amount).

trentmc (changes requested):

tests/resources/helper_functions.py
```diff
@@ -350,3 +351,37 @@ def int_units(amount, num_decimals):
...
+    return {
+        "providerFeeAddress": provider_fee_address,
+        "providerFeeToken": provider_fee_token,
+        "providerFeeAmount": provider_fee_amount,
+        "v": signature.v,
+        "r": signature.r,
+        "s": signature.s,
+        "validUntil": valid_until,
+        "providerData": Web3.toHex(Web3.toBytes(text=provider_data)),
+    }
```
trentmc, line 387: Having a helper method for _tests_ to abstract away the messiness of messages & signing doesn't work for actual _users_. 

We need to properly abstract away this messiness for actual users.

trentmc (changes requested):
OK, I'm done reviewing for this round

tests/resources/helper_functions.py
calina-c: this is not it. We use a response from provider when we use provider fees in ocean.py. However, for tests, not to publish and call provider every time, we simulate a message from provider with the right data.

calina-c: the main issue with previous tests were that they were not simulating the response from provider, but instead doing all the signing and generating for every parameter. Now it simulates provider better and in just one place. It is just a test method, much like a fixture, and actual users don't need it.

ocean_lib/models/datatoken_enterprise.py
calina-c: I would do that in a next iteration. I can mark this PR as "work on ..." instead of "closes ..." and I propose we merge it in (of course after fixing the other comments), then I'll do the fee grouping in a different PR to help with conflict management. Would that work?

trentmc: Yes! 

I'd actually suggested the exact same thing in slack :)

trentmc (approved):

ocean_lib/models/test/test_datatoken_enterprise.py
```diff
@@ -87,24 +77,10 @@ def test_buy_from_dispenser_and_order(
         {"from": publisher_wallet},
     )
 
-    provider_fee_address = publisher_wallet.address
-    provider_fee_token = DAI.address
-    provider_fee_amount = 0
-    provider_data = json.dumps({"timeout": 0}, separators=(",", ":"))
     valid_until = 1958133628  # 2032
```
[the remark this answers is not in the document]
calina-c: in non-unit tests, it is exactly what we do, see e.g. test_reuse_order_fees, where we create the timestamp on the fly and send it towards the model class. For unit testing purposes we added a timestamp directly.

Would you like:
- the function itself to take in a datetime? -> I would not recommend it, because I'd prefer we stay closer to the API
- create this known `valid_until` value in tests, on the fly, instead of directly writing it?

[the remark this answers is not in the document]
trentmc: The function itself should still take the Unix time, that's ok. Just compute it here in the test using datetime. Should be a short single line.

calina-c:
@trentmc this is ready to be reviewed again after adding the interfaces for dispense_and_order and buy_from_exchange_and_order in regular Datatoken template.

ocean_lib/models/datatoken.py
```diff
@@ -311,6 +307,96 @@ def build_access_service(
...
+            ),
+            consume_market_order_fee_token=(
+                consume_market_order_fee_token or ZERO_ADDRESS
+            ),
+            consume_market_order_fee_amount=consume_market_order_fee_amount,
+            transaction_parameters=transaction_parameters,
+        )
+
+    @enforce_types
+    def buy_from_exchange_and_order(
```
trentmc, line 361: I realized that a better label yet is "buy_DT_and_order". 
- more consistent with "buy_DT()"
- more precise. (Are you buying a DT or basetoken?)

ocean_lib/models/test/test_datatoken_enterprise.py
trentmc: This is now 
```python
_ = DT.buy_from_exchange_and_order(...)
```

But the comment above still needs to be resolved.

ocean_lib/models/datatoken_enterprise.py
trentmc: I see you've got partway here (for provider fees), by having a "provider_fees" dict. It already greatly simplified things.

But you'll notice that by having the dict
- it doesn't have an easy way to create defaults
- when grabbing parameter values it's slightly sloppy. E.g. `provider_fees["providerFeeAddress"]`

If you create a class from it, ie have a `ProviderFees` class, then you will be able to resolve both
- defaults are easy
- and cleaner to grab parameter values, e.g. `provider_fees.address` (`provider_fees` is an instance of `ProviderFee`)

Also, you'll want a class for `ConsumeMarketOrderFees` too. Whereas `ProviderFees` will have 6+ attributes, `ConsumeMarketOrderFees` will just have 3 or so.

Like mentioned above, you can look to fixed_rate_exchange.py::ExchangeDetails class, and fixed_rate_exchange.py::FeesInfo class for inspiration. 

(This is probably the largest single remaining thing to do. And it will help a lot.)

trentmc (changes requested):
This is looking really great. I provided some comments, once we're through this last 10% or so it's gonna be excellent, and I will be happy to approve. Great stuff.

ocean_lib/models/datatoken_enterprise.py
calina-c: I don't agree with changing provider fees to a class. We get them from provider as a dict, through the response itself, so we can use them straight through. There is no case where we "build" the provider fees except for tests where we emulate the provider. I will do it if needed but I think it is overengineering in this case

trentmc: OK. Those are good arguments to keep provider fees as a dict. Fine, then:)

What about ConsumeMarketOrderFees?

trentmc: UPDATE: we'll put ConsumeMarketOrderFees etc into another github issue

trentmc (approved):

ocean_lib/models/datatoken_enterprise.py
trentmc: Github issue: https://github.com/oceanprotocol/ocean.py/issues/1222

PR #1220, opened 2022-12-20
Fix #1222: Fee objects, to make fees less cumbersome

### LINKED ISSUE #1222: Fee tuples are cumbersome, in both arguments and return tuples
### Background / motivation

Many methods have arguments like:
```python
        consume_market_order_fee_address: str,
        consume_market_order_fee_token: str,
        consume_market_order_fee_amount: int,
```

It's cumbersome to track the three related variables, and to enter them.

Also, fee "getters" from smart contracts typically return a 3-tuple, and sometimes larger tuples. This is cumbersome too.

### Suggested fix

This is best as some sort of "fees" class. It's a recurring pattern that holds a 3-tuple of fees info: address, token, amount. Not just for consume_market_order_fee, but other things. Call it e.g. "FeeGroup" or "ConsumeMarketOrderFees".

Recommended: create such a fees class, and use it liberally to replace input arguments (like here), and to replace raw "getter" calls to smart contracts such as DT.getPublishingMarketFee().

We can look to fixed_rate_exchange.py::ExchangeDetails class, and fixed_rate_exchange.py::FeesInfo class for inspiration. E.g. copy & paste one of those, then chop it down into holding 3 attributes ( address, token, amount).

calina-c:
Fix #1222 "Fee tuples are cumbersome, in both arguments and return tuples"

Continue cleanup from [PR#1214](https://github.com/oceanprotocol/ocean.py/pull/1214), as a follow-up of orders by using fee token triples.


trentmc:
I created a github issue for this PR (# 1222), and brought in comments from the previous PR's discussion that led to this issue/PR. 

In this PR, I linked to the new github issue (# 1222).

----

I reviewed this PR. In general this looks really great to clean up the "arguments" side.

I also see that there are "getters" on the smart contracts that return the same three fee values, as a tuple. It would be useful to convert the return values of such getters to the same object as well.

(And the class name would have to generalize beyond "_Arguments" too.)

trentmc (changes requested):
Change requested - see my other comment

calina-c:
> I created a github issue for this PR (# 1222), and brought in comments from the previous PR's discussion that led to this issue/PR.
> 
> In this PR, I linked to the new github issue (# 1222).
> 
> I reviewed this PR. In general this looks really great to clean up the "arguments" side.
> 
> I also see that there are "getters" on the smart contracts that return the same three fee values, as a tuple. It would be useful to convert the return values of such getters to the same object as well.
> 
> (And the class name would have to generalize beyond "_Arguments" too.)

Well the getters are specifically designed to be used for tuple conversion when used in contract calls, in that format, to avoid duplication. I don't understand the other requirement, what class should be returned where?

ocean_lib/models/arguments.py
```diff
@@ -18,6 +20,28 @@
...
+    ):
+        self.address = (
+            Web3.toChecksumAddress(address.lower()) if address else ZERO_ADDRESS
+        )
+        self.token = Web3.toChecksumAddress(token.lower()) if token else ZERO_ADDRESS
+
+        self.amount = amount
+
+    def to_tuple(self):
+        return (
```
trentmc, line 38: Why does it need to apply lower() and checksum() if it's already stored as these?

ocean_lib/models/arguments.py
```diff
@@ -18,6 +20,28 @@
...
+            Web3.toChecksumAddress(self.address.lower()),
+            Web3.toChecksumAddress(self.token.lower()),
+            self.amount,
+        )
+
```
trentmc, line 43: Add str() operator like FeesInfo and others do. It really helps for usability.

ocean_lib/models/arguments.py
```diff
@@ -18,6 +20,28 @@
 logger = logging.getLogger("ocean")
 
 
+class FeeTokenArguments:
```
trentmc, line 23: Like recommended before: this class is not only used as arguments. It's also used via getters. 

Therefore rename appropriately.

Also, move it out of the arguments module (or rename the module to something like structs.py)

ocean_lib/ocean/ocean_assets.py
```diff
@@ -602,9 +588,11 @@ def pay_for_compute_service(
             self._start_or_reuse_order_based_on_initialize_response(
                 datasets[i],
                 item,
-                consume_market_order_fee_address,
-                datasets[i].consume_market_order_fee_token,
-                datasets[i].consume_market_order_fee_amount,
+                FeeTokenArguments(
```
trentmc, line 591: I think that the caller of this methods could construct the fees object, no?

Then it's much simpler within this method. 

Similar for other methods. The fee object should be set at a higher level, and then passed as args to lower level smart contract calls.

ocean_lib/models/arguments.py
```diff
@@ -6,6 +6,8 @@
 import warnings
 from typing import Any, Dict, List, Optional
 
+from web3.main import Web3
```
trentmc, line 9: There should be simple unit tests for each class in this module. Eg via a test module test_arguments.py

ocean_lib/models/arguments.py
```diff
@@ -18,6 +20,28 @@
...
+        self.amount = amount
+
+    def to_tuple(self):
+        return (
+            Web3.toChecksumAddress(self.address.lower()),
+            Web3.toChecksumAddress(self.token.lower()),
+            self.amount,
+        )
+
+
```
trentmc, line 44: I didn't see anywhere in this PR the changes to fee "getters" calls to smart contracts, where the getter call currently returns a 3-tuple. We should make it where the getter call returns the fees object.

trentmc (changes requested):

trentmc:
> > [Trent] I also see that there are "getters" on the smart contracts that return the same three fee values, as a tuple. It would be useful to convert the return values of such getters to the same object as well.
> > (And the class name would have to generalize beyond "_Arguments" too.)
> 
> [Calina] Well the getters are specifically designed to be used for tuple conversion when used in contract calls, in that format, to avoid duplication. I don't understand the other requirement, what class should be returned where?

The comment was [here](https://github.com/oceanprotocol/ocean.py/pull/1214#discussion_r1049572174).

For convenience, I will copy below.

---

I recommend: create [a] fees class, and use it liberally to replace input arguments (like here), and to replace raw "getter" calls to smart contracts such as DT.getPublishingMarketFee().

You can look to fixed_rate_exchange.py::ExchangeDetails class, and fixed_rate_exchange.py::FeesInfo class for inspiration. E.g. copy & paste one of those, then chop it down into holding 3 attributes ( address, token, amount).

trentmc (changes requested):
There are still 4-5 outstanding comments from my previous review.

ocean_lib/ocean/ocean_assets.py
calina-c: here we are iterating over the response from provider, datasets + algorithm, so we don't know from the caller, beforehand, which tokens and amounts correspond. We only know the fee address, which is why we create the FeeTokenInfo here, where we have the knowledge.

calina-c:
@trentmc I'm trying to dissolve arguments.py entirely and move everything into its own corresponding class. I think I can get it working. It was a previous idea and it caused some issues but I think we're starting to see even more cost from the current implementation (i.e. a lot of in-place imports). It's probably not going to work on first try/first commit, but I'll iterate today.

In the meantime, I also removed the double-checksumming and added `__str__`, so those can be resolved. They are moved in datatoken.py but done :)

ocean_lib/models/test/test_datatoken_order_both_templates.py
```diff
@@ -220,19 +221,24 @@ def test_buy_DT_and_order(
...
+        "exchange": exchange,
+        "transaction_parameters": {"from": publisher_wallet},
+    }
+
+    if template_index == 2:
+        args["max_base_token_amount"] = to_wei(2.5)
+        args["consume_market_swap_fee_amount"] = to_wei(0.001)  # 1e15 => 0.1%
+        args["consume_market_swap_fee_address"] = another_consumer_wallet.address
+
+    tx = DT.buy_DT_and_order(**args)
```
trentmc, line 241: Passing in `**args` to `buy_DT_and_order()` is harder to understand than before. Especially when `args` is still so big.

I recommend to have two unit tests for `buy_DT_and_order()`: one with good defaults and one for non-defaults. 

Here's a good-defaults call:

```python
order_tx_id = exchange.buy_DT_and_order({"from": publisher_wallet})
```

Notes:
- calls 'exchange` (`Exchange` class), not `DT` (`Datatoken` class). Just like `buy_DT()`. And `exchange` knows which datatoken to use, by definition
- it returns `order_tx_id`, not `tx`. Like calls to `pay_for_access_service()`.

Here's how the defaults would be set:
```python
consumer = transaction_parameters["from"].address
service_index = 1
provider_fees = (default to zero)
consume_market_fees = (default to zero)
```

And, here's a non-defaults call:
```python
provider_fees  = get_mock_provider_fees("MockDAI", publisher_wallet, valid_until=valid_until)
consume_market_fees": FeeTokenInfo(address=consume_fee_address, token=DAI.address)
order_tx_id = exchange.buy_DT_and_order(
  provider_fees=provider_fees,
  consume_market_fees=consume_market_fees,
  service_index=1,
  tx_dict={"from": publisher_wallet}
)
```

ocean_lib/ocean/ocean_assets.py
```diff
@@ -602,9 +583,11 @@ def pay_for_compute_service(
...
                 datasets[i],
                 item,
-                consume_market_order_fee_address,
-                datasets[i].consume_market_order_fee_token,
-                datasets[i].consume_market_order_fee_amount,
+                FeeTokenInfo(
+                    consume_market_order_fee_address,
+                    datasets[i].consume_market_order_fee_token,
+                    datasets[i].consume_market_order_fee_amount,
+                ),
```
trentmc, line 590: I see that `consume_market_order_fee_address` is an argument to this method (`pay_for_compute_service`). Yet `consume_market_order_fee_token` and `.._amount` are in each dataset `datasets[i]`. 

Could we have it where: don't have argument `consume_market_order_fee_address`, and instead each dataset holds a `FeeTokenInfo` class ?

trentmc (changes requested):

ocean_lib/assets/test/test_asset_downloader.py
```diff
@@ -171,9 +172,10 @@ def test_ocean_assets_download_destination_file(
         consumer=publisher_wallet.address,
         service_index=ddo.get_index_of_service(access_service),
         provider_fees=provider_fees,
-        consume_market_order_fee_address=publisher_wallet.address,
-        consume_market_order_fee_token=datatoken.address,
-        consume_market_order_fee_amount=0,
+        consume_market_fees=TokenFeeInfo(
+            address=publisher_wallet.address,
+            token=datatoken.address,
+        ),
```
trentmc, line 178: I recommend constructing `consume_market_fees = FeeTokenInfo(..)` right above the call to `datatoken.start_order()`. It will make the method call easier to read.

ocean_lib/ocean/ocean_assets.py
trentmc: Note: Calina said: There's no way to build the arguments beforehand, we only know the fee address until we retrieve the rest from provider

So I have my answer here. Resolving this.

trentmc (changes requested):
.

ocean_lib/models/test/test_datatoken_order_both_templates.py
```diff
@@ -6,7 +6,7 @@
 
 import pytest
```
trentmc, line 7: ### test_dispense_and_order_with_defaults()

We can optimize the defaults for this even better.

Status quo:

```python
    tx = DT.dispense_and_order(
        consumer=consumer_wallet.address,
        service_index=1,
        provider_fees=provider_fees,
        transaction_parameters={"from": publisher_wallet},
    )
```

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
order_tx_id = DT.dispense_and_order(provider_fees, {"from": publisher_wallet})
```

Where:
- it returns `order_tx_id`, not `tx`. Like calls to `pay_for_access_service()`.

With defaults set as:
```python
consumer = transaction_parameters["from"].address
service_index = 1
```

### test_buy_DT_and_order()

We should split test function this into two test functions: one with good defaults, and one with nondefaults. 

### [new] test_buy_DT_and_order_with_defaults()

Status quo: 
```python
    args = {
        "consumer": another_consumer_wallet.address,
        "service_index": 1,
        "provider_fees": provider_fees,
        "consume_market_fees": FeeTokenInfo(
            address=consume_fee_address,
            token=DAI.address,
        ),
        "exchange": exchange,
        "transaction_parameters": {"from": publisher_wallet},
    }

    if template_index == 2:
        args["max_base_token_amount"] = to_wei(2.5)
        args["consume_market_swap_fee_amount"] = to_wei(0.001)  # 1e15 => 0.1%
        args["consume_market_swap_fee_address"] = another_consumer_wallet.address

    tx = DT.buy_DT_and_order(**args)
```

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
order_tx_id = exchange.buy_DT_and_order(provider_fees, {"from": publisher_wallet})
```

Where:
- the call is to 'exchange` (`Exchange` class), not `DT` (`Datatoken` class). Just like `buy_DT()`. And `exchange` knows which datatoken to use, by definition
- it returns `order_tx_id`, not `tx`. Like calls to `pay_for_access_service()`.

With defaults set as:
```python
consumer = transaction_parameters["from"].address
service_index = 1
consume_market_fees = <zero>
```

### [new] test_buy_DT_and_order_with_non_defaults()

Status quo: (like shown above)

The "args" setup makes it harder to read. Let's remove that. And, it should also return `order_tx_id` like the defaults flow. So, it will look something like:

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
consume_market_order_fees = FeeTokenInfo(address=consume_fee_address, token=DAI.address)
consume_market_swap_fees = FeeTokenInfo(to_wei(0.001), address=another_consumer_wallet.address, token=..)
max_base_token_amount = to_wei(2.5)
order_tx_id = exchange.buy_DT_and_order(
        consumer = another_consumer_wallet.address,
        service_index = 1,
        max_base_token_amount = max_base_token_amount,
        provider_fees = provider_fees,
        consume_market_order_fees = consume_market_order_fees,
        consume_market_swap_fees = consume_market_swap_fees,
        transaction_parameters = {"from": publisher_wallet},
)
```

Where:
- both template 1 and 2 should be aware of max_base_token_amount and consume_market_swap_fees (not just template 2 like the status quo)

calina-c: > optimising to: DT.dispense_and_order(provider_fees, {"from": publisher_wallet})
this test uses the consumer_wallet for the consumer address, while the sender of the transaction is publisher_wallet. I think we the consumer is NOT by default the sender of the transaction. Especially because we mimic DatatokenEnterprise here.

Same comment for the exchange test. Even more so, since we have another_consumer_wallet as the consumer.

Ref: moving this class to Exchange and figuring out there whether to do a direct call from DatatokenEnterprise template vs. combo with multiple txs from Datatoken: I can do that, but there is a code smell here. If the max_base_token_amount, consume_market_swap_fee_amount and consume_market_swap_fee_address are not being used with a regular Datatoken (only used in Enterprise), then what is their purpose? Have I made mistakes in replicating the functionality of buy DT and order in the regular template? Is there some transaction I'm missing that would involve these parameters?

Also, ref. returning the tx_id instead of the transaction itself: pay_for_access_service is a higher level function used in ocean assets. For models that do contract calls, we either return the receipt or wrapped models where suitable (i.e. when returning a new Datatoken or tuple of DataNFT, Datatoken, Exchange objects). We should not switch to tx_ids for functions that reside inside models, since receipts are far more rich and usable. Or, if we do, we need to level everything and do a uniform refactor for all (which should probably be another issue)

trentmc: I'm not sure what the best answer is on this. To really know, I'd have to play around with the possibilities more myself.

Given that we aren't yet in sync on this, and it would be really useful to "ratchet up" ocean.py with all the great improvements of this PR, I recommend:

- merge the PR as-is
- copy these comments into a github issue for further discussion / playing with possibilities. (And probably lower priority than some other things)

PR #1228, opened 2022-12-21
Fixed config creation in setup-local.md

### LINKED ISSUE #1227: Mismatch between imported function and actual code in setup-local.md
For `setup-local.md`, more exactly for config creation, there is `get_config_dict` function which is imported, but not used. Instead of it, there is used the deprecated class `ExampleConfig`.

```
config = ExampleConfig.get_config("development")
```

mariacarmina:
Fixes #1227  .

Changes proposed in this PR:

- fixed the config creation in `setup-local.md`


trentmc (approved):
Good catch!

PR #1229, opened 2022-12-22
Fix README tests after refactoring

calina-c:


calina-c:
We will need to reinstate publish-flow-restapi another time, this is a blocker for a few other PRs. We can create a follow-up issue for that test in particular.

READMEs/main-flow.md
```diff
@@ -127,7 +132,7 @@ data_nft = DataNFT(config, data_nft_address)
 
 It's similar for Datatokens. In Python:
 
-```python
+```console
```
trentmc, line 135: python

trentmc (approved):
Looks great! Just a couple tiny tweaks, then good to go. Therefore I approved in advance.

READMEs/main-flow.md
```diff
@@ -118,7 +123,7 @@ Bonus: this README's appendices expand on the steps above with further flexibili
 
 Anytime in the future, you can reconstruct your data NFT as an object in Python, via:
 
-```python
+```console
```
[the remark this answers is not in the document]
calina-c: if I render them in python, they are picked up by the auto-readme creation mechanism and they are NOT valid python code with `...` and `<>`. Which is why, for them to be avoided by the mechanism, I added them as `console`. Do you suggest another way to formulate this or maybe change it to runnable Python code?

READMEs/main-flow.md
calina-c: Same comment here.

PR #1230, opened 2022-12-22
fix pytest workflow

mariacarmina:
Fixes # .

Changes proposed in this PR:

- updated pytest.yml to use correctly the envvar


calina-c:
@mariacarmina I cherry-picked this into #1229 because:
- your readme tests on this branch fail due to them being outdated (which I'm fixing in #1229) so you need my changes
- my workflow file was failing on that branch so I needed your changes

I decided since this is the shorter one, to cherry-pick yours into mine instead of the other way around.

Issue #1165, opened 2022-12-07
It's hard for py devs to see contract methods of DataNFT, Datatoken, etc


trentmc:
### Motivation

Recently, ocean.py got brownie support. This meant we could (and did) deprecate all the ocean.py methods that simply passed-through to smart contracts. 

For example, Datatoken.balanceOf(). And probably 200 other methods across modules in `ocean_lib/models/*.py`.

This _greatly_ reduced the amount of code to maintain & debug. It also made a smoother UX.

However, we _did_ lose one thing. Once developers go beyond the ocean.py READMEs, they don't have a great reference for what methods are available. They must resort to (a) looking at unit tests (b) looking at solidity code.

Can we...
- help developers to understand the smart contract methods that DataNFT, Datatoken, etc expose
- while maintaining the benefits of brownie

### Suggested Approach

For each module / class in `ocean_lib/models/*.py`, add a triple-quote string `"""foo bar"""` that spans many lines. Inside it, put the method definitions for each smart contract method, written in a py-like fashion (to maximize approachability to py devs).

E.g. for datatoken.py's Datatoken class:
```python

class Datatoken:
   # have the usual non-smart-contract methods first. Then...
 
"""
  def balanceOf(address: str) -> int:
    Returns the balance of this address.
    :param: address of interest
    :return: balance, in wei 

  def approve(...
"""
   
```

trentmc:
Note: the fix for #1095 included a new main flow, and its appendix nicely references what interfaces developers can use for data NFTs and datatokens. That will help soften the concern of this issue.

However, we should still do as this issue prescribes: a docstring in each py module listing the smartphone contract interfaces. That will help further.

calina-c:
I recommend `@method` annotations instead. https://manual.phpdoc.org/HTMLframesConverter/default/phpDocumentor/tutorial_tags.method.pkg.html

trentmc:
> I recommend something like `@method` annotations instead. https://manual.phpdoc.org/HTMLframesConverter/default/phpDocumentor/tutorial_tags.method.pkg.html
> 
> There has to be something similar in Python!

I [commented in the PR](https://github.com/oceanprotocol/ocean.py/pull/1270#issuecomment-1377133706).

calina-c:
@alexcos20 @trentmc I have completed the first step, an inventory of the methods needing documentation. I have some questions, so I'll include the report below. Please help with any information/guidance you may see fit.

### General issues
[...]

PR #1270, opened 2023-01-10
Remove magic method and transfer functions directly.

calina-c:
Helps #1165. Turns out both magically calling the functions and transferring them DOES keep their `__doc__`. However, only transfering them keeps them in the dir(class) function. This should now be compatible with autocompletions etc.


trentmc:
I'm confused. How does this solve #1165? 

For example: if I'm a python developer using ocean.py, where do I go to learn about the interface for `Datatoken.transfer()?`

If the answer is "Go look at the solidity code" then it isn't solved.

If the answer is "It's in the docstring" but we don't expose the docstring anywhere, then it isn't solved.

calina-c:
@trentmc 

I am attaching two screenshots of various ways the user can learn about the contract structure without looking at the abi directly. You can see the contract is a datatoken, but it can be any type of object instantiating ContractBase.

You can use `dir(contract)` to show its attributes, as well as `contract.__dict__`. From the list of attributes, you can select any function and get its interface. i.e. 

```
> contract.balanceOf
> <ContractCall 'balanceOf(address account)'>
```

As well as the docstring (taken from solidity)

```
> contract.balanceOf.__doc__
> '\n    A public view or pure contract method.\n\n    Attributes\n    ----------\n    abi : dict\n        Contract ABI specific to this method.\n    signature : str\n        Bytes4 method signature.\n    '
```

Screenshot 2023-01-10 at 15 00 29
Screenshot 2023-01-10 at 15 01 01

trentmc:
> You can use dir(contract) to show its attributes, as well as contract.__dict__

Thanks. FYI I was aware of these even when writing the function.

Those are only of limited use. I sometimes use them during interactive debugging.

But they do not meet the goals of this issue, which is to have a great reference for what methods are available. E.g. 
```python
"""
  def balanceOf(address: str) -> int:
    Returns the balance of this address.
    :param: address of interest
    :return: balance, in wei 
"""
```

I don't see it as a huge effort to do this. It will be manual and tedious, but in the end probably only take a few days. 

And it will serve as a great jumping-off point for the follow-up issue of having great docs.

calina-c:
- They weren't on the contract previously, so the interface should be much better now. I will keep this change if possible
- I'll also add the docstrings. I thought the purpose was for better debugging and autocomplete, but I can write these as well.

trentmc:
> They weren't on the contract previously, so the interface should be much better now. I will keep this change if possible

Sounds good.

> I'll also add the docstrings. 

Sounds good.

> I thought the purpose was for better debugging and autocomplete, but I can write these as well.

Having both is complementary. Glad we're in sync:)

trentmc (approved):

trentmc:
FYI I approved this PR. And changed the description from "solves X" to "helps X"

calina-c:
I re-requested a review now with docstrings added :)

PR #1272, opened 2023-01-11
Fix hanging CI tests

### LINKED ISSUE #1213: Txs to polygon hang. Why: gas price it tried << gas price needed
### Background

I'm working with polygon network to grab predict-eth round 2 results. I'm in the step where I'm given a datatoken address, and I need to call `ocean.assets.download_asset()`. Polygon block latency is a few seconds. However it takes me 2-5 min for a tx to go through. 

I observed that the gas price I'm giving (31.9058) is much lower than what polygon matic gas station suggests. That's the problem! (Or at least one of them). Details below.

### TODOs for this issue

Ensure that we pay a more reasonable gas price in Polygon, so that users aren't waiting around. (Without reinventing the wheel. Brownie will likely make this easy, if we're smart about it.)

### Details of my call

This is in `predict-eth-judges` repo.

```text
(venv) trentmc@trentmc-xps:~/code/pej$ pej getpreds outdir-round2/
pej getpreds: Begin
Args: CSV_DIR=outdir-round2/
alice_wallet.address=0xA54ABd42b11B7C97538CAD7C6A2820419ddF703E. bal=9.928745801526913 MATIC
Loaded 2 DT addrs.
DT #1/2: Begin
dt_addr=0xb8187192f79E88eF942E50Ba433B71b6d226D5e1, nft_addr=0x53105b05E5a78B26e70aE6E6169C7e9CDe15b47A, did=did:op:fe8ffa177b719476bef1b8cbfb28b4e8043515817304812f018554afd04d7ef0
INFO:ocean_lib.data_provider.data_service_provider:Service initialized successfully initializeEndpoint https://v4.provider.polygon.oceanprotocol.com/api/services/initialize
Transaction sent: 0x2cf67f1596a71bd47f0e61ec7242a6dd9b2ed09324863c61f4d5b567dc59564a
  Gas price: 31.905822902 gwei   Gas limit: 96654   Nonce: 8
```

The key thing to note here is "Gas price: 31.905822902 gwei".

### Details of gas station

And, doing a call (as of Dec 14, 21:00 UCT) to https://gasstation-mainnet.matic.network/v2 outputs the following: `{"safeLow":{"maxPriorityFee":35.848027238,"maxFee":44.510349484},"standard":{"maxPriorityFee":36.741326008,"maxFee":45.403648254000004},"fast":{"maxPriorityFee":39.64206791064286,"maxFee":48.30439015664286},"estimatedBaseFee":8.662322246,"blockTime":3,"blockNumber":36834265}`

Note that even the lowest of these values is _much_ higher than the gas price that our tx gave.

### Details: my current brownie network-config.yaml

```text
development:
  - name: development
    chainid: 8996
    host: http://127.0.0.1:8545
    id: development
    required_confs: 1
live:
  - name: polygon-test
    networks:
      - chainid: 80001
        host: https://rpc-mumbai.maticvigil.com
        id: polygon-test
        required_confs: 1
  - name: polygon-main
    networks:
      - chainid: 137
        host: https://polygon-rpc.com
        id: polygon-main
        required_confs: 15
```

mariacarmina:
Fixes #1213  .

Changes proposed in this PR:

- added GasNowScalingStrategy.
-  removed priority fee util function.
- set max_gas_price to create data NFT tx dict.


tests/integration/remote/util.py
```diff
@@ -105,8 +107,12 @@ def do_ocean_tx_and_handle_gotchas(ocean, alice_wallet):
 
     print("Call create() from data NFT, and wait for it to complete...")
     try:
+        tx_dict = {"from": alice_wallet}
```
calina-c, line 110: I suggest creating the entire tx dict at once since there is no conditional here.

I.e.
```
data_nft = ocean.data_nft_factory.create(
            DataNFTArguments(symbol, symbol), {"from": alice_wallet, "gas_price": gas_strategy.max_gas_price}
        )
```

calina-c (commented):
Just a minor form comment, I'll approve after all tests run with these changes.

tests/integration/remote/util.py
mariacarmina: Ok thank you

mariacarmina:
In previous [run](https://github.com/oceanprotocol/ocean.py/actions/runs/3891919478/jobs/6642727398#step:7:17), polygon test pass individually. There must be another cause for these hanging tests, I suspect that polygon test is not only the cause for CI freezing, because individually it passed. Maybe the tests are too heavy to accomplish by the CI or maybe barge freezes.

mariacarmina:
After some research, the problem comes from the fact that in `test_data_nft.py`, we try to catch reverted messages from the contract calls -> brownie freezes -> ganache freezes as well.

Snapshot of freezing brownie (KeyboardInterrupt):
```console
ocean_lib/models/test/test_data_nft.py 
>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>
> /home/maria/Desktop/ocean.py/ocean_lib/models/test/test_data_nft.py(64)test_permissions()
-> wrong_tx = data_nft.cleanPermissions({"from": another_consumer_wallet})
(Pdb) n
Transaction sent: 0xc932c8257d4d29ded232324c3c42ba27db968514f0c2eb91f396a48650ff7361
  Gas price: 0.0 gwei   Gas limit: 10000000000   Nonce: 4
^CKeyboardInterrupt
> /home/maria/Desktop/ocean.py/ocean_lib/models/test/test_data_nft.py(64)test_permissions()

```
Following this [article](https://n4n0b1t3.medium.com/solution-brownie-testing-for-reverted-transactions-does-not-work-with-pytest-raises-c4d2888db6f6) and this PR [discussion](https://github.com/eth-brownie/brownie/issues/1441), I noticed that the requirement which was not satisfied is that we use Python3.8 instead of Python3.9 with the latest version of brownie. When I updated the Python version from CI, the model tests worked as expected and the exceptions are caught.
So the problem wasn't due to polygon test. I will update the title of the issue.
cc @calina-c @trentmc

ocean_lib/models/test/test_data_nft.py
```diff
@@ -714,80 +619,19 @@ def test_fail_create_datatoken(
...
-                symbol="DT1Symbol",
-            ),
-            {"from": another_consumer_wallet},
-        )
-
-
-@pytest.mark.unit
-def test_datatoken_cap(publisher_wallet, consumer_wallet, data_nft_factory):
-    # create NFT with ERC20
-    with pytest.raises(Exception, match="Cap is needed for Datatoken Enterprise"):
```
calina-c, line 755: this is our in-house Exception, it has nothing to do with the contracts. Keep it, please

mariacarmina:
The reason that the tx froze, was that `required_confs` was set to 1, instead of `0`. The reverted message didn't show up yet, still investigating.

mariacarmina:
**Workaround**
Found the fix for this. Firstly, if the tx reverted, I retrieve pending transaction receipt with `required_confs` set to `0`, then wait separately for the tx to fetch all the info. If the tx was reverted, it got stuck in a while loop from brownie code, waiting for confirmations, that's why `required_confs` need to be `0` in this case, in order to avoid infinite loop. For reference, I let here the paragraph from [brownie doc](https://eth-brownie.readthedocs.io/en/stable/core-accounts.html#broadcasting-multiple-transactions).

```console
Broadcasting a transaction is normally a blocking action - 
Brownie waits until the transaction has confirmed before continuing. 
One way to broadcast transactions without blocking is to set required_confs = 0. 
``` 
The second problem that I encounter was that I could not retrieve the error message, the part which is the scope of the test. The `revert_msg` was None, also the `dev_revert_msg` was None and the only related thing was that the tx status was `0`, which means reverted, but without any specific reason. Taking this situation in consideration, I used web3 package from brownie in order to interrogate the previous block with `web3.eth.call()` and afterwards to get the error message. To catch the error, I still respected `pytest.raises()` format.

Output snippet of the reverted tx regarding brownie:
```json
"op":"REVERT",
"pc":42,
"stack":[
   "0000000000000000000000000000000000000000000000000000000000000084",
   "0000000000000000000000000000000000000000000000000000000000000000"
],
"storage":"None"
}
],
"_trace":"None",
"_events":{
   "CleanedPermissions":[
      "OrderedDict("[
         "(""signer",
         "0x529043886F21D9bc1AE0feDb751e34265a246e47"")",
         "(""timestamp",
         1673599595),
         "(""blockNumber",
         47)
      ]")"
   ],
   "AddedManager":[
      "OrderedDict("[
         "(""user",
         "0x529043886F21D9bc1AE0feDb751e34265a246e47"")",
         "(""signer",
         "0x529043886F21D9bc1AE0feDb751e34265a246e47"")",
         "(""timestamp",
         1673599595),
         "(""blockNumber",
         47)
      ]")"
   ]
},
"_return_value":"None",
"_revert_msg":"None",
"_dev_revert_msg":"None",
"_modified_state":"None",
"_new_contracts":"None",
"_internal_transfers":"None",
"_subcalls":"None",
"sender":"<LocalAccount""0x529043886F21D9bc1AE0feDb751e34265a246e47"">",
"status":"<Status.Reverted":0>,
"txid":"0x684012f6e2195c848c29f71f5db3ed88775d1ab7dd9fa85742195e0f61232c75",
"contract_name":"None",
"fn_name":"",
"_revert_pc":"None",
"receiver":"0xcf293068b28f99622FFE50C9C6f3e7046Ecda8d9",
"value":0,
"gas_price":0,
"max_fee":"None",
"priority_fee":"None",
"gas_limit":10000000000,
"input":"0x23b872dd00000000000000000000000002354a1f160a3fd7ac8b02ee91f04104440b28e7000000000000000000000000529043886f21d9bc1ae0fedb751e34265a246e470000000000000000000000000000000000000000000000000000000000000001",
"nonce":4,
"type":0,
"block_number":47,
"txindex":0,
"gas_used":84879,
"logs":[
   
],
"contract_address":"None",
"coverage_hash":"6f534eb3163a971ca53c44c83f4e1884c2b0baa5"
}
```
I will do the proper changes and add tests back.
cc @calina-c

trentmc:
> One way to broadcast transactions without blocking is to set required_confs = 0. 

If it doesn't wait for the tx to complete, won't basically any test that uses ganache fail? Because the tx won't have gone through, it means that state wouldn't have updated. 

E.g. if it sends off a tx to deploy an ERC721 contract, and it doesn't wait to go through, then it tests for existence of the ERC721, and it fails.

(And if this _doesn't_ happen, why not?)

calina-c:
@mariacarmina [and CC @trentmc] required_confs = 0 should only be set as a transaction parameters to functions that we are *expecting* to fail, due to pytest not catching these errors.

also, Maria, as I requested before, please separate just the solution into one branch. This is really cluttered and it contains many subsequent attempts we made. E.g. do we still need python 3.9? what about the gas strategy changes? Those were made for polygon on the (outdated/wrong) assumption that those tests hanged, but we discovered a different root problem in the meantime.

I also see all the exceptions are tested together, in one single function. Please do not do that. Keep everything in its proper (previous) place and create a helper function that abstracts away the checking part. But what is confusing to me is it seems now pytest *does* catch these? So the only difference would be passing in "required_confs": 0 in the tx dicts for expected failures.

mariacarmina:
> @mariacarmina [and CC @trentmc] required_confs = 0 should only be set as a transaction parameters to functions that we are _expecting_ to fail, due to pytest not catching these errors.
> 
> also, Maria, as I requested before, please separate just the solution into one branch. This is really cluttered and it contains many subsequent attempts we made. E.g. do we still need python 3.9? what about the gas strategy changes? Those were made for polygon on the (outdated/wrong) assumption that those tests hanged, but we discovered a different root problem in the meantime.
> 
> I also see all the exceptions are tested together, in one single function. Please do not do that. Keep everything in its proper (previous) place and create a helper function that abstracts away the checking part. But what is confusing to me is it seems now pytest _does_ catch these? So the only difference would be passing in "required_confs": 0 in the tx dicts for expected failures.

I will create shortly a commit with all the changes needed and cherry-pick into another branch.

mariacarmina:
> > One way to broadcast transactions without blocking is to set required_confs = 0.
> 
> If it doesn't wait for the tx to complete, won't basically any test that uses ganache fail? Because the tx won't have gone through, it means that state wouldn't have updated.
> 
> E.g. if it sends off a tx to deploy an ERC721 contract, and it doesn't wait to go through, then it tests for existence of the ERC721, and it fails.
> 
> (And if this _doesn't_ happen, why not?)

As @calina-c said, `required_confs` are only used for transactions that we expect to fail in order to retrieve the revert messages.
Apparently neither `pytest.raises()`, nor `brownie.reverts()` can catch those reverts without setting the `required_confs` to `0`. I wait for the tx to fetch properly with `tx.wait`, otherwise I cannot retrieve the tx parameters.

calina-c:
@mariacarmina do you want to recycle/keep anything regarding polygon from here? If so, I suggest opening a new PR which is more focused on those issues. However, I do not see this happening too soon, since the entire test suite need to be shortened, so Polygon tests might need to be adjusted at a later time. Please close this PR when convenient to you.

mariacarmina:
> @mariacarmina do you want to recycle/keep anything regarding polygon from here? If so, I suggest opening a new PR which is more focused on those issues. However, I do not see this happening too soon, since the entire test suite need to be shortened, so Polygon tests might need to be adjusted at a later time. Please close this PR when convenient to you.

I will create a separate PR since the issues have mixed up here, which has a better fee management on polygon based on this PR's approach.

PR #1276, opened 2023-01-13
Unblock CI run

mariacarmina:
Changes proposed in this PR:

- added fix from this [commit](https://github.com/oceanprotocol/ocean.py/pull/1272/commits/04cf6b43ab79ca5c9e0d2739cfdda4ec35270e68), #1272

### DIFF
--- a/ocean_lib/models/data_nft.py
+++ b/ocean_lib/models/data_nft.py
@@ -6,9 +6,10 @@ import json
 import warnings
 from base64 import b64encode
 from enum import IntEnum, IntFlag
-from typing import Optional
+from typing import Optional, Union
 
 from brownie import network
+from brownie.network.transaction import TransactionReceipt
 from enforce_typing import enforce_types
 
 from ocean_lib.models.datatoken import Datatoken
@@ -46,7 +47,9 @@ class Flags(IntFlag):
 class DataNFT(ContractBase):
     CONTRACT_NAME = "ERC721Template"
 
-    def create_datatoken(self, datatoken_args, tx_dict) -> Datatoken:
+    def create_datatoken(
+        self, datatoken_args, tx_dict
+    ) -> Union[Datatoken, TransactionReceipt]:
         return datatoken_args.create_datatoken(self, tx_dict)
 
     def calculate_did(self):
@@ -98,7 +101,9 @@ class DataNFTArguments:
             json.dumps(data, separators=(",", ":")).encode("utf-8")
         )
 
-    def deploy_contract(self, config_dict, tx_dict) -> DataNFT:
+    def deploy_contract(
+        self, config_dict, tx_dict
+    ) -> Union[DataNFT, TransactionReceipt]:
         from ocean_lib.models.data_nft_factory import (  # isort:skip
             DataNFTFactoryContract,
         )
@@ -119,6 +124,8 @@ class DataNFTArguments:
             self.owner or wallet_address,
             tx_dict,
         )
+        if "required_confs" in tx_dict.keys():
+            return receipt
 
         with warnings.catch_warnings():
             warnings.filterwarnings(
--- a/ocean_lib/models/datatoken.py
+++ b/ocean_lib/models/datatoken.py
@@ -106,7 +106,7 @@ class DatatokenArguments:
                 address=wallet_address, token=OCEAN_address
             )
 
-        data_nft.contract.createERC20(
+        tx = data_nft.contract.createERC20(
             self.template_index,
             [self.name, self.symbol],
             [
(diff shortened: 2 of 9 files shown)


ocean_lib/models/test/test_data_nft_factory.py
```diff
@@ -330,10 +331,18 @@ def test_nonexistent_template_index(data_nft_factory, publisher_wallet):
     )
     assert non_existent_nft_template >= 0, "Non existent NFT template not found."
 
-    with pytest.raises(Exception, match="Template index doesnt exist"):
-        data_nft_factory.create(
+    with pytest.raises(ValueError, match="Template index doesnt exist"):
```
calina-c, line 334: why do we still need interrogation? You needed to interrogate the chain and recheck the transaction status *only on the assumption* that pytest.raises did not catch these. Only then we needed to confirm

mariacarmina: Because it doesn't use a contract view function such as `getTokenTemplate`, it uses `deployERC721Contract` which is external/payable.

mariacarmina:
I think it's a good idea to comment this in a README or to put in the brownie config just for ganache transactions to have the confirmations set to 0 in case any user will mistake their contract calls and will encounter ganache freezing from barge. What do you think @calina-c @trentmc?

ocean_lib/models/test/test_data_nft_factory.py
calina-c: I don't understand. So pytest.raises works, it sees the error (now, with required confs set to 0). That is enough for us to consider the tx failed. No need to do the sleep and interrogation.

calina-c: and same everywhere... No need to confirm the transaction failed if it raises the error. The problem was it was hanging instead of failing. Now that it properly fails (as indicated by pytest.raises), there is no need to confirm that it failed on-chain.

mariacarmina: So, I create a wrong tx and in order to extract some values from it, I need to wait to be completely settled, otherwise I get weird errors from brownie such as `TypeError: int() can't convert non-string with explicit base`. 
`pytest.raises` works now because of that interrogation of blockchain `web3.eth.call()` from where I could retrieve the error message.
Between consecutive exception handling blocks, I need to put a sleep, otherwise the test fails with this error `TypeError: int() can't convert non-string with explicit base`.
As far as I saw from their code, they use a thread to make those transactions and maybe this can be a synchronization problem.

mariacarmina: That error ocurrs at this line `step["pc"] = int(step["pc"], 16)` and basically `step["pc"]` is not fetched properly from the transaction.

ocean_lib/models/data_nft.py
```diff
@@ -98,7 +101,9 @@ def get_default_token_uri(self):
             json.dumps(data, separators=(",", ":")).encode("utf-8")
         )
 
-    def deploy_contract(self, config_dict, tx_dict) -> DataNFT:
+    def deploy_contract(
+        self, config_dict, tx_dict
+    ) -> Union[DataNFT, TransactionReceipt]:
```
calina-c, line 106: I don't think you should modify these. I think we should throw an exception if the transaction fails

ocean_lib/models/data_nft.py
```diff
@@ -119,6 +124,8 @@ def deploy_contract(self, config_dict, tx_dict) -> DataNFT:
             self.owner or wallet_address,
             tx_dict,
         )
+        if "required_confs" in tx_dict.keys():
```
calina-c, line 127: please remove this. This is a test-related dependency an shouldn't have a different return type. It's bad coupling.

ocean_lib/models/datatoken.py
```diff
@@ -120,6 +120,8 @@ def create_datatoken(self, data_nft, tx_dict, with_services=False):
             tx_dict,
         )
 
+        if "required_confs" in tx_dict.keys():
```
calina-c, line 123: same here

ocean_lib/models/test/helpers.py
```diff
@@ -0,0 +1,18 @@
+from brownie.network import web3
+
+
+def interrogate_blockchain_for_reverts(
```
calina-c, line 4: no need for this anymore, I think

mariacarmina: I will remove `pytest.raises` block and  keep this function to retrieve the error message.

mariacarmina:
Code enhancements for this PR are on #1277. Closing this since #1277 has the same fix as it is exposed here.

PR #1277, opened 2023-01-18
Unblock ci 2

calina-c:
Refinements on #1276. Supersedes it.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Test with pytest
         run: |
           mkcodes --github --output tests/generated-readmes/test_{name}.{ext} READMEs
-          coverage run --source ocean_lib -m py.test ocean_lib/models
+          coverage run --source ocean_lib -m py.test
           coverage report
           coverage xml
         env:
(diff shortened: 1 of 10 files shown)


mariacarmina:
As for intermediary fix for mining the transaction, I had the same approach in #1276 [this commit](https://github.com/oceanprotocol/ocean.py/pull/1276/commits/09c42808b068471b2e4d901bc21e7a4f2d6bfc84), but you said that this is not the proper solution for fixing `TypeError("int() can't convert non-string with explicit base")`. What would be a stable approach on this? Maybe we can put it in an issue ticket.

mariacarmina (approved):

calina-c:
> As for intermediary fix for mining the transaction, I had the same approach in #1276 [this commit](https://github.com/oceanprotocol/ocean.py/pull/1276/commits/09c42808b068471b2e4d901bc21e7a4f2d6bfc84), but you said that this is not the proper solution for fixing `TypeError("int() can't convert non-string with explicit base")`. What would be a stable approach on this? Maybe we can put it in an issue ticket. I will approve this for now, because it took us a bit more than expected.

Yes, that was while we were still investigating the reason for the prolonged test runs. We tried a few solutions for the freezing, including chain mining and sending dummy transactions, and that uncovered that the "long duration" of the test suite and the "total freezing" issues are separate. I also talked to Alex about the prolonged duration of the test suite and the fact that we had to increase Aquarius waiting time both here and in the Provider repo. We will work on redesigning the flows to improve the test suite duration, but I consider the freezing part is done. Waiting for those failed transactions was simply the only solution.

PR #1283, opened 2023-01-20
Added gas fee strategy for remote tests.

### LINKED ISSUE #1213: Txs to polygon hang. Why: gas price it tried << gas price needed

mariacarmina:
Fixes #1213  .

Changes proposed in this PR:

- querying gas station from polygon
- updated tx dicts
- deleted util function for setting the priority fee

### DIFF
--- a/tests/integration/remote/util.py
+++ b/tests/integration/remote/util.py
@@ -8,8 +8,10 @@ import string
 import time
 import warnings
 
+import requests
 from brownie.exceptions import ContractNotFound, TransactionError, VirtualMachineError
 from brownie.network import accounts, chain, priority_fee
+from brownie.network.web3 import Web3
 from enforce_typing import enforce_types
 
 from ocean_lib.models.data_nft import DataNFTArguments
@@ -61,12 +63,6 @@ def get_wallets():
 
 
 @enforce_types
-def set_aggressive_gas_fees():
-    # Polygon & Mumbai uses EIP-1559. So, dynamically determine priority fee
-    priority_fee(chain.priority_fee)
-
-
-@enforce_types
 def do_nonocean_tx_and_handle_gotchas(ocean, alice_wallet, bob_wallet):
     """Call wallet.transfer(), but handle several gotchas for this test use case:
     - if the test has to repeat, there are nonce errors. Avoid via unique
@@ -80,7 +76,13 @@ def do_nonocean_tx_and_handle_gotchas(ocean, alice_wallet, bob_wallet):
 
     print("Do a send-Ether tx...")
     try:
-        alice_wallet.transfer(bob_wallet.address, f"{amt_send:.15f} ether")
+        # Polygon & Mumbai uses EIP-1559. So, dynamically determine priority fee
+        gas_fees = requests.get("https://gasstation-mainnet.matic.network/v2").json()
+        alice_wallet.transfer(
+            bob_wallet.address,
+            f"{amt_send:.15f} ether",
+            priority_fee=Web3.toWei(gas_fees["fast"]["maxPriorityFee"], "gwei"),
+        )
         bob_eth_after = accounts.at(bob_wallet.address).balance()
     except ERRORS_TO_CATCH as e:
         if error_is_skippable(str(e)):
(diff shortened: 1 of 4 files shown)


tests/integration/remote/util.py
```diff
@@ -105,8 +107,15 @@ def do_ocean_tx_and_handle_gotchas(ocean, alice_wallet):
 
     print("Call create() from data NFT, and wait for it to complete...")
     try:
+        # Polygon & Mumbai uses EIP-1559. So, dynamically determine priority fee
+        gas_fees = requests.get("https://gasstation-mainnet.matic.network/v2").json()
```
calina-c, line 111: there is some duplication with the non_ocean_tx. Could we make a function that does the request and returns the priority fee and the max fee?

Could we also add some fallbacks, i.e. if the response from the API is not 200 or any key doesn't exist, we should handle that (maybe fail the tests gracefully?)

calina-c (changes requested):
Good work, but it needs a bit more DRY

tests/integration/remote/util.py
mariacarmina: Ok, I will also add a fallback for that response.

tests/integration/remote/util.py
```diff
@@ -41,6 +43,23 @@ def remote_config_polygon(tmp_path):
...
+    gas_resp = requests.get("https://gasstation-mainnet.matic.network/v2")
+
+    if not gas_resp or gas_resp.status_code != 200:
+        print(
+            f"Invalid response from Polygon gas station. Retry with brownie values..."
+        )
+
+        return chain.priority_fee, chain.base_fee + 2 * chain.priority_fee
+
+    return Web3.toWei(gas_resp.json()["fast"]["maxPriorityFee"], "gwei"), Web3.toWei(
```
calina-c, line 58: nice work. If I were to nitpick, I would recommend assigning gas_resp.json() e.g. `gas_resp = gas_resp.json()` before, to make this easier to read, but only if you want to. Also, there's an empty line in nonocean_tx after the try which I would remove, but that is also minor. I'm approving, feel free to merge now or with these tiny changes.

calina-c (approved):

PR #1286, opened 2023-01-23
Retry mechanism for reverted transactions.

mariacarmina:
Changes proposed in this PR:

- fixed `TypeError` thrown by brownie
- implemented retry mechanism

### DIFF
--- a/tests/resources/helper_functions.py
+++ b/tests/resources/helper_functions.py
@@ -384,10 +384,26 @@ def get_mock_provider_fees(mock_type, wallet, valid_until=0):
     }
 
 
-@contextlib.contextmanager
-def delay_transaction():
-    yield
-    time.sleep(2)
+def retry_failed_transaction(func_name, contract, *args, **kwargs):
+    expires_in = 1  # in seconds
+    while expires_in <= 10:
+        try:
+            func = getattr(contract, func_name)
+            kwargs["from"] = kwargs.pop("wallet")
+            tx = func(*args, dict(**kwargs))
+            tx.wait(1)
+
+            return tx
+        except TypeError as err:
+            assert (
+                "int() can't convert non-string with explicit base"
+                in err.args[0]["message"]
+            )
+            time.sleep(expires_in)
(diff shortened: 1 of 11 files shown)


tests/resources/helper_functions.py
```diff
@@ -384,10 +385,32 @@ def get_mock_provider_fees(mock_type, wallet, valid_until=0):
...
+            else:
+                tx = func(*args, dict(**kwargs))
+            tx.wait(1)
+
+            return tx
+        except TypeError as err:
+            print(err)
+            assert "int() can't convert non-string with explicit base" in err.args[0]
+            time.sleep(expires_in)
+            expires_in += 5
```
idiom-bytes, line 409: The following steps => all of the steps of `expires_in`:
1. initialized = 1
2. **check_signature() x 1** 
3. exception
4. time.sleep(expires_in) (1) // wait 1 second
5. expires_in += 5 (6)
6. expires_in (6) <= 10 (true, continue)
7. **check_signature() x 1**
8. exception
9. time.sleep(expires_in) (6) // wait 6 seconds
10. expires_in += 5 (11)
11. expires_in (11) <= 10 (false, break) // conditional loop break

By my rough estimate, this will run twice at most **(check_signature() x 2)**.
The timing of the sleep cycles are also a bit fibonnaci-like. Is that the intention?

**Maybe the retry != sleep length? I.E.**
L392 expires_in <= 3
L408 time.sleep(5)
L409 expires_in += 1

idiom-bytes (commented):
Retry loop innards. :snake: 
Had fun reviewing this.

tests/resources/helper_functions.py
mariacarmina: The intention was to increase by 5 the waiting time. Thank you for review.
As I mentioned in the DM, this branch was used to discover the fix, but for an enhanced option you can check also #1288.

mariacarmina:
According to the enhancements brought in #1288 this PR will be closed.

Issue #1289, opened 2023-01-26
Retry mechanism enhancements


mariacarmina:
This issue was created as a follow-up on #1288 where we discussed how to improve the code quality & readability.
The following list contains problems that we noticed such as:

- [ ] assert `TypeError` message

- [ ] use `tx.wait(1)` to avoid `invalid operand` error

- [ ] #1291
(*first approach*: encapsulating the arguments in `args` as well)

MBadea17:
Closed as obsolete.

PR #1290, opened 2023-01-26
Fix randomly failing datatoken tests.

mariacarmina:
Fixes the last checkbox from #1289  .

Changes proposed in this PR:

- added `confirm_failed` function in `pytest.raises`
- fixed `args` assignment.

### DIFF
--- a/ocean_lib/models/test/test_datatoken.py
+++ b/ocean_lib/models/test/test_datatoken.py
@@ -97,15 +97,14 @@ def test_main(
     assert not permissions[DatatokenRoles.PAYMENT_MANAGER]
 
     # not erc20 deployer role
-    args = (
-        DatatokenArguments(
-            name="DT1",
-            symbol="DT1Symbol",
-        ),
-        {"from": another_consumer_wallet, "required_confs": 0},
-    )
     with pytest.raises(Exception, match="new data token has no address"):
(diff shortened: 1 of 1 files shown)


ocean_lib/models/test/test_datatoken.py
```diff
@@ -97,14 +97,15 @@ def test_main(
...
+    )
     with pytest.raises(Exception, match="new data token has no address"):
-        data_nft.create_datatoken(
-            DatatokenArguments(
-                name="DT1",
-                symbol="DT1Symbol",
-            ),
-            {"from": another_consumer_wallet, "required_confs": 0},
-        )
+        confirm_failed(data_nft, "create_datatoken", args, "")
```
calina-c, line 108: I also discovered the problem where I didn't remove the contract call in args, below. Please keep all the other changes except this one and let's discuss this particular fix later.

calina-c (changes requested):

ocean_lib/models/test/test_datatoken.py
mariacarmina: why that particular fix could not be fixed here, at least temporary because it will fail further? and why is not a good solution?

calina-c (approved):

Issue #1166, opened 2022-12-07
For action combos, minimize # txs by leveraging helper contract methods


trentmc:
### Summary

We have some potentially low-hanging fruit to minimize # txs when doing (a) create_asset + post for free/priced, (b) download file, (c) maybe more. Via helper smart contract methods.

This issue is about doing that. But note that we have other things to do first. (See Priorities section.)

### Motivation

_Alex:_  (in slack)
I would have expected this [create exchange] in ocean_assets.create, since we have a lot of helpers in the contracts factory to reduce no of tx

_Trent:_

I agree to the value of that. We need to walk before we run.

The first priority has been to ensure that atomic actions -- atomic from perspective of user -- are clean.

And some of these naturally bundle. Examples:
- download_file() - does (1) do the order (2) do the download
- create() - could do (1) create data nft (2) create datatoken (3) create ddo (4) deploy a dispenser or fixed-price. Right now it does (1)(2)(3) but we could extend to (4) too.

With those natural bundles, we can then optimize the # txs by leveraging the special smart contracts calls. (And for cases we don't have that, use Gnosis Transaction Builder.)

But it's one step at a time. 

### TODOs

- [ ] Do bundle: `ocean_assets::create()` lets the user optionally set a pricing scheme (free or fixed-price) with associated parameters. Minimize # txs doing this
- [ ] Do bundle: in `download_file()`, minimize # txs when doing this. Must work for all combos of {free, priced} x {template 1, template 2}
- [ ] Identify other existing helper contract methods to identify opportunities. Add to this issue's TODOs to leverage them
- [ ] Look at flows in ocean.py and predict-eth READMEs, to find any other natural bundles. Identify them here, and add to this issue's TODOs to leverage them. If we don't have existing helper contract method for a highly-used bundle, add "try out Gnosis Transaction Builder" to the TODOs
- [ ] By the time we're here, we've added several TODOs. Prioritize them, and execute on them. For larger issues (eg Gnosis), perhaps create a new issue, then treat _this_ issue like an epic.
 
### On Priority

Priority of this issue: only do _after_ we've fixed bugs (there are currently >5 that really matter), and done other UX cleanups.

Recently-completed UX cleanups:
- [#977](https://github.com/oceanprotocol/ocean.py/issues/977) use brownie
- [#910](https://github.com/oceanprotocol/ocean.py/issues/910) Streamline dispenser.py & dispenser-flow README 
[...]

PR #1292, opened 2023-01-27
Use combo tx in create 1dt.

mariacarmina:
Helps #1166.

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -13,7 +13,6 @@ from typing import List, Optional, Tuple, Type, Union
 
 from brownie import network
 from enforce_typing import enforce_types
-from web3 import Web3
 
 from ocean_lib.agreements.consumable import AssetNotConsumable, ConsumableCodes
 from ocean_lib.agreements.service_types import ServiceTypes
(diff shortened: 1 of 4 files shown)


ocean_lib/ocean/ocean_assets.py
```diff
@@ -248,16 +258,73 @@ def _default_metadata(self, name: str, tx_dict: dict, type="dataset") -> dict:
         return metadata
 
     @enforce_types
-    def _create_1dt(self, metadata, files, tx_dict, wait_for_aqua):
-        """Call create(), focusing on just one datatoken"""
+    def _create_1dt(
```
mariacarmina, line 261: Is a possible way to find another name for this bundle function? Because `_create_1dt` focuses on just one datatoken, but here we included the data NFT, the datatoken & the data asset

calina-c: this is still an internal function. In the next iteration, when I add the possibility to add pricing schema, I will rename it something smart, like `create_with_bundle` or something similar. This is Trent's naming for now and I chose to modify just the contents until I change the "effect" as well. Do you think that could work for now?

PR #1296, opened 2023-01-30
Reverted error handling from brownie to pytest raises approach

mariacarmina:
Context:
Brownie version was updated to 1.19.3 where web3.py solved their race condition problem [here](https://github.com/eth-brownie/brownie/commit/f5a31ff02b068795ff093adf05a4661917c65ea9). We can use `pytest.raises` to catch the expected errors now.

Changes proposed in this PR:

- removed helper functions
- replaced the logic before for error handling with `pytest.raises`


mariacarmina:
Should we remove the `try-except` blocks from `data_nft` and `datatoken` classes as well?

calina-c (approved):

calina-c:
> Should we remove the `try-except` blocks from `data_nft` and `datatoken` classes as well?

yes

PR #1298, opened 2023-01-31
Fix error message in data nft creation.

calina-c:


calina-c:
Discussed with Maria: let's table this until we see this error again. It might have been a fluke and it's a lot damaging to the Mumbai tests right now.

Issue #1174, opened 2022-12-09
Exception: Address not found for development. Please check your address file


trentmc:
### Background

It seems that the recent changes broke the ability for ocean to connect to ganache. (Commit https://github.com/oceanprotocol/ocean.py/commit/236b88567304955a46dbd8d38010d9078958982e, from PR https://github.com/oceanprotocol/ocean.py/pull/1171, for issue https://github.com/oceanprotocol/ocean.py/issues/1164).

The error is easy to reproduce. Basically, try to do _anything_ with ocean. Here's one way..

- In [developers.md](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/developers.md), go through steps 1, 2, 3 on installation etc -- all good
- Then, in console: `cp network-config.sample ~/.brownie/network-config.yaml` (This needs to be in developers.md. But we're about to change that anyway, so let's not worry about it yet.)
- Did [simple-flow](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/data-nfts-and-datatokens-flow.md) step 1 "Setup in Python". This is connecting to network, setting wallets. All good.
- did simple-flow step 2 "Publish Data NFT & Datatoken". In Python console: `data_nft = ocean.create_data_nft('NFT1', 'NFT1', alice_wallet)`

This will fail, with: 
```text
  File "/home/trentmc/code/ocean.py/ocean_lib/web3_internal/contract_utils.py", line 60, in get_contracts_addresses
    raise Exception(
Exception: Address not found for development. Please check your address file
```

Full traceback below

### More digging

In python:
```python
from ocean_lib.web3_internal.contract_utils import \
    get_addresses_with_fallback, get_contracts_addresses
addresses = get_addresses_with_fallback(config)

addresses.keys()
```

Which returns: `dict_keys(['rinkeby', 'mumbai', 'ropsten', 'moonbase', 'polygon', 'bsc', 'energyweb', 'moonriver', 'mainnet', 'goerli', 'polygonedge', 'gaiaxtestnet'])`

Note that it does _not_ include a `development` key. 

Where does it get this dict from? Let's dig...

The first line of `get_addresses_with_fallback()` does the following. Let's do it too: 

```python
address_file = config.get("ADDRESS_FILE")
```

Then let's see the value:
```python
print(address_file)
```

Which returns: `PosixPath('/home/trentmc/code/ocean.py/venv/lib/python3.10/site-packages/addresses/address.json')`

[...]

PR #1301, opened 2023-02-01
Fixes to the test runs

calina-c:
Helps with #1174. Includes:

- using psf/action-black, which is the official black action and allows version grounding. Previously, we were using a specific version in pre-commits and an auto-resolvable version in the Github Action check. If those are different, black fails.
- Display the address file in barge to help debugging for the "addresses not found" issue.
- Dependency grounding of eth-brownie, a fix for Python 3.11 dependencies and details in the installation readme. This was because pip was downgrading eth-brownie to a very old version just to satisfy its vyper dependency. This meant brownie was actually incompatible with the current code and broke imports. I also created https://github.com/oceanprotocol/ocean.py/issues/1302 as a follow-up to prevent similar issues in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,12 +35,11 @@ install_requirements = [
     "eciespy",
     "eth-brownie==1.19.3",
     "yarl==1.8.1",
-    "bitarray>=2.6.0,<3",
+    "bitarray>=2.6.0,<3"
     # web3 requires eth-abi, requests, and more,
     # so those will be installed too.
     # See https://github.com/ethereum/web3.py/blob/master/setup.py
 ]
-
 # Required to run setup.py:
 setup_requirements = ["pytest-runner"]
 
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ install_requirements = [
     "eciespy",
     "eth-brownie==1.19.3",
     "yarl==1.8.1",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -33,14 +33,13 @@
     "enforce-typing==1.0.0.post1",
     "json-sempai==0.4.0",
     "eciespy",
-    "eth-brownie",
+    "eth-brownie==1.19.3",
```
mariacarmina, line 36: Please do not modify this version, this will raise dependency conflicts.

calina-c: This is not what is causing dependency conflicts. Look at all previous runs. Those with python 3.11 are downgrading brownie until 1.8 and now it's 1.19.3, the only release where the race conditions are fixed. I can explain in detail in a call. I've tried changing pip, dependency magic and everything but now the main release is broken for all python versions. At least this one is only broken for python 3.11. Thats because vyper 3.7.0, a dependency of brownie, only recently introduceda python 3.11 as a possibility and they haven't released a 3.8, making brownie incompatible with python 3.11

calina-c: We also have to clean and ground all dependencies for dependabot to actually do its job.

mariacarmina (approved):

PR #1303, opened 2023-02-01
Simplify usage of Arguments classes in direct create calls

calina-c:
Follow-up on #1166, where we introduced new Argument classes and we decided to simplify the interface a bit.


mariacarmina (approved):
Nice work here @calina-c!

PR #1304, opened 2023-02-01
Update get_datatoken() to return also DatatokenEnterprise.

### LINKED ISSUE #1311: Update get_datatoken() to return also DatatokenEnterprise.
`get_datatoken()` function does not return explicitly `DatatokenEnterprise` where needed.
The instance is plain `Datatoken` and when I want to call `dispenser_and_order`, it should point to the `buyFromDispenserAndOrder`, that is present in `datatoken_enterprise`, but instead it access the method `dispense_and_order` from `datatoken.py`, which is not the right flow.

Test snippet provided in #1300.

mariacarmina:
Fixes #1311  .

Changes proposed in this PR:

- updated get_datatoken()

### DIFF
--- a/ocean_lib/ocean/ocean.py
+++ b/ocean_lib/ocean/ocean.py
@@ -158,9 +158,11 @@ class Ocean:
         :param token_address: Token contract address, str
         :return: `Datatoken` or `DatatokenEnterprise` instance
         """
+        datatoken = Datatoken(self.config, token_address)
+
         return (
-            Datatoken(self.config, token_address)
-            if Datatoken(self.config, token_address).contract.getId() == 1
+            datatoken
+            if datatoken.getId() == 1
             else DatatokenEnterprise(self.config, token_address)
         )
 
--- a/ocean_lib/ocean/ocean.py
+++ b/ocean_lib/ocean/ocean.py
@@ -18,6 +18,7 @@ from ocean_lib.models.compute_input import ComputeInput
 from ocean_lib.models.data_nft import DataNFT
 from ocean_lib.models.data_nft_factory import DataNFTFactoryContract
 from ocean_lib.models.datatoken import Datatoken
+from ocean_lib.models.datatoken_enterprise import DatatokenEnterprise
 from ocean_lib.models.df.df_rewards import DFRewards
 from ocean_lib.models.df.df_strategy_v1 import DFStrategyV1
 from ocean_lib.models.dispenser import Dispenser
@@ -153,9 +154,13 @@ class Ocean:
     def get_datatoken(self, token_address: str) -> Datatoken:
         """
         :param token_address: Token contract address, str
-        :return: `Datatoken` instance
+        :return: `Datatoken` or `DatatokenEnterprise` instance
         """
-        return Datatoken(self.config, token_address)
+        return (
+            Datatoken(self.config, token_address)
+            if Datatoken(self.config, token_address).contract.getId() == 1
+            else DatatokenEnterprise(self.config, token_address)
+        )
 
     # ======================================================================
     # orders
--- a/ocean_lib/ocean/ocean.py
+++ b/ocean_lib/ocean/ocean.py
@@ -151,7 +151,9 @@ class Ocean:
         return DataNFT(self.config, token_address)
 
     @enforce_types
-    def get_datatoken(self, token_address: str) -> Datatoken:
+    def get_datatoken(
+        self, token_address: str
+    ) -> Union[Datatoken, DatatokenEnterprise]:
         """
         :param token_address: Token contract address, str
         :return: `Datatoken` or `DatatokenEnterprise` instance


trentmc:
Why is this needed?

DatatokenEnterprise inherits from Datatoken after all. So any type declaration should be fine with just Datatoken being checked. That's part of the point of OO.

cc @calina-c

mariacarmina:
I discussed with @calina-c today and she said that we can add the same approach as it is in `datatoken.py`, in [create_erc20](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/models/datatoken.py#L130). I created just for enhancement & consistency with `datatoken.py`, but I can close this, because I understood your point.
Most of the times I connect the issues to PRs, but we wanted to do a release today of ocean-lib and I was in a hurry to put it in review, sorry.
As you mentioned previously, I will close this and thank you @trentmc  for your input.

calina-c:
@trentmc due to the polymorphism we employed, if an object is instantiated as Datatoken and not DatatokenEnterprise, the "mock" functions for buying & ordering or dispensing and ordering will be incorrectly applied through the logic of Datatoken.   We have the same names and arguments on those methods and they "cover" the correct behaviour when used from an incorrectly instantiated object. I think this is in fact necessary.

calina-c:
@mariacarmina

ocean_lib/ocean/ocean.py
```diff
@@ -150,12 +151,18 @@ def get_nft_token(self, token_address: str) -> DataNFT:
...
+    ) -> Union[Datatoken, DatatokenEnterprise]:
         """
         :param token_address: Token contract address, str
-        :return: `Datatoken` instance
+        :return: `Datatoken` or `DatatokenEnterprise` instance
         """
-        return Datatoken(self.config, token_address)
+        return (
+            Datatoken(self.config, token_address)
+            if Datatoken(self.config, token_address).contract.getId() == 1
```
calina-c, line 163: please use templateId if possible

ocean_lib/ocean/ocean.py
```diff
@@ -150,12 +151,18 @@ def get_nft_token(self, token_address: str) -> DataNFT:
...
+        self, token_address: str
+    ) -> Union[Datatoken, DatatokenEnterprise]:
         """
         :param token_address: Token contract address, str
-        :return: `Datatoken` instance
+        :return: `Datatoken` or `DatatokenEnterprise` instance
         """
-        return Datatoken(self.config, token_address)
+        return (
+            Datatoken(self.config, token_address)
```
calina-c, line 162: I would suggest
```
datatoken = Datatoken(self.config, token_address)

return datatoken if datatoken.getId() == 1 else DatatokenEnterprise(self.config, token_address)

**Comment by mariacarmina** [comment]
According to what @calina-c mentioned above, I have taken some screenshots of a test that has a datatoken enterprise address and wants to be retrieved with `ocean.get_datatoken()` function. I want to add that this code snippet is related and solves a part from issue #1300, because I picked up the problem from a Discord thread.

I took a [free asset](https://market.oceanprotocol.com/asset/did:op:7b14ebca07410ba3226178255cea467df6c9ce49b3ca17e10f94fdfdbecc29ee) from mumbai, which as an enterprise datatoken attached with `0xA63d7C622547E197321Cbd7A0886a524c4Ba7b06` address.

I put a breakpoint to check the datatoken instance and contract:
Screenshot from 2023-02-03 14-20-55

The instance is plain `Datatoken` and when I want to call `dispenser_and_order`, it should point to the `buyFromDispenserAndOrder`, that is present in `datatoken_enterprise`, but instead it access the method `dispense_and_order` from `datatoken.py`, which is not the right flow.

Screenshot from 2023-02-03 14-22-26

**Code snippet**

```python
import os

from ocean_lib.web3_internal.utils import connect_to_network
connect_to_network("polygon-test") # mumbai is "polygon-test"s

config = get_config_dict("polygon-test")
ocean = Ocean(config)

OCEAN = ocean.OCEAN_token

ddo = ocean.assets.resolve("did:op:7b14ebca07410ba3226178255cea467df6c9ce49b3ca17e10f94fdfdbecc29ee")
datatoken = ocean.get_datatoken("0xA63d7C622547E197321Cbd7A0886a524c4Ba7b06")
print(datatoken)

from brownie import accounts

wallet = accounts.add(os.environ["REMOTE_TEST_PRIVATE_KEY1"])

provider_fees = ocean.retrieve_provider_fees(
    ddo,
    ddo.services[0],
    publisher_wallet=wallet
)

tx = datatoken.dispense_and_order(wallet.address, 0,
                              provider_fees,
                              {"from": wallet})

asset_dir = ocean.assets.download_asset(ddo, wallet, './', tx.txid)
```

ocean_lib/ocean/ocean.py
mariacarmina: sure, thank you!

calina-c (approved):

trentmc:
Thanks for the explanations. Go ahead and merge this.

This is not perfect however. There's a more core issue with the current class setup of Datatoken and DatatokenEnterprise. I just created the following issue to explain it, and a potential solution. Not urgent, but important. https://github.com/oceanprotocol/ocean.py/issues/1315

PR #1307, opened 2023-02-03
Fix invalid timestamps

Breta01:
Hi Bret here from @FELT-Labs,

We run into issues with timestamps when being located in different time zones. The issue is that using `datetime.utcnow()` generates UTC time, but it doesn't set `tzinfo` which is the timezone information. But `.timestamp()` function actually takes timezone into consideration, so if `datetime` object doesn't have timezone info set, it will subtract the timezone difference during timestamp generation (you can read the source code here: https://github.com/python/cpython/blob/3.11/Lib/datetime.py#L1905). Easiest solution for that is just to use `datetime.now()` and let timestamp() handle the timezones. We could also use `datetime.now(timezone.utc)` which set's the timezone info to UTC, but that's more wordy.

There is also mistake in nonce generation for compute jobs which should be in milliseconds

Changes proposed in this PR:

- Fix timestamp creating
- Nonce for compute job must be in milliseconds

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -152,7 +152,7 @@ datasets, algorithm = ocean.assets.pay_for_compute_service(
     consume_market_order_fee_address=bob.address,
     tx_dict={"from": bob},
     compute_environment=free_c2d_env["id"],
-    valid_until=int((datetime.utcnow() + timedelta(days=1)).timestamp()),
+    valid_until=int((datetime.now() + timedelta(days=1)).timestamp()),
     consumer_address=free_c2d_env["consumerAddress"],
 )
 assert datasets, "pay for dataset unsuccessful"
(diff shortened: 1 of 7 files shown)


AnaLoznianu:
Hello, thanks for your contribution 👍

mariacarmina:
Hello @Breta01, I will also create a test to reproduce the behaviour and keep you posted with a review.

calina-c:
Reported in #1410 and started work on all repos for safe measure.
Thank you for reporting this @Breta01

calina-c:
@mariacarmina If I'm not mistaken, you covered the miliseconds part and I covered the timezone formatting part previously. Do you think this issue can be closed?

mariacarmina:
> @mariacarmina If I'm not mistaken, you covered the miliseconds part and I covered the timezone formatting part previously. Do you think this issue can be closed?

This issue can be closed, the problem was the nonce computation in seconds instead of miliseconds which generated a different datetime object.

Issue #1300, opened 2023-02-01
Provide a clearer flow for people to create and consume custody-light data assets (ie DatatokenEnterprise)


mariacarmina:
I have seen some reports from Discord lately regarding how we can consume free assets through ocean.py. And my first thought was that we can improve certain comments that we have in `main-flow.md`.

For example in [this section](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/main-flow.md#3-bob-consumes-the-dataset), we can specify that: 
*"For a free asset, `order_tx_id` is retrieved from the `dispense_and_order().txid` call, not from `pay_for_access_service`. "*, or something similar.

On the same page, I saw that some users want to mint enterprise datatokens to `start_order` which is impossible, due to the fact that enterprise datatokens are burnt after the `startOrder` call and maybe we can add this too, if it's needed.

Discord thread where the problem was discussed: [here](https://discord.com/channels/612953348487905282/1067812647375994941)

**Temporary fix for the user**
```python

import os
from ocean_lib.models.datatoken_enterprise import DatatokenEnterprise
from ocean_lib.web3_internal.utils import connect_to_network
connect_to_network("polygon-test") # mumbai is "polygon-test"s

from ocean_lib.example_config import get_config_dict
from ocean_lib.ocean.ocean import Ocean
config = get_config_dict("polygon-test")
ocean = Ocean(config)

OCEAN = ocean.OCEAN_token

ddo = ocean.assets.resolve("did:op:7b14ebca07410ba3226178255cea467df6c9ce49b3ca17e10f94fdfdbecc29ee")
datatoken = DatatokenEnterprise(ocean.config_dict, "0xA63d7C622547E197321Cbd7A0886a524c4Ba7b06")
print(datatoken)

from brownie import accounts

wallet = accounts.add(os.environ["REMOTE_TEST_PRIVATE_KEY1"])

provider_fees = ocean.retrieve_provider_fees(
    ddo,
    ddo.services[0],
    publisher_wallet=wallet
)

tx = datatoken.dispense_and_order(wallet.address, 0,
                              provider_fees,
                              {"from": wallet})

asset_dir = ocean.assets.download_asset(ddo, wallet, './', tx.txid)
```
I have also let the community know about this [here](https://discord.com/channels/612953348487905282/720631837122363412/1070094271778136134).

I have also discussed with @AnaLoznianu about this and she agreed on specifying those information in the README, but I wanted to have your input on this as well @trentmc.

[...]

PR #1312, opened 2023-02-03
Bundle pay for service transactions

### LINKED ISSUE #1166: For action combos, minimize # txs by leveraging helper contract methods

calina-c:
Closes #1166. Adds support for detecting templates and bundling together the txs needed for `pay_for_service`.

### DIFF
--- a/ocean_lib/models/datatoken_base.py
+++ b/ocean_lib/models/datatoken_base.py
@@ -397,7 +397,7 @@ class DatatokenBase(ABC, ContractBase):
     def get_publish_market_order_fees(self):
         return TokenFeeInfo.from_tuple(self.contract.getPublishingMarketFee())
 
-    def buy_1dt_from_pricing_schema(self, *args, **kwargs):
+    def get_from_pricing_schema_and_order(self, *args, **kwargs):
         dispensers = self.dispenser_status().active
         exchanges = self.get_exchanges()
 
--- a/ocean_lib/models/datatoken_base.py
+++ b/ocean_lib/models/datatoken_base.py
@@ -77,7 +77,7 @@ class DatatokenArguments:
         cap: Optional[int] = None,
     ):
         if template_index == 2 and not cap:
-            raise Exception("Cap is needed for Datatoken Enterprise token deployment.")
+            raise Exception("Cap is needed for Datatoken Template 2 token deployment.")
 
         self.cap = cap if template_index == 2 else MAX_UINT256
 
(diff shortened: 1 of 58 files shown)


calina-c:
Blocker for #1300

ocean_lib/models/datatoken_base.py
```diff
@@ -397,6 +397,58 @@ def build_access_service(
     def get_publish_market_order_fees(self):
         return TokenFeeInfo.from_tuple(self.contract.getPublishingMarketFee())
 
+    def buy_1dt_from_pricing_schema(self, *args, **kwargs):
```
alexcos20, line 400: We should rename this to something like "get_from_pricing_schema_and_order", because, depending on conditions, at the end, we call one of:
 - buy_DT_and_order
 - dispense_and_order

alexcos20 (changes requested):

alexcos20 (approved):

PR #1314, opened 2023-02-06
Ground dependencies and remove extras

### LINKED ISSUE #1302: Ground and clean ocean.py dependencies
We should cleanup and ground (declare explicitly) dependencies of ocean-lib for several reasons:
- it's the only way dependabot will actually do its job
- we previously had some c2d examples that used pillow and matplotlib, but I think we're not using those anymore. We probably have a more cases where we declare dependencies but don't actually need/use them. Let's prune that dependency tree.
- having dependencies that are not explicit results in issues like those we have previously seen: libs that resolve to a really old version and we have no idea about that, we just assume it works if the lib is resolved. Unexpected issues with new releases where we don't know why we suddenly have failures etc.

### LINKED ISSUE #1305: ocean-lib PyYAML requirement prevents interoperability with other libs
**Describe the bug**
I am trying to build a docker image that uses ocean-lib and langchain. One error I noticed in the installation of my requirements.txt, is that ocean-lib uses PyYAML 5.4.1 whereas langchain uses newer versions, preventing the build of the docker image.

Is the requirement for the PyYAML version strict or is there some way to overcome this issue?

calina-c:
Closes #1302. Closes #1305 

Still pending:
- [x] update the classifier regarding release status from pre-alpha (to what?) -> see https://pypi.org/classifiers/
- [x] should we keep the original package author?
- [x] coloredlogs, used in testing settings, introduces both the dependency to coloredlogs and pyyaml to setup a yaml log file. I have never seen this file, are we using it? We should look into removing those dependencies if not essential, or at least downgrade them to dev/test dependencies if they belong there. (experimental branch exists for this)


AnaLoznianu:
Blocker for: #1303

trentmc:
> * [ ]  update the classifier regarding release status from pre-alpha (to what?) -> see https://pypi.org/classifiers/

I'm fine with either of:
- [Development Status :: 4 - Beta](https://pypi.org/search/?c=Development+Status+%3A%3A+4+-+Beta) 
- [Development Status :: 5 - Production/Stable](https://pypi.org/search/?c=Development+Status+%3A%3A+5+-+Production%2FStable) 

And I lean to (4) for now. 

We can get to (5) when installing ocean.py is less finicky, which includes:
- it's got so stable that the "install" readme can be just a single line `pip install ocean-lib` rather than half a page of gotchas & workarounds, AND
- ocean.py runs out-of-the-box on anaconda (this is more useful than you may realize: once we have this, we can get teed up to be part of the conda distribution itself(

trentmc:
> should we keep the original package author?

What's the motivation to change it?

FWIW, I think "leucothia" is hard to pronounce or understand why it's named that way. Better is "ocean-core-team" or something

PR #1320, opened 2023-02-08
Fix "Tx dropped" error on polygon test

### LINKED ISSUE #1319: CI fails randomly with `Tx dropped without known replacement` error in remote tests
Randomly, the CI waits for confirmation in the mempool on polygon and when the timeout expires and this error occurs: `Tx dropped without known replacement `. [Link to the failed run](https://github.com/oceanprotocol/ocean.py/actions/runs/4084391264/jobs/7041032603#step:8:159).

According to this [article](https://klaytn.foundation/a-comparison-of-blockchain-network-latencies/), on Polygon Mainnet we should wait upon 20 confirmations to consider the tx settled.

Mumbai requires 4 confirmations for a tx according [this](https://hyperledger.github.io/firefly/tutorials/chains/polygon_testnet.html).

My approach is to set the required confirmations either explicitly in the tx dict (which works smoothly and tested on datarella project), either in the `network-config.yaml`.

mariacarmina:
Fixes #1319  .

Changes proposed in this PR:

- added explicitly the `required_confs` to the tx.


tests/integration/remote/util.py
```diff
@@ -48,11 +48,22 @@ def get_gas_fees_for_remote() -> tuple:
...
+        return (
+            Web3.toWei(gas_resp.json()["fast"]["maxPriorityFee"], "gwei"),
+            Web3.toWei(gas_resp.json()["fast"]["maxFee"], "gwei"),
+            4,
+        )
+
+    return (
+        Web3.toWei(gas_resp.json()["fast"]["maxPriorityFee"], "gwei"),
+        Web3.toWei(gas_resp.json()["fast"]["maxFee"], "gwei"),
+        20,
```
calina-c, line 66: the only thing that is different in both cases is a number, right? Why not assign that number instead of copy-ing 4 lines of code?

mariacarmina: Sure, thank you.

tests/integration/remote/util.py
```diff
@@ -48,11 +48,25 @@ def get_gas_fees_for_remote() -> tuple:
 
     if not gas_resp or gas_resp.status_code != 200:
         print("Invalid response from Polygon gas station. Retry with brownie values...")
+        required_confs = 20
+        if chain.id == 80001:
```
calina-c, line 52: required_confs = 4 if chain.id == 80001 else 20

:)

mariacarmina: the chain ID needs to be changed then, because for polygon mainnet is 20 and it has a different chain ID.

So it would be:
required_confs = 4 if chain.id == 137 else 20

tests/integration/remote/util.py
```diff
@@ -119,23 +126,32 @@ def do_ocean_tx_and_handle_gotchas(ocean, alice_wallet):
...
+                symbol,
+                symbol,
+            )
+            data_nft_symbol = data_nft.symbol()
+            break
+        except ERRORS_TO_CATCH as e:
+            if error_is_skippable(str(e)):
+                warnings.warn(UserWarning(f"Warning: EVM reported error: {e}"))
+                return
+            if "Tx dropped" in str(e):
```
calina-c, line 148: is "Tx dropped" not part of the skippable errors?

mariacarmina: "Tx dropped" is not mentioned in `error_is_skippable`

calina-c (approved):

PR #1321, opened 2023-02-08
Add abstract DatatokenBase.

### LINKED ISSUE #1315: ocean.py handling of datatoken templates is error-prone and hard to extend
### Background / motivation

This issue was partly spurred by discussion in [PR 1304](https://github.com/oceanprotocol/ocean.py/pull/1304), but also something that's been brewing for a while in my head:)

Here are the problems:
- The current class setup of Datatoken and DatatokenEnterprise is messy. "Datatoken" class both defines the (a) baseline interface for all Ocean datatokens in Python, and (b) implements template 1 for it. Then "DatatokenEnterprise" class reuses some of it, but to understand you have to review both libraries.
- The labels are poor. "Datatoken" class labels both (a) interface of all datatokens, and (b) implementation for template 1. And "DatatokenEnterprise" labels template 2, and there's little correlation with "enterprise"; and having "enterprise" in its label is counter-productive. 

Here's another way to illustrate the problem: imagine if we wanted to add another template. It would be super-messy, because of all the weird relations above. (And, there's a very high chance we do want to add templates, probably in the next 6 months. So this is a real challenge.)

This issue is not urgent to solve, though it is important. And it will become urgent if we want to add another template.

### Towards a solution

There's a better way:
- have an _abstract_ class "Datatoken" or "AbstractDatatoken" that defines the interface
- have classes "Datatoken1" and "Datatoken2" for templates 1 and 2 respectively. Each of these inherits (i) the abstract class for interface definition, and (ii) appropriate mixins with the implementations as needed. 
- then implement the mixins too. There will probably be a few. It'll become obvious as you build it.

### TODOs

- Build "towards a solution" as a prototype, get feedback, iterate until good
- Update READMEs and documentation

### Appendix / Notes

I recognize that the smart contracts themselves are [named](https://github.com/oceanprotocol/contracts/tree/main/contracts/templates) similarly confusingly: ERC20Template.sol and "ERC20TemplateEnterprise.sol". That's harder to change given that these are deployed immutably. I do recommend that future templates get named "ERC20Template3.sol", "ERC20Template4.sol", etc.

calina-c:
Closes #1315


mariacarmina (approved):
LGTM

PR #1323, opened 2023-02-09
Issue 1239 optimise dt functions

### LINKED ISSUE #1239: Maybe: optimize DT.dispense_and_order() and buy_DT_and_order() more
### Background / motivation

(This first came up within [PR#1220](https://github.com/oceanprotocol/ocean.py/pull/1220) "Fee objects, to make fees less cumbersome", in [this comment](https://github.com/oceanprotocol/ocean.py/pull/1220#discussion_r1056342572). It's moved to _this_ new github issue, so that we can merge that PR. I'll copy the key stuff from PR#1220.)

----

From Trent:

### test_dispense_and_order_with_defaults()

We can optimize the defaults for this even better.

Status quo:

```python
    tx = DT.dispense_and_order(
        consumer=consumer_wallet.address,
        service_index=1,
        provider_fees=provider_fees,
        transaction_parameters={"from": publisher_wallet},
    )
```

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
order_tx_id = DT.dispense_and_order(provider_fees, {"from": publisher_wallet})
```

Where:
- it returns `order_tx_id`, not `tx`. Like calls to `pay_for_access_service()`.

With defaults set as:
```python
consumer = transaction_parameters["from"].address
service_index = 1
```

### test_buy_DT_and_order()

We should split test function this into two test functions: one with good defaults, and one with nondefaults. 

### [new] test_buy_DT_and_order_with_defaults()

Status quo: 
```python
    args = {
        "consumer": another_consumer_wallet.address,
        "service_index": 1,
        "provider_fees": provider_fees,
        "consume_market_fees": FeeTokenInfo(
            address=consume_fee_address,
            token=DAI.address,
        ),
        "exchange": exchange,
        "transaction_parameters": {"from": publisher_wallet},
    }

    if template_index == 2:
        args["max_base_token_amount"] = to_wei(2.5)
        args["consume_market_swap_fee_amount"] = to_wei(0.001)  # 1e15 => 0.1%
        args["consume_market_swap_fee_address"] = another_consumer_wallet.address

    tx = DT.buy_DT_and_order(**args)
```

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
order_tx_id = exchange.buy_DT_and_order(provider_fees, {"from": publisher_wallet})
```

Where:
- the call is to 'exchange` (`Exchange` class), not `DT` (`Datatoken` class). Just like `buy_DT()`. And `exchange` knows which datatoken to use, by definition
- it returns `order_tx_id`, not `tx`. Like calls to `pay_for_access_service()`.

With defaults set as:
```python
consumer = transaction_parameters["from"].address
service_index = 1
consume_market_fees = <zero>
```

### [new] test_buy_DT_and_order_with_non_defaults()

Status quo: (like shown above)

The "args" setup makes it harder to read. Let's remove that. And, it should also return `order_tx_id` like the defaults flow. So, it will look something like:

Optimize into:
```python
provider_fees  = get_mock_provider_fees(..)
consume_market_order_fees = FeeTokenInfo(address=consume_fee_address, token=DAI.address)
consume_market_swap_fees = FeeTokenInfo(to_wei(0.001), address=another_consumer_wallet.address, token=..)
max_base_token_amount = to_wei(2.5)
order_tx_id = exchange.buy_DT_and_order(
        consumer = another_consumer_wallet.address,
        service_index = 1,
        max_base_token_amount = max_base_token_amount,
        provider_fees = provider_fees,
        consume_market_order_fees = consume_market_order_fees,
        consume_market_swap_fees = consume_market_swap_fees,
        transaction_parameters = {"from": publisher_wallet},
)
```

Where:
- both template 1 and 2 should be aware of max_base_token_amount and consume_market_swap_fees (not just template 2 like the status quo)

-----

Response by Calina:

> optimising to: DT.dispense_and_order(provider_fees, {"from": publisher_wallet})
this test uses the consumer_wallet for the consumer address, while the sender of the transaction is publisher_wallet. I think we the consumer is NOT by default the sender of the transaction. Especially because we mimic DatatokenEnterprise here.

Same comment for the exchange test. Even more so, since we have another_consumer_wallet as the consumer.

Ref: moving this class to Exchange and figuring out there whether to do a direct call from DatatokenEnterprise template vs. combo with multiple txs from Datatoken: I can do that, but there is a code smell here. If the max_base_token_amount, consume_market_swap_fee_amount and consume_market_swap_fee_address are not being used with a regular Datatoken (only used in Enterprise), then what is their purpose? Have I made mistakes in replicating the functionality of buy DT and order in the regular template? Is there some transaction I'm missing that would involve these parameters?

Also, ref. returning the tx_id instead of the transaction itself: pay_for_access_service is a higher level function used in ocean assets. For models that do contract calls, we either return the receipt or wrapped models where suitable (i.e. when returning a new Datatoken or tuple of DataNFT, Datatoken, Exchange objects). We should not switch to tx_ids for functions that reside inside models, since receipts are far more rich and usable. Or, if we do, we need to level everything and do a uniform refactor for all (which should probably be another issue)

----

Response by Trent:

I'm not sure what the best answer is on this. To really know, I'd have to play around with the possibilities more myself.

Given that we aren't yet in sync on this, and it would be really useful to "ratchet up" ocean.py with all the great improvements of this PR, I recommend:

- merge the PR as-is
- copy these comments into a github issue for further discussion / playing with possibilities. (And probably lower priority than some other things)

calina-c:
Closes #1239.

### DIFF
--- a/ocean_lib/models/datatoken_enterprise.py
+++ b/ocean_lib/models/datatoken_enterprise.py
@@ -2,12 +2,12 @@
 # Copyright 2022 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
-from typing import Any, Union
+from typing import Any, Optional, Union
 
 from enforce_typing import enforce_types
 
 from ocean_lib.models.datatoken import Datatoken, TokenFeeInfo
-from ocean_lib.ocean.util import get_address_of_type
+from ocean_lib.ocean.util import get_address_of_type, get_from_address
 from ocean_lib.web3_internal.contract_base import ContractBase
 
 checksum_addr = ContractBase.to_checksum_address
@@ -27,16 +27,19 @@ class DatatokenEnterprise(Datatoken):
     @enforce_types
     def buy_DT_and_order(
         self,
-        consumer: str,
         provider_fees: dict,
         exchange: Any,
         max_base_token_amount: Union[int, str],
         consume_market_swap_fee_amount: Union[int, str],
         consume_market_swap_fee_address: str,
         tx_dict: dict,
+        consumer: Optional[str] = None,
         service_index: int = 1,
         consume_market_fees=None,
     ) -> str:
+        if not consumer:
+            consumer = get_from_address(tx_dict)
+
         fre_address = get_address_of_type(self.config_dict, "FixedPrice")
 
         # import now, to avoid circular import
@@ -77,14 +80,17 @@ class DatatokenEnterprise(Datatoken):
     @enforce_types
     def dispense_and_order(
         self,
-        consumer: str,
         provider_fees: dict,
         tx_dict: dict,
+        consumer: Optional[str] = None,
         service_index: int = 1,
         consume_market_fees=None,
     ) -> str:
         if not consume_market_fees:
             consume_market_fees = TokenFeeInfo()
+
+        if not consumer:
+            consumer = get_from_address(tx_dict)
 
         dispenser_address = get_address_of_type(self.config_dict, "Dispenser")
         return self.contract.buyFromDispenserAndOrder(
--- a/ocean_lib/models/datatoken_enterprise.py
+++ b/ocean_lib/models/datatoken_enterprise.py
@@ -8,6 +8,7 @@ from enforce_typing import enforce_types
 
 from ocean_lib.models.datatoken import Datatoken, TokenFeeInfo
 from ocean_lib.ocean.util import get_address_of_type, get_from_address
+from ocean_lib.web3_internal.constants import ZERO_ADDRESS
 from ocean_lib.web3_internal.contract_base import ContractBase
 
 checksum_addr = ContractBase.to_checksum_address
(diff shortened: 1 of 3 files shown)


ocean_lib/models/datatoken_enterprise.py
```diff
@@ -77,15 +85,18 @@ def buy_DT_and_order(
     @enforce_types
     def dispense_and_order(
         self,
-        consumer: str,
-        service_index: int,
         provider_fees: dict,
         tx_dict: dict,
+        consumer: Optional[str] = None,
+        service_index: int = 1,
```
mariacarmina, line 91: The index is not by default 0?

calina-c: I think there's 1-based indexing. I just transfered it here from the tests.

calina-c: And I think it's a remnant from when we didn't have string-based ids, so indices started at 1 because identifying with 0 is a bit weird

mariacarmina (approved):

PR #1327, opened 2023-02-13
Added custody light README.

### LINKED ISSUE #1300: Provide a clearer flow for people to create and consume custody-light data assets (ie DatatokenEnterprise)

mariacarmina:
Fixes #1300  .

Changes proposed in this PR:

- added a custody-light README
- connect to main README

### DIFF
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -343,18 +343,11 @@ class OceanAssets:
         name = metadata["name"]
         template_index = 2 if use_enterprise else 1
         data_nft_args = DataNFTArguments(name, name)
-        datatoken_args = (
-            DatatokenArguments(
-                f"{name}: DT1",
-                files=files,
-                template_index=template_index,
-                cap=MAX_UINT256,
-            )
-            if template_index == 2
-            else DatatokenArguments(
-                f"{name}: DT1", files=files, template_index=template_index
-            )
+        datatoken_args = DatatokenArguments(
+            f"{name}: DT1", files=files, template_index=template_index
         )
+        if template_index == 2:
+            datatoken_args.cap = MAX_UINT256
 
         if not pricing_schema_args:
             data_nft, datatoken = self.data_nft_factory.create_with_erc20(
--- a/ocean_lib/ocean/ocean_assets.py
+++ b/ocean_lib/ocean/ocean_assets.py
@@ -342,12 +342,11 @@ class OceanAssets:
         self._assert_ddo_metadata(metadata)
         name = metadata["name"]
         template_index = 2 if use_enterprise else 1
+        cap = MAX_UINT256
         data_nft_args = DataNFTArguments(name, name)
         datatoken_args = DatatokenArguments(
-            f"{name}: DT1", files=files, template_index=template_index
+            f"{name}: DT1", files=files, template_index=template_index, cap=cap
         )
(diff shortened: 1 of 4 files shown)


ocean_lib/ocean/ocean_assets.py
```diff
@@ -342,8 +343,17 @@ def create_bundled(
         name = metadata["name"]
         template_index = 2 if use_enterprise else 1
         data_nft_args = DataNFTArguments(name, name)
-        datatoken_args = DatatokenArguments(
-            f"{name}: DT1", files=files, template_index=template_index
+        datatoken_args = (
```
calina-c, line 346: I think you should assign this previously and simplify.
e.g. dt_args=DatatokenArguments(
                f"{name}: DT1", files=files, template_index=template_index
            )
            
            if template_index == 2:
            dt_args.cap=MAX_UINT256
```            
maybe the init part itself has the exception, but either way this needs to be cleaner

calina-c (changes requested):

ocean_lib/ocean/ocean_assets.py
mariacarmina: Sure, thank you

mariacarmina: Added an assignment before initializing the DatatokenArguments object

PR #1328, opened 2023-02-14
Adds example for bundled pricing schema args creation.

### LINKED ISSUE #1306: Document bundled asset creation + pricing schema in readmes
As part of #1166 we introduced the option of creating assets + pricing schemas. I bundled NFT+ERC20 either way for `create_<type>_asset` functions, which will save gas and do just one transaction, but the pricing schema part is optional and not that visible unless users find it in the tests.

We need to showcase these options in readmes.

calina-c:
Closes #1306.

### DIFF
--- a/READMEs/main-flow.md
+++ b/READMEs/main-flow.md
@@ -402,7 +402,28 @@ ExchangeFeeInfo:
 ```
 
 
-<h2 id="appendix-consume-details">Appendix: Consume Details</h4>
+<h2 id="appendix-create-bundled">Create asset and pricing schema simultaneously </h2>
+Ocean Assets allows you to bundle several common scenarios as a single transaction, thus lowering gas fees.
+
+Any of the `ocean.assets.create_<type>_asset()` functions can also take an optional parameter that describes a bundled pricing schema (Dispenser or Fixed Rate Exchange).
+This can be either a DispenserArguments or an ExchangeArguments object. The parameters for these Arguments classes are identical to those for creating the object itself.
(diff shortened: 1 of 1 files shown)


READMEs/main-flow.md
```diff
@@ -402,7 +402,28 @@ ExchangeFeeInfo:
...
+
+Any of the `ocean.assets.create_<type>_asset()` functions can also take an optional parameter that describes a bundled pricing schema (Dispenser or Fixed Rate Exchange).
+This can be either a DispenserArguments or an ExchangeArguments object. The parameters for these Arguments classes are identical to those for creating the object itself.
+E.g. adding `pricing_schema_args=DispenserArguments(to_wei(1), to_wei(1))` to the `create` function is equivalent to performing the creation and creating a dispenser later using `dt.create_dispenser(to_wei(1), to_wei(1))`.
+
+Here is an example involving an exchange:
+
+```python
+from ocean_lib.models.fixed_rate_exchange import ExchangeArguments
+(data_nft, datatoken, ddo) = ocean.assets.create_url_asset(
```
mariacarmina, line 416: I think this fits better in `Publish Appendix` rather than in consume.

calina-c: I thought about both options. I chose this because it's *after* the whole Dispenser/Exchange part. By the point of the publish tutorial, there is no reference to these aspects so I thought it was better here.

mariacarmina: Yes, but the user will see after publishing already an asset. I got your point, but maybe is a good point to bring in the first phase and after he can work with the flow. @trentmc what do you consider to be the best option which does not break the readability of this README?

trentmc: I prefer the status quo (after whole Dispenser/Exchange part), for the reasons Calina gave.

Issue #1259, opened 2023-01-03
CI fails: "Exception: Address not found for network_name=development". Barge issue?


trentmc:
### The Bug

Example: [this CI log](https://github.com/oceanprotocol/ocean.py/actions/runs/3829882340/jobs/6517062682) from [this PR](https://github.com/oceanprotocol/ocean.py/pull/1256).

Every test fails.

We are also seeing this in df-py CI, reported in [df-py#421](https://github.com/oceanprotocol/df-py/issues/421).

### Example traceback

```
_________________________ ERROR at setup of test_init __________________________

config = {'ADDRESS_FILE': '/home/runner/.ocean/ocean-contracts/artifacts/address.json', 'DOWNLOADS_PATH': 'consume-downloads', 'METADATA_CACHE_URI': 'http://172.15.0.5:5000/', 'NETWORK_NAME': 'development', ...}

    @pytest.fixture
    def ocean_address(config) -> str:
>       return get_address_of_type(config, "Ocean")

conftest_ganache.py:109: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/enforce_typing/decorator.py:29: in wrapper
    return func(*args, **kwargs)
ocean_lib/ocean/util.py:20: in get_address_of_type
    addresses = get_contracts_addresses(config_dict)
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/enforce_typing/decorator.py:29: in wrapper
    return func(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

config = {'ADDRESS_FILE': '/home/runner/.ocean/ocean-contracts/artifacts/address.json', 'DOWNLOADS_PATH': 'consume-downloads', 'METADATA_CACHE_URI': 'http://1[72](https://github.com/oceanprotocol/ocean.py/actions/runs/3829882340/jobs/6517062682#step:7:73).15.0.5:5000', 'NETWORK_NAME': 'development', ...}

    @enforce_types
    def get_contracts_addresses(config: dict) -> Optional[Dict[str, str]]:
[...]

PR #1330, opened 2023-02-15
Added retry mechanism for addresses retrieval.

### LINKED ISSUE #1329: Exception raised randomly: Address not found for network_name=development.
From a recent [CI run](https://github.com/oceanprotocol/ocean.py/actions/runs/4183486056/jobs/7247898911#step:8:20), this error occurred randomly which says that if cannot find the addresses development network.

**Solution**
Implement a retry mechanism for this issue.

mariacarmina:
Fixes #1329 #1259  .

Changes proposed in this PR:

- modified logic for error handling
- added number of retries

### DIFF
--- a/ocean_lib/web3_internal/contract_utils.py
+++ b/ocean_lib/web3_internal/contract_utils.py
@@ -51,7 +51,7 @@ def get_contracts_addresses_all_networks(config: dict):
 
 
 @enforce_types
-def get_contracts_addresses(config: dict) -> Optional[Dict[str, str]]:
+def get_contracts_addresses(config: dict, retries: int = 2) -> Optional[Dict[str, str]]:
     """Get addresses for given NETWORK_NAME, from info in ADDRESS_FILE"""
     network_name = config["NETWORK_NAME"]
     if network_name == "polygon-test":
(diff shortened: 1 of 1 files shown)


ocean_lib/web3_internal/contract_utils.py
```diff
@@ -51,24 +51,26 @@ def get_contracts_addresses_all_networks(config: dict):
...
+def get_contracts_addresses(config: dict, retries: int = 2) -> Optional[Dict[str, str]]:
     """Get addresses for given NETWORK_NAME, from info in ADDRESS_FILE"""
     network_name = config["NETWORK_NAME"]
     if network_name == "polygon-test":
         network_name = "mumbai"
 
     if network_name == "polygon-main":
         network_name = "polygon"
+
+    if retries == 0:
```
calina-c, line 63: there are several things I don't understand here:
- how do the retries work without sleeping? What makes the address suddenly appear between two subsequent instructions?
- I am concerned about the stop condition. What if a user calls this with retries=-3?
- The issue seems to be that even though there is an address file and barge marks itself as `ready`, it does not contain all keys. Please see the work I've done for showing the file itself. This would indicate there is a problem with barge itself, not ocean.py. Barge marks itself as `ready` before it is in fact ready. So what are we doing here: tying ocean.py functionality to a quirk in a dependency (barge). Even moreso, we are tying ocean.py functionality to a quirk in our dependency, when used inside our CI flows. Let's reconsider first the behaviour of barge and our CIs, not introduce such solutions

calina-c (changes requested):

ocean_lib/web3_internal/contract_utils.py
mariacarmina: I'll close this PR according to the last bullet, but keep the issue up.

PR #1350, opened 2023-02-27
Created gas strategy README for remote networks.

### LINKED ISSUE #1326: Expose gas strategy util function for polygon network in the README
In [our code](https://github.com/oceanprotocol/ocean.py/blob/main/tests/integration/remote/util.py#L45), we have an util function that retrieves the fees for polygon network and I think that we can mention it in the `setup-remote` README for more visibility for users, especially for ETH-Predict challenge. 
Another approach would be to create another function to include other networks as well, not only Polygon related ones.

mariacarmina:
Fixes #1326  .

Changes proposed in this PR:

- created a new README which exposes gas strategy util function
- added binding in main `README.md`

### DIFF
--- a/READMEs/gas-strategy-remote.md
+++ b/READMEs/gas-strategy-remote.md
@@ -25,27 +25,31 @@ Ensure that you've already (a) [installed Ocean](install.md), and (b) [set up re
 Fees are defined for `polygon-main` & `polygon-test` networks.
 
 ```python
-from tests.integration.remote.util import get_gas_fees_for_remote
+from ocean_lib.web3_internal.utils import get_gas_fees_for_remote
 
 priority_fee, max_fee = get_gas_fees_for_remote()
 ```
 
 ## 3. Alice publishes the asset using gas strategy
 
+The gas strategy can be added to any `tx_dict`, and this is just an example of usage.
 ```python
 #data info
 name = "Branin dataset"
 url = "https://raw.githubusercontent.com/trentmc/branin/main/branin.arff"
-
-#create data asset
-(data_nft, datatoken, ddo) = ocean.assets.create_url_asset(
-    name,
-    url,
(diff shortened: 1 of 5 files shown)


READMEs/gas-strategy-remote.md
```diff
@@ -0,0 +1,57 @@
...
+## 1. Setup
+
+Ensure that you've already (a) [installed Ocean](install.md), and (b) [set up remotely](setup-remote.md).
+
+## 2. Define gas strategy
+
+Fees are defined for `polygon-main` & `polygon-test` networks.
+
+```python
+from tests.integration.remote.util import get_gas_fees_for_remote
```
calina-c, line 28: if we want to make these "public", they should not reside in tests, they should be in a public module.

READMEs/gas-strategy-remote.md
```diff
@@ -0,0 +1,57 @@
...
+
+priority_fee, max_fee = get_gas_fees_for_remote()
+```
+
+## 3. Alice publishes the asset using gas strategy
```
calina-c, line 33: I would also make it clear that these can be added to any tx_dict, and this is just an example

ocean_lib/web3_internal/utils.py
```diff
@@ -82,3 +84,25 @@ def check_network(network_name: str):
         raise Exception(
             'Brownie network is connected to {active_network}. Please call network.connect("{network_name}")'
         )
+
+
+@enforce_types
+def get_gas_fees_for_remote() -> tuple:
```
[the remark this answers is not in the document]
mariacarmina: "Priority" is too specific, because this function returns a tuple of **gas fees**. I suggest to name it `get_gas_fees`.

calina-c (approved):

Issue #1358, opened 2023-03-07
Stopgap: Since barge doesn't currently work on MacOS or Win, say so in local-setup README 


trentmc:
### Background / Motivation

Veronica: Issue [ocean.py#1313](https://github.com/oceanprotocol/ocean.py/issues/1313) reports basically that Barge is incompatible with MacOS and Windows in the current setup-local.md. 

Ana: Making Barge compatible with MacOS & Windows is not a walk in the park. It can easily take weeks of an experienced engineer. We decided to add better documentation in barge and let people know that windows and macos are not supported.

### TODO

Add the "better documentation" as described above, to setup-local.md.

PR #1361, opened 2023-03-07
Helps #1358: clarify mac/win instructions wrt barge/local

graceful-coder:
Further improves on #1358 "Stopgap: Since barge doesn't currently work on MacOS or Win, say so in local-setup README"

Changes proposed in this PR:

- Removed some Windows/MacOS mentions from Setup-local.md
- Added setup-remote.md link to the bottom of the install.md next steps section

### DIFF
--- a/READMEs/install.md
+++ b/READMEs/install.md
@@ -60,7 +60,6 @@ Thanks to Brownie, ocean.py treats each Ocean smart contract as a Python class, 
 
 You've now installed Ocean, great!
 
-As a MacOS or Windows user, your next step is to [setup remotely](setup-remote.md).
-
 As a Linux user, your next step is to [setup locally](setup-local.md).
 
+As a MacOS or Windows user, your next step is to [setup remotely](setup-remote.md). Note: this is a temporary workaround because Ocean's Barge tool does not support MacOS nor Windows yet for local setup.
--- a/READMEs/install.md
+++ b/READMEs/install.md
@@ -60,4 +60,7 @@ Thanks to Brownie, ocean.py treats each Ocean smart contract as a Python class, 
 
 You've now installed Ocean, great!
 
-Your next step is to [setup locally](setup-local.md).
+As a MacOS or Windows user, your next step is to [setup remotely](setup-remote.md).
+
+As a Linux user, your next step is to [setup locally](setup-local.md).
+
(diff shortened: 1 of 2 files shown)


READMEs/install.md
```diff
@@ -60,4 +60,7 @@ Thanks to Brownie, ocean.py treats each Ocean smart contract as a Python class,
 
 You've now installed Ocean, great!
 
-Your next step is to [setup locally](setup-local.md).
+As a MacOS or Windows user, your next step is to [setup remotely](setup-remote.md).
+
+As a Linux user, your next step is to [setup locally](setup-local.md).
```
trentmc, line 65: The "default" flow is supposed to be to setup locally. Setup remote is supposed to come later.

We only need to make changes for a temporary workaround.

Accordingly, my suggestion for the changes is either:
- don't make _any_ changes, since "setup local" redirects the user at the very top anyway
- flip around the order of the changes above, and mention that for Mac/Win it's a workaround because of barge/local issues.

READMEs/setup-local.md
```diff
@@ -52,9 +52,8 @@ cd my_project
 source venv/bin/activate
```
trentmc, line 52: I wonder if we want to have these other changes that remove Mac & Win.

When barge gets fixed, them we'd have to re-insert them. I worry that it will be error-prone to reinsert everything.

Two options:
1. Don't make these changes
2. Keep these changes, but also keep track of possible things to re-insert once Mac/Win are working on local

I'm on the fence on this, ie I'm fine either way.
I'm a bit on the fence on this.

trentmc (changes requested):

READMEs/install.md
graceful-coder: I feel that the redirection at the top of the readme could easily be missed by skimming the content. Therefore, I recommend that we follow the second option.

READMEs/setup-local.md
graceful-coder: I suggest to re-insert the changes later as I will be filming the local setup again on a Mac, so I will notice then the instructions need clarity for MacOS + Windows at that point in time.

READMEs/install.md
trentmc: OK. Please go ahead with that:)

READMEs/setup-local.md
trentmc: OK. Please go ahead with that:)

graceful-coder (commented):
Reviewed changes.

trentmc (approved):

Issue #929, opened 2022-09-18
ocean.py CI tests take 27 min (too slow). Optimize to <=10 min


trentmc:
**Is your feature request related to a problem? Please describe.**

It takes 27 min for all CI tests to complete. It's about this long locally too (depending on the machine).

This is SLOW. It hurts the ability to have fast turnaround. An upper bound should be 10 min.

**Describe the solution you'd like**

Optimize the test-running so that total CI runtime is <10 min.

**Possible approaches**

- run benchmarks. Optimize slowest tests. Possible ways to optimize: 
  - optimize the test itself. Eg repeat a 1000 step loop on the chain with advancing the chain by 1000 blocks (single function call). Etc.
  - optimize the ocean.py functionality
  - remove the test or merge it with other tests
- skip importing conftest_ganache.py unless needed. Currently all but one test*.py files import it, despite many of them note even using ganache / barge. Identify these, and don't import those. (This is an opportunity recently introduced due to fixing #923)
- parallelize the CI. (Is it already partly parallel? If yes, use the same pattern to make it even more parallel. Or, what other tricks do Github Actions or pytest offer?)

calina-c:
ocean.py is the designated "dumpsite" for all integration tests possible, and it always has been.

- Testing the contract fees? ocean.py should do that.
- Testing full publish-to-c2d flows? Yes, ocean.py should do that with 3 different compute examples and installing another 5 3rd party dependencies just for testing the examples.
- Testing local and remote user flows, contract functions, automations? The answer was always yes

Ocean.py now has: separate slow tests which last about 3 hours, library compatibility tests which run every tuesday to make sure our library is up to date with the rest of the stack, automated readme tests which ensure our users always have things working out of the box.

We need to have a discussion about what ocean.py needs to test and how we should structure it, then write a few issues. I also don't think we can reach under 10 minutes, our best performing CIs in Aquarius and Provider are still around 15 minutes with publish waiting times etc.

trentmc:
Update: we made some nice progress on this with recent changes by @calina-c :
- "remove enterprise README", and therefore its associated test - https://github.com/oceanprotocol/ocean.py/commit/3bc3478e529e8b2d64d4185e3bfd527dcfaa2ac3
- "remove slow tests" - commit  https://github.com/oceanprotocol/ocean.py/commit/acb029caa1b78285c3cfe491c5d651e15dd25323

More to go:
- remove other slow tests?
- profiling & optimizing tests
[...]

PR #1393, opened 2023-04-11
Optimise tests

### LINKED ISSUE #1396: Add flexible parameters in create_<type>_asset
Per discussions with @alexcos20 we keep needing to add new parameters to those functions that call create_bundled. We should make an implementation similar to that for DatatokenArguments.

calina-c:
Work on #929. Shaves off ~3mins from current pytest job. Please note that @mariacarmina also did a great job recently by reducing the time for the Barge job, so my work is just a bonus :).

What I did:
- reuse a `basic_asset` fixture when doing non-update tests
- leverage ocean.py creation methods in c2d flow and others -> I'll follow up with some ideas here in another ticket
- remove most duplications of default metadata and ddo_helpers
- leverage wait_for_aqua = False in the former start_order_fees tests: assets are now published during the test class setup, we only wait for them when needed and the concurrent creation makes waiting times MUCH shorter
- combine reuse_order and start_order tests to leverage the latter's receipt when needed, instead of doing a duplicated call
- re-enable c2d readme test, ensure local setup only runs once for readme tests
- add flexibility for metadata and compute values handling in `create_<type>_asset` functions to reduce number of update calls

I suggest merging this in, and doing more optimisations later.

Closes #1396.


alexcos20 (approved):
lgtm

PR #1409, opened 2023-04-17
Add timezone info to all objects.

calina-c:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/READMEs/c2d-flow.md
+++ b/READMEs/c2d-flow.md
@@ -124,7 +124,7 @@ compute_service = DATA_ddo.services[1]
 algo_service = ALGO_ddo.services[0]
 free_c2d_env = ocean.compute.get_free_c2d_environment(compute_service.service_endpoint, DATA_ddo.chain_id)
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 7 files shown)


trentmc:
What's the problem this is trying to solve?

calina-c:
> What's the problem this is trying to solve?

As reported here https://github.com/oceanprotocol/ocean.py/pull/1307 the timezone objects we are using in our Python repos are unaware of timezone settings. They represent the correct UTC time but are unaware if that time is in the UTC timezone. It's a caveat of the Python time functions, which offer both timezone-aware and timezone-naive functions. Therefore, for safe measure, I've added timezone-awareness in all Python repos.

Aqua: https://github.com/oceanprotocol/aquarius/pull/1017
Provider: https://github.com/oceanprotocol/provider/pull/640
and this one for ocean.py

trentmc:
Thanks for the explanation @calina-c. Makes sense.

To help tracking: could you
- create a github issue for this (basically saying what you said in the your comment above)
- in this PR's description, link to the issue

Thanks!

alexcos20 (approved):

PR #1412, opened 2023-04-18
Leverage true provider methods instead of hardcoding.

### LINKED ISSUE #1411: Method does not correspond in DataServiceProvider class
Currently from provider we expose the method and the URLs for different endpoints. The dataservice provider class has functions that resolve those in order to create flexible requests.

While the endpoint part is good, the "method" does not always correspond to the actual method that needs to be used by requests, throwing a 405 error if I use it directly. Instead, the code relies on prior knowledge of methods.

Possible solutions and arguments:
1. fix these in Provider -> not sure this is appropriate, what if an endpoint accepts both GET and POST requests?
2. continue to rely on prior knowledge of the method -> In this case, there is no point in resolving the method and we could remove that code. Having it there is prone to confusion.

calina-c:
Closes #1411

### DIFF
--- a/ocean_lib/data_provider/base.py
+++ b/ocean_lib/data_provider/base.py
@@ -67,7 +67,9 @@ class DataServiceProviderBase:
         """
         Return the service endpoints from the provider URL.
         """
-        provider_info = DataServiceProviderBase.get(url=provider_uri)
+        provider_info = DataServiceProviderBase._http_method(
+            "get", url=provider_uri
+        ).json()
 
         return provider_info["serviceEndpoints"]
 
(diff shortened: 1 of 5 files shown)


trentmc (commented):
This is a corner of code I don't know well, so I can't really do a thorough review.

However I can ask: 
- were there tests for this before? 
- could there be a test added or modified that would have failed with old approach, and would pass with your mods? (if not too complex)

calina-c:
> This is a corner of code I don't know well, so I can't really do a thorough review.
> 
> However I can ask:
> 
> * were there tests for this before?
> * could there be a test added or modified that would have failed with old approach, and would pass with your mods? (if not too complex)

The code worked because we were using prior knowledge of the endpoints instead of using the method returned by provider. When I tried leveraging the `method` itself during test fixes, I encountered 405 errors (method not allowed) because the initialize compute was incorrectly configured as a GET request instead of POST. I updated the configuration but I can show you a previous test run: https://github.com/oceanprotocol/ocean.py/actions/runs/4689970392/jobs/8312421605

For the purposes of test fixes, I decided to revert to hardcoded methods and treat the issue later (in this ticket).

trentmc:
@calina-c thanks for the info. Worksforme.

The changes look sane, but for a proper review I defer to Alex.

mariacarmina (approved):

alexcos20 (approved):

PR #1425, opened 2023-05-11
Predictoor with barge

alexcos20:
- For [this epic](https://github.com/oceanprotocol/predictoor-pm/issues/4) "[Predictoor BE] Ship MVP" -- Add DT3 to ocean.py repo
- Fix [ocean.py#1426](https://github.com/oceanprotocol/ocean.py/issues) - proper usage of fixedrates/dispensers


alexcos20:
Ready for review.
Two compute tests are failing, unrelated with code from this PR

.github/workflows/pytest.yml
```diff
@@ -17,20 +17,32 @@ jobs:
...
+          echo "Login to Docker Hub";echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+        env:
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
       - name: Run Barge
         working-directory: ${{ github.workspace }}/barge
         run: |
           bash -x start_ocean.sh --no-dashboard 2>&1 --with-provider2 --with-c2d --with-thegraph --skip-subgraph-deploy > start_ocean.log &
+        env:
+          CONTRACTS_VERSION: predictoor
```
trentmc, line 44: I'm thinking more about this. Needing a special version just for predictoor -- just to work around immunefi policy -- is gonna be a huge pain.

I created a new github issue to discuss this more, and get to a resolution. [predictoor-pm#18](https://github.com/oceanprotocol/predictoor-pm/issues/18)

README.md
```diff
@@ -3,6 +3,10 @@ Copyright 2023 Ocean Protocol Foundation
 SPDX-License-Identifier: Apache-2.0
 -->
 
+**[Predictoor README](READMEs/predictoor.md).** (Remove this when done prototype)
+
+----
+
```
trentmc, line 9: Let's create a special "Predictoor" section closer to the bottom of the README. Include this here. Later, we'll split out a predictoor-dev.md (current contents of predictoor.md) from predictoor.md (flows for predictoors, opf, traders)

trentmc (changes requested):
Started reviewing, a couple comments. More to come, but have to run for now.

trentmc (changes requested):
More comments. More to come (on Mon:)

ocean_lib/models/datatoken3.py
```diff
@@ -0,0 +1,128 @@
...
+         - Exchange's base token == self.stake_token
+         - Exchange's owner is == this DT itself (!). Why: for payouts
+        """
+        self.exchange = self.create_exchange(
+            tx_dict,
+            rate,
+            base_token_addr=self.stake_token(),
+        )
+
+    def get_empty_provider_fee(self):
```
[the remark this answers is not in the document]
alexcos20: no, all other templates will invalidate this structure and revert

ocean_lib/models/datatoken1.py
```diff
@@ -256,17 +255,19 @@ def buy_DT_and_order(
...
+        exchanges = self.get_exchanges()
+        assert (
+            len(exchanges) > 0
+        ), "there are no fixed rate exchanges for this datatoken"
         # import now, to avoid circular import
         from ocean_lib.models.fixed_rate_exchange import OneExchange
 
+        if not isinstance(exchange, OneExchange):
+            exchange = exchanges[0]
+
```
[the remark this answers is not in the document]
alexcos20: no, because function buy_DT_and_order already has exchange parameter.  Maybe I want to buy from the 3rd exchange

ocean_lib/models/datatoken2.py
```diff
@@ -41,13 +41,15 @@ def buy_DT_and_order(
...
+        exchanges = self.get_exchanges()
+        assert (
+            len(exchanges) > 0
+        ), "there are no fixed rate exchanges for this datatoken"
         # import now, to avoid circular import
         from ocean_lib.models.fixed_rate_exchange import OneExchange
 
         if not isinstance(exchange, OneExchange):
-            exchange = OneExchange(fre_address, exchange)
+            exchange = exchanges[0]
```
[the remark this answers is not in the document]
alexcos20: no, because function buy_DT_and_order already has exchange parameter. Maybe I want to buy from the 3rd exchange

.github/workflows/pytest.yml
```diff
@@ -17,20 +17,32 @@ jobs:
...
           python-version: '3.8'
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         name: Checkout Barge
         with:
           repository: "oceanprotocol/barge"
           path: 'barge'
+      - name: Set ADDRESS_FILE
+        run: echo "ADDRESS_FILE=${HOME}/.ocean/ocean-contracts/artifacts/address.json" >> $GITHUB_ENV
+      - name: Login to Docker Hub
```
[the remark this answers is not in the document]
alexcos20: See [here](https://docs.docker.com/docker-hub/download-rate-limit/#whats-the-download-rate-limit-on-docker-hub)
I personally added auth to all of our repos a year ago.  But looks like in the meantime, somebody removed it. God knows why.

.github/workflows/pytest.yml
alexcos20: it's not just immunify.   I will release a new version of contracts as soon as we have the entire flow tested, with UI.    Otherwise, we will release a new patch every couple of hours :)

ocean_lib/models/datatoken_base.py
```diff
@@ -175,7 +176,12 @@ def get_typed(config, address):
         except Exception:
             template_id = 1
 
-        return datatoken if template_id == 1 else Datatoken2(config, address)
+        if template_id == 1:
+            return Datatoken1(config, address)
+        if template_id == 2:
+            return Datatoken2(config, address)
+        if template_id == 3:
+            return Datatoken3(config, address)
```
trentmc, line 184: Change to:
```python
        if template_id == 1:
            return Datatoken1(config, address)
        elif template_id == 2:
            return Datatoken2(config, address)
        elif template_id == 3:
            return Datatoken3(config, address)
        else:
            return ValueError(template_id)
```

Why: more pythonic if/else. And, catch error when unknown input arg.

ocean_lib/models/datatoken_base.py
```diff
@@ -287,16 +293,22 @@ def create_exchange(
...
         addrs_and_exchange_ids = self.getFixedRates()
-        exchanges = [
-            OneExchange(FRE, exchange_id) for _, exchange_id in addrs_and_exchange_ids
-        ]
+        for address, exchange_id in addrs_and_exchange_ids:
+            fre = FixedRateExchange(self.config_dict, address)
+            exchange = OneExchange(fre, exchange_id)
+            if active and not exchange.is_active():
+                continue
+            exchanges.append(exchange)
```
trentmc, line 311: Slightly cleaner, because two lines and avoids negative:
```python
            if active and exchange.is_active():
                exchanges.append(exchange)
```

ocean_lib/models/datatoken_base.py
```diff
@@ -338,6 +350,22 @@ def create_dispenser(self, tx_dict: dict, *args, **kwargs):
...
+        from ocean_lib.models.dispenser import Dispenser
+
+        dispensers = []
+        addrs = self.getDispensers()
+        for address in addrs:
+            dispenser = Dispenser(self.config_dict, address)
+            is_active = dispenser.status(self.address)
+            if active and not is_active:
+                continue
+            dispensers.append(dispenser)
```
trentmc, line 366: FYI, `dispenser.status()` returns a tuple. Therefore the line `is_active = dispenser.status(self.address)` has a bug. Better is to call `dispenser_status()` to get a `DispenserStatus` object.

Here's a fix to the bug, where code is cleaner too. 
```python
            if active and self.dispenser_status().active:
                dispensers.append(dispenser)
```

ocean_lib/models/datatoken_base.py
```diff
@@ -394,9 +422,26 @@ def build_access_service(
...
         )
 
+    @enforce_types
     def get_publish_market_order_fees(self):
         return TokenFeeInfo.from_tuple(self.contract.getPublishingMarketFee())
 
+    @enforce_types
+    def approve_publish_market_order_fees(self, tx_dict: dict):
+        fees = self.get_publish_market_order_fees()
+        if fees.amount > 0 and fees.address != ZERO_ADDRESS:
```
trentmc, line 432: Recommend: change this line to:
```python
if fees.amount == 0 or fees.address == ZERO_ADDRESS or fees.token == ZERO_ADDRESS:
    return
```

And also un-indent the rest of the method.

Why: 
- `fees.token` portion added, because it's also a reason to skip. And, `ERC20Template3.sol::orderStarted()` has similar logic. 
- simpler & easier to maintain: you avoid having to indent the remaining 11 lines of this method.

trentmc (approved):
OK, reviewed. Nice job.

I added a couple TODOs to "[Predictoor BE] MVP" issue, so that we don't need to do them _here_.
-  In test_datatoken3.py, properly test payout to Predictoors and OPF
-  Add unit test(s) for new routine contract_utils::get_contract_filename()

ocean_lib/models/test/test_datatoken3.py
```diff
@@ -0,0 +1,297 @@
...
+    OCEAN.approve(exchange.address, to_wei(OCEAN_needed), {"from": trader})
+
+    consume_market_fee_addr = ZERO_ADDRESS
+    consume_market_fee = 0
+    tx = exchange.buy_DT(
+        datatoken_amt=to_wei(1),
+        consume_market_fee_addr=consume_market_fee_addr,
+        consume_market_fee=consume_market_fee,
+        tx_dict={"from": trader},
+    )
```
trentmc, line 104: Can simplify, because in `buy_DT()` the default consume market fee is zero. [Ref](https://github.com/oceanprotocol/ocean.py/blob/a8a8a94fd97a149629cb69df1701108d3a637b41/ocean_lib/models/fixed_rate_exchange.py#L309)

And, you don't need the return `tx`.

Therefore it will look like:
```python
    exchange.buy_DT(datatoken_amt=to_wei(1), tx_dict={"from": trader})
```

ocean_lib/models/test/test_datatoken3.py
```diff
@@ -0,0 +1,297 @@
...
+    # trader starts subscription. Good for 24h.
+    DT.approve(DT.address, to_wei(1), {"from": trader})
+    DT.start_subscription_with_DT({"from": trader})  # good for next 24h
+
+    assert DT.isValidSubscription(trader.address) == True
```
trentmc, line 109: Can remove ` == True`, that's redundant

ocean_lib/models/test/test_datatoken3.py
```diff
@@ -0,0 +1,297 @@
...
+            aa = OCEAN_approved(p, DT_treasurer)
+            n_predns[p_i] += 1
+            actions_s += (
+                f"Predictoor{p_i+1} "
+                + f" submitted pred'n #{n_predns[p_i]}"
+                + f" at block B={predict_blocknum}, epoch E={DT.curEpoch()}."
+                + f" Now, OCEAN approved={aa:.1f}\n"
+            )
+            blocks_predicted.add(predict_blocknum)
+        print(blocks_predicted)
```
trentmc, line 143: Add a space before new section `# TRADER gets...`

.github/workflows/pytest.yml
```diff
@@ -17,20 +17,32 @@ jobs:
...
         uses: actions/setup-python@v2
         with:
           python-version: '3.8'
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         name: Checkout Barge
         with:
           repository: "oceanprotocol/barge"
           path: 'barge'
+      - name: Set ADDRESS_FILE
```
calina-c, line 30: can you set it as an env directly for barge?

ocean_lib/models/datatoken3.py
```diff
@@ -0,0 +1,123 @@
...
+        subscr_addr = tx_dict["from"].address
+        assert not self.is_valid_subscription(
+            subscr_addr
+        ), "this account has already started a subscription"
+        self.approve_publish_market_order_fees(tx_dict)
+        exchanges = self.get_exchanges()
+        assert exchanges, "there are no fixed rate exchanges for this datatoken"
+        exchange = exchanges[0]
+        amt_needed = exchange.BT_needed(to_wei(1), 0)
+        provider_fees = self.get_empty_provider_fee()
```
calina-c, line 80: I am a bit confused. The function defined above is get "zero" provider fee, this is "empty" provider fee. Are they different? Where is this defined?

ocean_lib/models/datatoken3.py
```diff
@@ -0,0 +1,123 @@
...
+        """How much e.g. OCEAN does this contract have?"""
+        return from_wei(self.stake_token.balanceOf(self.treasurer))
+
+    def _sell_1DT(self):
+        amt_DT_wei = to_wei(1.0)
+        amt_BT_wei = self.exchange.BT_received(amt_DT_wei, consume_market_fee=0)
+
+
+@enforce_types
+def _cur_blocknum() -> int:
```
calina-c, line 122: this doesn't seem used in other places. Do we need it?

ocean_lib/models/datatoken_base.py
```diff
@@ -287,16 +293,22 @@ def create_exchange(
...
         """return List[OneExchange] - all the exchanges for this datatoken"""
         # import now, to avoid circular import
         from ocean_lib.models.fixed_rate_exchange import OneExchange
 
-        FRE = self._FRE()
+        # import now, to avoid circular import
+        from ocean_lib.models.fixed_rate_exchange import FixedRateExchange
+
+        exchanges = []
         addrs_and_exchange_ids = self.getFixedRates()
```
calina-c, line 305: I think there's a function that retrieves the objects directly, something like "get_exchanges" but I'm not sure about that. It could save some lines and clarity with instantiating everything

ocean_lib/models/datatoken3.py
```diff
@@ -0,0 +1,123 @@
...
+         - Exchange's base token == self.stake_token
+         - Exchange's owner is == this DT itself (!). Why: for payouts
+        """
+        self.exchange = self.create_exchange(
+            tx_dict,
+            rate,
+            base_token_addr=self.stakeToken(),
+        )
+
+    def get_zero_provider_fee(self):
```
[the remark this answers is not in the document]
trentmc: It's not applicable to other flows. It's specific to Predictoor.

ocean_lib/models/datatoken3.py
trentmc: Good catch. We don't need it. It's an artifact of an earlier prototype. I defer to Alex to delete.

ocean_lib/web3_internal/contract_utils.py
```diff
@@ -31,6 +31,35 @@ def get_contract_definition(contract_name: str) -> Dict[str, Any]:
         return json.load(f)
 
 
+@enforce_types
+def get_contract_filename(contract_name: str) -> str:
```
calina-c, line 35: I think this can be simplified. Let's find and assert the path first, then go through the contracts. It will be much clearer to read. And it can be split in two different functions:
- one for getting the base location of address file and contracts (I think something similar already exists but it only accounts for the artifacts imported path and not ADDRESS_FILE env)
- one for retrieving the contract based on that location from (1)

PR #1429, opened 2023-05-18
Fix nonce computation

### LINKED ISSUE #1430: Fix nonce computation for signature
**Describe the bug**
Nonce in ocean.py is computed in seconds UNIX timestamp format and provider endpoint returns the timestamp in miliseconds. This kind of mismatch generates `InvalidSignature error` from provider side. As fallback, in case nonce is not stored in provider sqllite instance, generate a new nonce in miliseconds format for UTC timezone.

**Logs**
```console
ocean_provider.utils.accounts ERROR Invalid signature expected nonce (1680730988846) > current nonce (1684325597.242274).

1680730988846
1684325597       .242274 -> see missing digits
```

mariacarmina:
Fixes #1430 .

Changes proposed in this PR:

- transformed UNIX timestamp in milisecs


calina-c:
The nonce endpoint is just for... orientation. The nonce should not be formed via the endpoint, but using the nonce logic we have defined (needs to be larger than last time).

mariacarmina:
> The nonce endpoint is just for... orientation. The nonce should not be formed via the endpoint, but using the nonce logic we have defined (needs to be larger than last time).

I have converted to miliseconds format because by default `timestamp()` function converts to UNIX seconds.

calina-c (approved):

calina-c:
Does this supersede the rest of #1307?

mariacarmina:
> Does this supersede the rest of #1307?

It is an important aspect and this PR can be related to #1307 issue.

PR #1453, opened 2023-07-28
Use nonce endpoint to retrieve nonce in asset download.

### LINKED ISSUE #1452: Verify & update if needed the nonce logic
We've refactored the user nonce logic in the provider code via this PR: https://github.com/oceanprotocol/provider/pull/649 due to various bugs we were facing(like the invalid signature bug).
In the ocean.js we already have a PR: https://github.com/oceanprotocol/ocean.js/pull/1747 to integrate these changes. 
I understand that this shouldn't be a breaking change for ocean.py but we still need to make sure things are still working as expected. 
cc @calina-c @trentmc

calina-c:
Closes #1452.

### DIFF
--- a/ocean_lib/data_provider/base.py
+++ b/ocean_lib/data_provider/base.py
@@ -46,8 +46,23 @@ class DataServiceProviderBase:
 
     @staticmethod
     @enforce_types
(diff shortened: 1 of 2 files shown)


alexcos20 (approved):
lgtm

PR #1488, opened 2023-10-25
Change config to use RPC URL directly.

### LINKED ISSUE #1479: RPC envvar label shouldn't be specific to the network
Quote from [README](https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/setup-remote.md):

> Export env vars of the format NETWORKNAME_RPC_URL e.g. export POLYGON_RPC_URL=https://polygon-rpc.com
> In case you have an Infura project, you need to also export the WEB3_INFURA_PROJECT_ID variable alongside the base rpc urls.

What if I want to use another rpc provider? Or my own network ?  Who do I know what ENV to use for EWC? etc...

Ocean.py should not deal with ENVs, it should use the rpc in the following way:

```
config = get_config_dict(RPC_URL)
ocean = Ocean(config)
```

where get_config_dict should:
 - create web3 object
 - detect the chain and return the proper config

calina-c:
Closes #1479.

### DIFF
--- a/ocean_lib/example_config.py
+++ b/ocean_lib/example_config.py
@@ -7,6 +7,7 @@ import copy
 import logging
 import os
 from pathlib import Path
+from typing import Optional
 
 import addresses
 from enforce_typing import enforce_types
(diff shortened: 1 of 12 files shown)


ocean_lib/example_config.py
```diff
@@ -22,73 +23,63 @@
...
-
-    if os.getenv("WEB3_INFURA_PROJECT_ID"):
-        base_url = f"{base_url}{os.getenv('WEB3_INFURA_PROJECT_ID')}"
-
-    if base_url:
-        return base_url
-
-    raise Exception(f"Need to set {network_name.upper()}_RPC_URL env variable.")
+NAME_PER_NETWORK = {
+    1: "mainnet",
```
alexcos20, line 45: we should get network name from [addresses.json](https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json) , but we can create another ticket for this

alexcos20 (approved):

PR #1534, opened 2024-03-27
Fix Eth Sepolia Chain ID

therealkarubinu:
Eth Sepolia: 11155111
Sepolia PGN (Public Goods Network): 58008

Fixes # .

Changes proposed in this PR:

- 
- 
-


alexcos20 (approved):
approved, is a match according to https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json#L281

PR #1544, opened 2024-10-14
Update crypto.py

famouswizard:
Optimizations:

Inline Operations: Combined multiple steps, like encoding/decoding, directly within the return statement. This reduces the code's verbosity and makes it more efficient.

Avoid Unnecessary Variable Assignments: Removed redundant variable assignments (e.g., intermediate variables like value_b, symkey_b, etc.), which optimizes memory usage and improves readability.

SHA256 Hash Encoding: Directly uses .digest() instead of .hexdigest() for more compact byte representation, making encoding faster.

### DIFF
--- a/ocean_lib/ocean/crypto.py
+++ b/ocean_lib/ocean/crypto.py
@@ -1,69 +1,31 @@
-#
-# Copyright 2023 Ocean Protocol Foundation
-# SPDX-License-Identifier: Apache-2.0
-#
-
 from base64 import b64encode
 from hashlib import sha256
-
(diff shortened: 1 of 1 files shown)


trentmc:
Thank you for the proposed contribution.

This is Python, not Perl. The point of >1 lines is to make each step clear and obvious. Squishing it all into one line makes it impossible to understand what the code is doing, let alone debug it. Having some intermediate variables doesn't matter that much wrt efficiency because efficiency is not the main goal here. 

We will _not_ be merging this.

